PR #21, opened 2020-08-30
Improve parser API

patrickelectric:
The actual `peek` and `nth` functions of `Parser` are a bit trick, both should return the same object since they are an index based access (`peek` == `nth(0)` IMO).
The PR changes `nth` to return `Token` (like the `peek` function) and also adds the `kind` function of `Token` to allows access of the `kind` property. The final result is a much more flexible API, where the user will be able to access `nth` `Token` object and `SyntaxKind` property, otherwise it's only possible to access the current `Token` and the `nth` `SyntaxKind`.

PS: This change is helping me a lot to develop/debug some functionalities since `Token` also derives `Debug`.

### DIFF
--- a/sixtyfps_compiler/parser.rs
+++ b/sixtyfps_compiler/parser.rs
@@ -378,6 +378,10 @@ impl Token {
     pub fn as_str(&self) -> &str {
         self.text.as_str()
     }
+
+    pub fn kind(&self) -> SyntaxKind {
+        self.kind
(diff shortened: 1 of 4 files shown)


patrickelectric:
CLA assigned.

tronical:
Cool, thanks :) .

I hope you don't mind if I use "rebase and merge".

PR #22, opened 2020-08-30
Allow bidirectional conversion of length <-> float and length to string.

patrickelectric:
Allows to debug size of elements and visualization of such `Length` variables.

Screen Capture_select-area_20200830192338

### DIFF
--- a/sixtyfps_compiler/generator/rust.rs
+++ b/sixtyfps_compiler/generator/rust.rs
@@ -679,7 +679,9 @@ fn compile_expression(e: &Expression, component: &Rc<Component>) -> TokenStream 
         Expression::Cast { from, to } => {
             let f = compile_expression(&*from, &component);
             match (from.ty(), to) {
-                (Type::Float32, Type::String) | (Type::Int32, Type::String) => {
+                (Type::Length, Type::String)
+                | (Type::Float32, Type::String)
+                | (Type::Int32, Type::String) => {
                     quote!(sixtyfps::re_exports::SharedString::from(format!("{}", #f).as_str()))
                 }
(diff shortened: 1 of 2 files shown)


ogoffart:
Hi @patrickelectric  thanks for your pull requests. It feels amazing to see people contributing on our project. Before accepting the contribution, we will have to set up a CLA, we haven't set that up yet so we'll come back at you soon with that.

Regarding the change itself, this goes against our ideas of having strong units in the language. The idea is that we would want explicit conversion as it avoid mistake. We don't want implicit conversion from length to number. If you want to print a string, you can do:  `text: width / 1px;`  Clearly, this should be more discoverable: the error message should probably tell you to do that.
I'm wondering what you think about that. Maybe the idea to have strongly typed unit too complicated? We just wanted to try it out and see how it goes, and your feedback is good.

Implementation wise, for this kind of changes, we should have tests in  `tests/cases/` that tests the different backend. You have implemented the conversion on the rust code generator, but this might not work properly with the C++ code generator or the interpreter.

ogoffart:
(Commit 2bb7616b80f01708f4f0d47468df1410b6daa738 added a hint in the error message that should help users wanting to do the conversion)

patrickelectric:
Hi @ogoffart, my initial idea was to aim to a more QML~javascript approach where the type conversion is not explicit, regards that the PR can be closed.

Just to make it clear, are you aiming to an unit approach where you turn `'X Length' / 'Y Length'` in a `X/Y` dimensionless number ? I saw some code that points to this direction and may explain what is the idea behind the `/` operator.

ogoffart:
Thanks for your pull request anyway.

Yes,  when you do  `width/1px` you end up with a dimensionless number  and that can be converted to string

PR #58, opened 2020-09-07
Fix windows build with warm cache

tronical:
This is one solution for the warm cache problem causing the lack of the nodejs import library.


tronical:
Going for #59 instead

PR #69, opened 2020-10-02
Create format function for .60

patrickelectric:
I was unable to come up with something easy to implement related to this, so I'm opening the patch as a draft to seek help.
My idea is to create a format function that works like rust `format!` macro for .60 templates, any tip ?

Signed-off-by: Patrick José Pereira <patrickelectric@gmail.com>

### DIFF
--- a/sixtyfps_compiler/expression_tree.rs
+++ b/sixtyfps_compiler/expression_tree.rs
@@ -43,6 +43,7 @@ impl Hash for NamedReference {
 pub enum BuiltinFunction {
     GetWindowScaleFactor,
     Debug,
+    Format,
     SetFocusItem,
 }
 
@@ -54,6 +55,9 @@ impl BuiltinFunction {
             }
             BuiltinFunction::Debug => {
                 Type::Function { return_type: Box::new(Type::Void), args: vec![Type::String] }
+            }
+            BuiltinFunction::Format => {
+                Type::Function { return_type: Box::new(Type::String), args: vec![Type::String, Type::String] }
             }
             BuiltinFunction::SetFocusItem => Type::Function {
                 return_type: Box::new(Type::Void),
--- a/sixtyfps_compiler/generator/rust.rs
+++ b/sixtyfps_compiler/generator/rust.rs
@@ -883,6 +883,7 @@ fn compile_expression(e: &Expression, component: &Rc<Component>) -> TokenStream 
                 quote!(#window_ref.scale_factor)
             }
             BuiltinFunction::Debug => quote!((|x| println!("{:?}", x))),
+            BuiltinFunction::Format => quote!(|x| format!("{:?}", x)),
             BuiltinFunction::SetFocusItem => panic!("internal error: SetFocusItem is handled directly in CallFunction")
         },
         Expression::ElementReference(_) => todo!("Element references are only supported in the context of built-in function calls at the moment"),
--- a/examples/gallery/gallery.60
+++ b/examples/gallery/gallery.60
@@ -59,7 +59,7 @@ App := Window {
         }
         Row {
             GroupBox {
-                title: "SpinBox";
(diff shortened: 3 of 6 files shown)


tronical:
Welcome back! :)

There are two aspects to it, I think:

1. What's the syntax?
2. How can it be implemented?

Regarding the syntax, we could indeed use `format` like in Rust (or `fmt` in C++). There's also the option of using JavaScript's template literal syntax (`hello {world}`). I'm personally leaning towards the JS syntax, but I admit that `format` has also a nice touch to it!

Regarding the implementation, I think that this could be implemented using string concatenation. So something like `format("Hello {} World {} Bar", a, b)` could be implemented like this (pseudo code):

```
    let mut output = "Hello ";
    output.push_str(a.to_string());
    output.push_str(" World ");
    output.push_str(b.to_string());
    output.push_str(" Bar");
```

What do you think?

sixtyfps_compiler/expression_tree.rs
```diff
@@ -55,6 +56,9 @@ impl BuiltinFunction {
             BuiltinFunction::Debug => {
                 Type::Function { return_type: Box::new(Type::Void), args: vec![Type::String] }
             }
+            BuiltinFunction::Format => {
+                Type::Function { return_type: Box::new(Type::String), args: vec![Type::String, Type::String] }
```
ogoffart, line 60: The best would be to have support for variadic function.  I believe the `debug`  function would also gain in being variadic.  
Either a new bolean `is_variadic` in `Type::Function`   or a new enum variant in Type, such as `Type::VariadicParameter(Box<Type>)`

sixtyfps_compiler/generator/rust.rs
```diff
@@ -883,6 +883,7 @@ fn compile_expression(e: &Expression, component: &Rc<Component>) -> TokenStream
                 quote!(#window_ref.scale_factor)
             }
             BuiltinFunction::Debug => quote!((|x| println!("{:?}", x))),
+            BuiltinFunction::Format => quote!(|x| format!("{:?}", x)),
```
ogoffart, line 886: Remember this is  supposed to return a function with two arguments.  If we want to support variadic, this cannot be done here and need to be done in  the handler for `CallFunction` like `BuiltinFunction::SetFocusItem` this is not ideal and we probably should refactor that.

ogoffart (commented):
Hi, thank you for the contribution again!

Indeed a way to format string is missing.
As Simon said, we were thinking about adding support for interpolated string in the syntax itself, something like
```
   text:  `The window with is { root.width / 1px } and its title is { root.title }` ;
```
This of course this syntax is just a suggestion we are not sure about.
Another possible syntax would be to just use normal quotes
```
 text:  "The window with is \{ root.width / 1px } and its title is \{ root.title }";
```

If you wanted to implement this, you would need to adjust the lexer to split the string in token that can be parsed independently: `StringLiteral  OpenInterpolation Identifier Dot Identifier ...  CloseInterpolation StringLiteral OpenInterpolation ... `  Then the parser could build an expression tree that would have a `Expression::InterpolatedString { components: Vec<Expression> }` that would then be produced.  That is of course just one possible way to do it.

Anyway, this is orthogonal to a format function which would also have its uses.

a `format` function would also be useful. We would need support for variadic argument in functions.
We should discuss the syntax. Should the format string be known at compile time, or also at runtime?  In which case, we would need to call into a runtime format function.

Thanks again for your contribution.
I'm also curious about how you found this project and your motivations for contributing.
Would you be interested in joining our chat room to discuss about it? (We currently have a private instance of mattermost, but we should probably move to some more open chat)

ogoffart:
I created issue https://github.com/sixtyfpsui/sixtyfps/issues/70 to discuss the string interpolation feature

ogoffart:
I'm going to close this PR because it is outdated, and we have string interpolation in master.

PR #90, opened 2020-10-13
Update printerdemo.60

giannissc:
Fixed alignment of the back arrow and the title


tronical:
Thanks! That looks indeed much better.

PR #99, opened 2020-11-06
vtable: VRc and VWeak pointer

ogoffart:


helper_crates/vtable/macro/macro.rs
```diff
@@ -77,6 +77,12 @@ For function type fields:
   the vtable without having a valid pointer to the actual object. But if the original function was
   marked unsafe, the unsafety is forwarded to the trait.
  - If a field is called `drop`, then it is understood that this is the destructor for a VBox.
+   It must have the type `fn(VRefMut<MyVTable>)`
+ - If a field is called `drop_in_place`, then it is understood as the in place destructor.
+   It must have the signature `fn(VRefMut<MyVTable>) -> Layout`
+   The returned layout will be passed to the deallocate function
+ - If a field is called `dealloc`, then it is a deallocation function.
+   It must have the signature `fn(&MyVTable, ptr: *mut u8, layout: Layout)`
```
tronical, line 85:
```suggestion
 - If two fields called `drop_in_place` and `dealloc` are present, then they are understood to be
    in-place destructors and deallocation functions. `drop_in_place` must have the signature
    `fn(VRefMut<MyVTable> -> Layout`, and `dealloc` must have the signature
    `fn(&MyVTable, ptr: *mut u8, layout: Layout)`.
    `drop_in_place` is responsible for destructing the object and returning the memory layout that
    was used for the initial allocation. It will be passed to `dealloc`, which is responsible for releasing
    the memory. These two functions are used to enable the use of `VRc` and `VWeak`.
```

tronical: Just a "nitpick" to emphasise that the two fields belong to together. If I misread the code and they don't have to exist at the same time, then the suggestion doesn't make sense :)

helper_crates/vtable/src/vrc.rs
```diff
@@ -0,0 +1,206 @@
...
+        .unwrap();
+        let mem = unsafe { std::alloc::alloc(layout) as *mut VRcInner<T, X> };
+        let inner = NonNull::new(mem).unwrap();
+        assert!(!mem.is_null());
+
+        unsafe {
+            mem.write(VRcInner {
+                vtable: X::static_vtable(),
+                strong_ref: Cell::new(1),
+                weak_ref: Cell::new(1),
```
tronical, line 113: Shouldn't the initial weak count be 0?

helper_crates/vtable/src/vrc.rs
```diff
@@ -0,0 +1,206 @@
...
+                    let layout = i.cast::<VRcInner<T, Layout>>().as_ref().data;
+                    T::dealloc(vtable, i.cast().as_ptr(), layout);
+                }
+            }
+        }
+    }
+}
+
+impl<T: VTableMetaDropInPlace> VWeak<T> {
+    /// Retruns the `VRc` if there is still reference count to this item.
```
tronical, line 191:
```suggestion
    /// Returns a new `VRc` if some other instance still holds a strong reference to this item.
```

helper_crates/vtable/tests/test_vrc.rs
```diff
@@ -0,0 +1,53 @@
...
+    assert!(Rc::ptr_eq(&string, &string_copy));
+    drop(string_copy);
+    let w = rc.downgrade();
+    {
+        let rc2 = w.upgrade().unwrap();
+        let string_copy = rc2.borrow().rc_string();
+        assert!(Rc::ptr_eq(&string, &string_copy));
+    }
+    drop(rc);
+    assert_eq!(Rc::strong_count(&string), 1);
```
tronical, line 51: For debugging purposes I suppose `VRc::strong_count()` would also be useful to have.

helper_crates/vtable/src/vrc.rs
```diff
@@ -0,0 +1,206 @@
...
+
+use core::cell::Cell;
+
+use super::*;
+
+/// This trait is implemented by the `#[vtable]` macro.
+///
+/// It is implemented if the macro has a "drop_in_place" function.
+pub unsafe trait VTableMetaDropInPlace: VTableMeta {
+    /// Safety: the Target needs to be pointing to a valid allocated pointer
```
tronical, line 21:
```suggestion
    /// Safety: the target ptr argument needs to be pointing to a valid allocated pointer
```

helper_crates/vtable/src/vrc.rs
```diff
@@ -0,0 +1,206 @@
...
+            inner.strong_ref.set(inner.strong_ref.get() - 1);
+            if inner.strong_ref.get() == 0 {
+                let data = &inner.data as *const u8 as *mut u8;
+                let layout = T::drop_in_place(inner.vtable, data);
+                inner.weak_ref.set(inner.weak_ref.get() - 1);
+                if inner.weak_ref.get() == 0 {
+                    let vtable = inner.vtable;
+                    T::dealloc(vtable, self.inner.cast().as_ptr(), layout);
+                } else {
+                    self.inner.cast::<VRcInner<T, Layout>>().as_mut().data = layout;
```
tronical, line 83: Just for sanity's sake, is there a way to assert/ensure that the size of T is at least as big as the size of Layout?

tronical (commented):
Looks good, a few comments/suggestions inline. The only technical thing I'm uncertain about is the initial value of the weak count.

helper_crates/vtable/src/vrc.rs
ogoffart: We allocate enough room in VRc::new.  I added a comment

helper_crates/vtable/tests/test_vrc.rs
ogoffart: Added

helper_crates/vtable/src/vrc.rs
ogoffart: 1 is correct, because VRc::drop is decreasing the weak_ref count when the strong_count reaches 0.
Added comments

tronical (approved):
Thanks, that makes sense :)

PR #100, opened 2020-11-06
vtable: change VRc and VWeak to add the concrate type as a parameter

ogoffart:
... or `Dyn` if unkown.

So we can implement Deref for the concreate type.

Sicne Deref is implemented for VRc, all the method of VRc are now
associated function.


tronical (approved):
Looks good. One minor spelling nitpick: "concrate" -> "concrete" :)

PR #136, opened 2020-12-24
Update memory example to use row and column count properties

gondolyr:
Using row and column count properties improves the readability, extensibility, and maintainability by removing the magic numbers for the other calculated properties. Previously, it took me some time to understand what these numbers represented and this change should remove this difficulty.

As this is a non-breaking refactor, the game still appears the same:

memory


tronical:
Thanks :)

PR #139, opened 2020-12-30
Add rgb function

sanxiyn:


sixtyfps_compiler/generator/cpp.rs
```diff
@@ -1313,6 +1313,7 @@ fn compile_expression(
             BuiltinFunction::Round => "[](float a){ return std::round(a); }".into(),
             BuiltinFunction::Ceil => "[](float a){ return std::ceil(a); }".into(),
             BuiltinFunction::Floor => "[](float a){ return std::floor(a); }".into(),
+            BuiltinFunction::Rgb => panic!(),
```
ogoffart, line 1316:
```suggestion
            BuiltinFunction::Rgb => "[](int r, int g, int b){ return sixtyfps::Color::from_rgb_uint8(std::clamp(r, 0, 255), std::clamp(g, 0, 255), std::clamp(b, 0, 255)); }".into(),
```

examples/rgb_slider/build.rs
```diff
@@ -0,0 +1,3 @@
+fn main() {
```
ogoffart, line 1: Every new rust / .60 files will need a copyright header (we have a test for that)

ogoffart (changes requested):
Thanks for the contribution!

Just some minor requests:
 - Add the C++ variant
 - Add license header
 - There should be a test. I recommand adding a fiew colors in `tests/cases/types/color.60`

Please let me know if you can do these changes.  Otherwise i'll just merge as is and do the fixes myself.

ogoffart:
Another point worth discuting: the CSS spec allow for both int between 0 and 255,  or a % value.
Maybe it is worth supporting something like `rgb(100%, 45%, 30%)`.  The problem is that our type system does not allow for overloads with different type yet.

tronical:
We would like to merge this and apply a few changes on top:

 * Support for percentage number literals
 * Add tests
 * Add C++ support

PR #216, opened 2021-05-02
LSP: Add Sublime Text instructions and syntax highlighting

jturcotte:


tools/lsp/README.md
```diff
@@ -118,3 +118,28 @@ exists under the `language` server section:
   }
 }
 ```
+
+## Sublime Text
+
+To setup the lsp:
+1. Install the sixtyfps-lsp binary with `cargo install` as explained in the *Generic Usage* section above.
+2. Using Package Control in Sublime Text, install the LSP package (sublimelsp/lsp)
```
ogoffart, line 126: Should that be a link?

jturcotte: You have to select it from a menu:

image

jturcotte:
Just saw that the vscode extension had syntax highlighting too and the format is much more similar, I should have started from there. I synchronized the scope names a bit.

PR #218, opened 2021-05-06
Layout refacoring

### LINKED ISSUE #105: Layout consistency
Add support for width/height/x/y on all layout types so that the layout itself behave like an element

Since layout by default occupy the whole area of the parent element, should that be the case for every element?

ogoffart:
Instead of having an `apply_layout` function in the component's vtable, use the property system to cache and propagate layout information.
Also change the code not to remove the layout element, but replace it with a Rectangle. This fixes #105

The principle is that each layout stores its constraints (max/min/preferred size) in a `LayoutInfo` property.
And each layout also stores the results of their computation in a `SharedArray<f32>` layout cache property. Then the x/y/width/height are just property with binding that reads the layout cache property.

Currently, i know that the PathLayout  with repeater is broken, but that's not the priority right now, and it was broken before.
Also this does not try to optimize the cache, which stores the x,y,w,h of each elements, instead of being smart at only storing the coordinate that matters (eg, split horizontal / vertical columns of a Grid, and don't store the orthogonal coordinate for a box)

The other cons is that this adds item and properties to the resulting binary. Hopefully later we can optimize away "useless" elements.

But this makes the code already much simpler as there are less duplication between rust and c++ and the interpreter.


ogoffart:
The `test_interpreter_image_geometry` is failling on windows because it doesn't work with the GL backend, this is not a regression, because there was no interpreter tests before for this test,  but the test is failing for me with nodejs, rust, and cpp also in master with the GL backend. The problem is that we can't compte image size with the GL backend without having a window visible :-(

I don't know why the CI is working with C++ and nodejs tests on windows on master and in this branch. This is suspicious, maybe something is wrong there and the tests are not properly failing.  Also i noticed that no rust tests are run on windows.  It could be that the code that extract the tests don't work on windows because of the `\r\n` line endings

api/sixtyfps-cpp/include/vtable.h
```diff
@@ -133,7 +133,7 @@ class VRc {
         return a.inner == b.inner;
     }
     friend bool operator!=(const VRc &a, const VRc &b) {
-        return a.inner == b.inner;
+        return a.inner != b.inner;
```
tronical, line 136: Ouch, good catch!

sixtyfps_compiler/expression_tree.rs
```diff
@@ -398,6 +388,16 @@ pub enum Expression {
     EnumerationValue(EnumerationValue),
 
     ReturnStatement(Option<Box<Expression>>),
+
+    LayoutCacheAccess {
+        layout_cache_prop: NamedReference,
+        index: usize,
+        /// When set, this is the index within a repeater, and the index is then the location of another offset.
+        /// So this looks like `layout_cache_prop[layout_cache_prop[index] + repeater_index]`
+        repeater_index: Option<Box<Expression>>,
```
tronical, line 397: Maybe it's worth expression this using an enum instead, since users have to check repeater_index anyway? Like

```rust
enum LayoutCacheIndex {
    ItemIndex(usize),
    ItemInRepeater {
        repeater_index: Box<Expression>,
        index_in_repeater: usize
    }
}
```

Not sure it's worth it though, just a though.

api/sixtyfps-cpp/include/sixtyfps.h
```diff
@@ -365,6 +364,29 @@ inline LayoutInfo LayoutInfo::merge(const LayoutInfo &other) const
                         std::min(vertical_stretch, other.vertical_stretch) };
 }
 
+/// FIXME! this should be done by cbindgen
+namespace cbindgen_private {
+inline bool operator==(const LayoutInfo &a, const LayoutInfo &b)
```
tronical, line 369: Just a thought, but while this is done manually perhaps it would be worthwhile to add a comment on the Rust side of the data structure declaration to remind anyone to make changes here when adding new fields (removal or rename will cause build breakage, good, but addition will silently "pass").

sixtyfps_compiler/expression_tree.rs
```diff
@@ -43,7 +43,7 @@ pub enum BuiltinFunction {
     ColorBrighter,
     ColorDarker,
     Rgb,
-    ImplicitItemSize,
+    ImplicitLayoutInfo,
```
tronical, line 46: With ImplicitSize I can see the counter part being the "set" size. With the LayoutInfo, is there a counter-part as well?

AFAICS the implementation basically calls `layouting_info()` on the item vtable, so perhaps `ItemLayoutInfo` is cleaner.

Anyway, just a nitpick - feel free to ignore this one :)

sixtyfps_compiler/passes/materialize_fake_properties.rs
```diff
@@ -77,3 +82,34 @@ fn has_declared_property(elem: &Element, prop: &str) -> bool {
         _ => false,
     }
 }
+
+/// Initialize a sensible default binding for the now materialized property
+fn initialize(elem: ElementRc, name: &str) {
+    let expr = match name {
+        "minimum_height" => layout_constraint_prop(&elem, "min_height"),
```
tronical, line 89: Perhaps either these two functions or just the default binding definitions could go into the `typeregister.rs` or somewhere else, just so that the definition of our reserved properties and associated details like defaults are in one place. This doesn't necessarily belong into this change set though.

sixtyfps_runtime/corelib/items.rs
```diff
@@ -79,9 +79,6 @@ pub struct ItemVTable {
     pub layouting_info:
         extern "C" fn(core::pin::Pin<VRef<ItemVTable>>, window: &ComponentWindow) -> LayoutInfo,
 
-    pub implicit_size:
```
tronical, line 82: Nice to see this go :)

sixtyfps_compiler/expression_tree.rs
ogoffart: yes, when `maximum-width` or `minimum-height` or so are set, that's the explicit.

sixtyfps_compiler/expression_tree.rs
ogoffart: Yeah, that would be nicer. but would force a different enum and more matching.  I can see that this structure will change in the future if we get smarter about how the cache is represented.

sixtyfps_runtime/interpreter/eval_layout.rs
```diff
@@ -0,0 +1,336 @@
...
+                    .height_reference
+                    .as_ref()
+                    .map(expr_eval)
+                    .unwrap_or(0.),
+                x: 0.,
+                y: 0.,
+                spacing,
+                padding: &padding,
+                cells: Slice::from(cells.as_slice()),
+            }))
```
tronical, line 89: <nitpick>For consistency I think using `.into()` here instead of `Value::LayoutCache(...` is nicer :-)</nitpick>

tests/cases/image_geometry.60
```diff
@@ -45,14 +45,15 @@ TestCase := Rectangle {
     property <bool> image_in_layout_custom_fit_ok: image_in_layout_with_explicit_fit.image-fit == ImageFit.fill;
     property <bool> image_with_missing_height_ok <=> image_with_missing_height.expected_height_ok;
     property <bool> image_with_missing_width_ok <=> image_with_missing_width.expected_width_ok;
+
+    property <bool> test: fixed_image_default_image_fit_ok && fixed_image_image_fit_override_ok && image_in_layout_fit_ok && image_in_layout_custom_fit_ok && image_with_missing_height_ok && image_with_missing_width_ok;
```
tronical, line 49: There's something for future assert syntax perhaps, that we could express here, use from the interpreter (as opposed to querying for the all-combining test property) as well as from the generated code.

tools/syntax_updater/from_0_0_6.rs
```diff
@@ -0,0 +1,51 @@
...
+        ) {
+            if let Some(elem) =
+                find_parent_element(binding.into()).and_then(|x| find_parent_element(x.into()))
+            {
+                let elem_name = elem
+                    .QualifiedName()
+                    .map(|qn| object_tree::QualifiedTypeName::from_node(qn).to_string())
+                    .unwrap_or_default();
+                if !matches!(elem_name.as_str(), "GridLayout" | "Row") {
+                    write!(file, "/* {} // REMOVED BY THE SYNTAX UPDATOR */", node.to_string())?;
```
tronical, line 35: Nitpick: updator -> updater :)

tronical (commented):
Some initial comments. Looks like a definite improvement so far - especially on the run-time side. Missing is lower_layout and the two generators. The interpreter bit looks good.

sixtyfps_compiler/passes/lower_layout.rs
```diff
@@ -49,246 +123,185 @@ fn lower_grid_layout(
...
         }
     }
-    component.optimized_elements.borrow_mut().push(layout_element.clone());
-    if !layout.elems.is_empty() {
-        Some(layout.into())
-    } else {
-        None
-    }
+    layout_element.borrow_mut().children = layout_children;
+    layout_cache_prop.element().borrow_mut().bindings.insert(
```
tronical, line 217: Perhaps if the `lower_$type_layout()` functions returned a `Layout`, then the binding insertion code for both properties could be moved to the caller side and de-duplicated.

tronical (approved):

sixtyfps_compiler/expression_tree.rs
tronical: Ok :)

sixtyfps_compiler/passes/lower_layout.rs
ogoffart: Yeah i actually thought about that, but then i also need to pass the layout_cache_prop and layout_info_prop to the function.

tests/cases/image_geometry.60
ogoffart: yeah, but that's not yet implemented.
But when this fails, the test driver will print out all the property to the debug output so we'll know what the problem is.

tools/syntax_updater/from_0_0_6.rs
ogoffart: Note: I was thinking perhaps we should print a warning from there.
Another thing that could be done is to add something like  `@warning("Syntax updater removed: colspan: 2")`

PR #220, opened 2021-05-10
Fix crashes on Windows when using the C++ build and DLLs

tronical:


sixtyfps_runtime/corelib/items.rs
```diff
@@ -57,6 +57,22 @@ type ItemRendererRef<'a> = &'a mut dyn crate::item_rendering::ItemRenderer;
 pub type VoidArg = ();
 type KeyEventArg = (KeyEvent,);
 
+#[macro_export]
+macro_rules! declare_item_vtable {
+    (fn $getter:ident() -> $item_vtable_ty:ident for $item_ty:ty) => {
+        ItemVTable_static! {
+            /// The VTable for `$ty`
```
ogoffart, line 64: the `$ty` will not work like this.  You need to use concat! in a #[doc] attribute.

sixtyfps_runtime/corelib/items.rs
```diff
@@ -57,6 +57,22 @@ type ItemRendererRef<'a> = &'a mut dyn crate::item_rendering::ItemRenderer;
...
+            #[no_mangle]
+            pub static $item_vtable_ty for $item_ty
+        }
+        #[no_mangle]
+        pub extern "C" fn $getter() -> *const ItemVTable {
```
ogoffart, line 69: Should we guard that against the #[cfg(feature="ffi")] ?  Also maybe #[cfg(platform="windows")]

sixtyfps_runtime/corelib/items.rs
```diff
@@ -57,6 +57,22 @@ type ItemRendererRef<'a> = &'a mut dyn crate::item_rendering::ItemRenderer;
 pub type VoidArg = ();
 type KeyEventArg = (KeyEvent,);
 
+#[macro_export]
+macro_rules! declare_item_vtable {
+    (fn $getter:ident() -> $item_vtable_ty:ident for $item_ty:ty) => {
```
ogoffart, line 62: Maybe with concat! and  #[link_name] it is possible to make this macro not take a $getter.

For example
```rust
#[link_name = concat!("sixtyfps_get_", stringify!($item_vtable_ty))]
fn $item_vtable_ty(){ ... }
```

But i think that's probably not worth it, and the way it is now is fine.

I actually like this syntax with `fn` and a return value.

ogoffart (approved):

sixtyfps_runtime/corelib/items.rs
tronical: Yes! Thanks. Will do.

sixtyfps_runtime/corelib/items.rs
tronical: Ohh yes that’s better. I’ll give it a try, it might improve it.

sixtyfps_compiler/langtype.rs
```diff
@@ -501,10 +501,13 @@ pub struct NativeClass {
 
 impl NativeClass {
     pub fn new(class_name: &str) -> Self {
-        let vtable_symbol = format!("{}VTable", class_name);
+        let cpp_vtable_getter = format!(
+            "sixtyfps::private_api::get_vtable(&sixtyfps::private_api::{}VTable, &sixtyfps::private_api::sixtyfps_get_{}VTable)",
```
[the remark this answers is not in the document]
tronical: You’re right, that’s the logical improvement. Will do.

tronical:
Thanks for the review. I’ll merge this and do follow up fixes in the main branch (especially the feature guard)

sixtyfps_compiler/langtype.rs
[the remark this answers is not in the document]
tronical: Done in c6a4841cd4da20037652c459d890214c552874ca

sixtyfps_runtime/corelib/items.rs
tronical: Done in 5d6c8e7ee1b4f49076f145ad40354978f2ecaaa2

PR #229, opened 2021-05-19
GL backend: new box drop shadow implementation

tronical:
This replaces the box gradient with a new implementation that
behaves like the HTML Canvas element, by applying a gaussian
blur to the alpha of the rectangle to shadow.

The drop-shadow-blur property is changed to be now just a "level", like
the HTML Canvas shadowBlur property, which is defined to be half of the
standard deviation of the gaussian blur to be applied.

### DIFF
--- a/sixtyfps_runtime/corelib/Cargo.toml
+++ b/sixtyfps_runtime/corelib/Cargo.toml
@@ -42,7 +42,7 @@ weak-table =  "0.3"
 scopeguard = "1.1.0"
 cfg-if = "1"
 # Optional dependency to have conversion trait between femtovg::Color and sixtyfps-corelib::Color
-femtovg = { version = "0.1.3", optional = true }
+femtovg = { version = "0.2.0", optional = true, git = "https://github.com/tronical/femtovg", branch = "simon/filter" }
 
 [target.'cfg(target_arch = "wasm32")'.dependencies]
 instant = { version = "0.1", features = [ "wasm-bindgen", "now" ] }
--- a/tests/cases/examples/box_shadow_use.60
+++ b/tests/cases/examples/box_shadow_use.60
@@ -27,10 +27,11 @@ TestCase := Window {
 
         Rectangle {
             background: blue;
+            border-radius: 10px;
             drop-shadow-offset-x: 10px;
             drop-shadow-offset-y: 10px;
             drop-shadow-color: #00000080;
-            drop-shadow-blur: 5px;
+            drop-shadow-blur: 5;
         }
 
         for r[i] in [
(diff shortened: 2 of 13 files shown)


sixtyfps_runtime/corelib/Cargo.toml
```diff
@@ -42,7 +42,7 @@ weak-table =  "0.3"
 scopeguard = "1.1.0"
 cfg-if = "1"
 # Optional dependency to have conversion trait between femtovg::Color and sixtyfps-corelib::Color
-femtovg = { version = "0.1.3", optional = true }
+femtovg = { version = "0.2.0", optional = true, git = "https://github.com/tronical/femtovg", branch = "simon/filter" }
```
ogoffart, line 45: ideally, we should remove this dependency.  The conversion can simply be done with an helper method.  (But this does not have to be done now)

tests/cases/examples/box_shadow_use.60
```diff
@@ -27,10 +27,11 @@ TestCase := Window {
 
         Rectangle {
             background: blue;
+            border-radius: 10px;
             drop-shadow-offset-x: 10px;
             drop-shadow-offset-y: 10px;
             drop-shadow-color: #00000080;
-            drop-shadow-blur: 5px;
+            drop-shadow-blur: 5;
```
ogoffart, line 34: The updater test will fail.  Maybe the updater can wrap expression of `drop-shadow-blur` in a `(...) / 1px`

tronical:
For comparison, here are before and after screenshots:

before

after

ogoffart (approved):

sixtyfps_runtime/corelib/Cargo.toml
tronical: You're right, as "convenient" as it is, making updates here is becoming annoying :). I'll make this a separate cleanup later.

tests/cases/examples/box_shadow_use.60
tronical: Ah yes, well spotted! Good idea.

tools/syntax_updater/from_0_0_6.rs
```diff
@@ -20,22 +20,39 @@ pub(crate) fn fold_node(
...
+                        write!(
+                            file,
+                            "/* {} // REMOVED BY THE SYNTAX UPDATER */",
+                            node.to_string()
+                        )?;
+                        return Ok(true);
+                    }
                 }
             }
+            prop @ "drop-shadow-blur" => {
```
tronical, line 42: @ogoffart Is this the correct approach?

ogoffart: yes, maybe also match "drop_shadow_blur"

tools/syntax_updater/from_0_0_6.rs
```diff
@@ -20,22 +20,39 @@ pub(crate) fn fold_node(
...
+                let expr = binding.BindingExpression();
+                write!(
+                    file,
+                    "{}: ({}) / 1px; // CHANGED BY SYNTAX UPDATER: drop-shadow-blur is now an abstract blur level, this is just an approximation.",
```
ogoffart, line 46: you should use /* */ comment  in case the property binding is not at the end of the line.

tronical: Good point, thanks

tools/syntax_updater/from_0_0_6.rs
tronical: Good idea

ogoffart:
:+1:

sixtyfps_runtime/corelib/Cargo.toml
peter-todorov: It would be nice if there's a universal "color" crate that we can all use.

PR #231, opened 2021-05-20
Optimisation of constant bindings

ogoffart:
By using Property::set  instead of Property::set_binding for bindings with constant values.

### DIFF
--- a/sixtyfps_compiler/expression_tree.rs
+++ b/sixtyfps_compiler/expression_tree.rs
@@ -121,6 +121,34 @@ impl BuiltinFunction {
             }
         }
     }
+
+    /// It is pure if the return value only depends on its argument and has no side effect
+    fn is_pure(&self) -> bool {
+        match self {
+            BuiltinFunction::GetWindowScaleFactor => false,
+            // Even if it is not pure, we optimize it away anyway
+            BuiltinFunction::Debug => true,
+            BuiltinFunction::Mod
+            | BuiltinFunction::Round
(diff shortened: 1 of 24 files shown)


sixtyfps_compiler/generator.rs
```diff
@@ -128,3 +133,62 @@ pub fn build_array_helper(component: &Component, mut visit_item: impl FnMut(&Ele
         }
     }
 }
+
+/// Will wall the `handle_property` callback for every property that needs to be initialized.
```
tronical, line 137:
```suggestion
/// Will call the `handle_property` callback for every property that needs to be initialized.
```

tronical:
Nice! The printerdemo goes from 2090 `set_binding()` calls down to 1665.

sixtyfps_compiler/expression_tree.rs
```diff
@@ -976,11 +1012,22 @@ impl Expression {
         }
     }
 
+    /// Try to mark this expression to a lvalue that can be assigned to.
+    ///
     /// Return true if the expression is a "lvalue" that can be used as the left hand side of a `=` or `+=` or similar
-    pub fn is_rw(&self) -> bool {
+    pub fn try_set_rw(&mut self) -> bool {
```
tronical, line 1018: Not in scope of this request: I find `_rw` a bit of a cryptic suffix :) . Maybe this could be called `try_mark_for_assignment` ?

tronical (approved):

PR #234, opened 2021-05-20
Add support for specifying the viewbox of path elements

tronical:


docs/builtin_elements.md
```diff
@@ -227,6 +227,10 @@ accordingly.
 * **`stroke-width`** (*length*): The width of the outline.
 * **`width`** (*length*): If non-zero, the path will be scaled to fit into the specified width.
 * **`height`** (*length*): If non-zero, the path will be scaled to fit into the specified height.
+* **`viewbox-x`**/**`viewbox-y`**/**`viewbox-width`**/**`viewbox-height`** (*float*) These four properties allow
+  defining the position and size of the viewport of the path in path coordinates. In the rendered output, the
+  If the `viewport-width` or `viewport-height` is less or equal than zero, the viewbox properties are ignored
```
ogoffart, line 232: viewport?

examples/iot-dashboard/iot-dashboard.60
```diff
@@ -118,6 +118,11 @@ PieChartPainted := Rectangle {
         height: 100%;
         stroke-width: parent.width / 6;
         stroke: #aaaaaa40;
+        viewbox-x: 0;
+        viewbox-y: 0;
```
ogoffart, line 122: do we need to specify the 0?

ogoffart (commented):
Great!

Also missing an entry in the changelog

ogoffart:
What happens if we draw outside of the viewbox ?

docs/builtin_elements.md
tronical: Ooops

examples/iot-dashboard/iot-dashboard.60
tronical: Indeed that’s not necessary

tronical:
Ah I forgot to git add the Changelog bit.

I think I need to amend the change to add clipping of the viewbox intersects with the bounding box of the path.

ogoffart:
> I think I need to amend the change to add clipping of the viewbox intersects with the bounding box of the path

Maybe only if clip is true.
We probably need to special case  clip on path to put it as a parent instead of as a child

tronical:
Good idea to re-use the `clip` property. I'll do that next.

ogoffart (approved):

docs/builtin_elements.md
```diff
@@ -231,6 +231,9 @@ accordingly.
   defining the position and size of the viewport of the path in path coordinates. In the rendered output, the
   If the `viewbox-width` or `viewbox-height` is less or equal than zero, the viewbox properties are ignored
   and instead the bounding rectangle of all path elements are used to define the view port.
+* **`clip`** (*bool*): By default, when a path has a view box defined and the elements render outside of it, they are still
```
ogoffart, line 234: Is it necessary to document this here since it is a property of all items?

ogoffart: ah, it is actually not a property on all items, only on rectangle.  then it make sense

ogoffart (approved):

PR #240, opened 2021-05-22
Add the Rust equivalent of sixtyfps::invoke_from_event_loop

tronical:
The same API exists in C++.

### DIFF
--- a/api/sixtyfps-rs/lib.rs
+++ b/api/sixtyfps-rs/lib.rs
@@ -87,7 +87,7 @@ fn main() {
 }
 ```
 
-### Generated components
+## Generated components
 
 As of now, only the last component of a .60 source is generated. It is planned to generate all exported components.
 
@@ -97,10 +97,10 @@ For example, if you have `export MyComponent := Window { /*...*/ }` in the .60 f
 This documentation contains a documented generated component: [`docs::generated_code::SampleComponent`].
 
 The component is created using the [`fn new() -> Self`](docs::generated_code::SampleComponent::new) function. In addition
-the following convenience functions are available through the [ComponentHandle] implementation:
+the following convenience functions are available through the [`ComponentHandle`] implementation:
 
-  - [`fn clone_strong(&self)`](docs::generated_code::SampleComponent::clone_strong): to create a strongly referenced clone.
-  - [`fn as_weak(&self)`](docs::generated_code::SampleComponent::as_weak): to create a weak reference.
+  - [`fn clone_strong(&self) -> Self`](docs::generated_code::SampleComponent::clone_strong): to create a strongly referenced clone.
+  - [`fn as_weak(&self) -> Weak`](docs::generated_code::SampleComponent::as_weak): to create a [weak](Weak) reference.
   - [`fn show(&self)`](docs::generated_code::SampleComponent::show): to show the window of the component.
   - [`fn hide(&self)`](docs::generated_code::SampleComponent::hide): to hide the window of the component.
   - [`fn run(&self)`](docs::generated_code::SampleComponent::run): a convenience function that first calls `show()`,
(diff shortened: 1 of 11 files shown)


tronical:
I should add a unit test as a follow up. Just wanted to make this available for @flukejones first

tools/lsp/Cargo.toml
```diff
@@ -21,6 +21,7 @@ sixtyfps-compilerlib = { version = "=0.0.6", path = "../../sixtyfps_compiler"}
 sixtyfps-interpreter = { version = "=0.0.6", path = "../../sixtyfps_runtime/interpreter"}
 sixtyfps-corelib = { version = "=0.0.6", path = "../../sixtyfps_runtime/corelib"}
 sixtyfps-rendering-backend-default = { version = "=0.0.6", path="../../sixtyfps_runtime/rendering_backends/default" }
+sixtyfps = { version = "=0.0.6", path = "../../api/sixtyfps-rs"}
```
ogoffart, line 24: Not sure how I feel about this, this add quite some dependency to the lsp, while we will never get rid of the private dependencies anywy

tools/viewer/Cargo.toml
```diff
@@ -9,9 +9,8 @@ repository = "https://github.com/sixtyfpsui/sixtyfps"
 homepage = "https://sixtyfps.io"
 
 [dependencies]
-sixtyfps-corelib = { version = "=0.0.6", path="../../sixtyfps_runtime/corelib" }
-sixtyfps-rendering-backend-default = { version = "=0.0.6", path="../../sixtyfps_runtime/rendering_backends/default" }
 sixtyfps-interpreter = { version = "=0.0.6", path = "../../sixtyfps_runtime/interpreter", features = ["display-diagnostics"] }
+sixtyfps = { version = "=0.0.6", path="../../api/sixtyfps-rs" }
```
ogoffart, line 13: Don't we already have this API in the interpreter?

ogoffart:
Maybe we should rename and document it in the corelib, and re-export it in both the interpreter and in the sixtyfps-rs crate.

flukejones:
I assume that a blocking task would block the whole app? If so, maybe that could be added to doc comments?

flukejones:
Might also be worth noting that a standard `Arc<Mutex<App>>` can't be used in this as the root construct isn't send or sync.

ogoffart:
> I assume that a blocking task would block the whole app? 

Yes, We could also add an API to run a future in the ui event loop.

ogoffart:
The problem with this API, as mentioned on mattermost, is that it won't allow to safely capture the handle to the window an re-use it from the main thread.
We need to come up with a better way.

Some idea:
```rust
let handle = MyApp::new();
let (sx, rx) = async_channel::unbounded();
sixtyfps::run_future(move async { 
  while let Ok(msg) = rx.recv().await { 
     // do something with msg and handle
   }
});
// ... move  sx  to a thread. and send messages
```

Another idea

```rust
let handle = MyApp::new();
let handle = sixtyfps::ThreadHandle::new(handle);
// handle is Send + Sync

//... send the handle to a thread, and from the thread do something like
handle.send_event(move |handle: &MyApp| handle.set_foo_property(foo_bar));
```

We could also try to get the ComponentHandle somehow Send, but that's a bigger task involving reviewing the whole architecture: https://github.com/sixtyfpsui/sixtyfps/issues/54   But maybe it's worth it to take threading more seriously because real life app should do most of the work in another thread.

tronical:
🤦‍♂️ indeed, this API is not very useful on Rust.

I very much like the ThreadHandle approach. A variant of it could be that perhaps it’s really a handle into a vec arena in tls. So it could be sent around but only in the ui thread it will return Some(vrc) and otherwise none.

ogoffart:
Well, one can also use this API by storing the window handle in a thread_local.

ThreadHandle can just wrap a Rc and we can mark it a Sync manually. But we have to make sure that it is indeed created in the same thread as the one that is running the event loop.

tronical:
We could add a std::thread::ThreadId to VWeak, initialize it appropriately, check it against the current thread id on upgrade() and then mark VWeak as Send maybe?

ogoffart:
Right, we should make the `sixtyfps::Weak`  Send,  and the `upgrade()` can check if it is in the right thread and otherwise return None.
I'll work on that.

api/sixtyfps-rs/lib.rs
```diff
@@ -247,9 +263,40 @@ pub fn run_event_loop() {
         .run_event_loop(sixtyfps_corelib::backend::EventLoopQuitBehavior::QuitOnLastWindowClosed);
 }
 
-/// This trait describes the common public API of a strongly referenced SixtyFPS component,
-/// held by a [vtable::VRc]. It allows creating strongly-referenced clones, a conversion into
-/// a weak pointer as well as other convenience functions.
+/// Adds the specified boxed function to an internal queue, notifies the event loop to wake up.
```
tronical, line 266: I guess the function is not boxed anymore - or that's an implementation detail :)

api/sixtyfps-rs/lib.rs
```diff
@@ -247,9 +263,40 @@ pub fn run_event_loop() {
...
+/// ```rust
+/// sixtyfps::sixtyfps! { MyApp := Window { property <int> foo; /* ... */ } }
+/// let handle = MyApp::new();
+/// let handle_weak = handle.as_weak();
+/// let thread = std::thread::spawn(move || {
+///     // ... Do some computation in the thread
+///     let foo = 42;
+///      // now forward the data to the main thread using invoke_from_event_loop
+///     let handle_copy = handle_weak.clone();
+///     sixtyfps::invoke_from_event_loop(move || handle_copy.unwrap().set_foo(foo));
```
tronical, line 286:
```suggestion
///     sixtyfps::invoke_from_event_loop(move || handle_copy.upgrade().unwrap().set_foo(foo));
```

api/sixtyfps-rs/lib.rs
```diff
@@ -281,40 +328,71 @@ pub trait ComponentHandle {
...
+    /// as this would produce a reference loop and leak the component.
+    /// Instead, the callback function should capture a weak component.
+    ///
+    /// The Weak component also implement `Send` and can be send to another thread.
+    /// but the upgrade function will only return a valid component from the same thread
+    /// as the one it has been created from.
+    /// This is usefull to use with [`invoke_from_event_loop()`].
+    pub struct Weak<T: ComponentHandle> {
+        inner: vtable::VWeak<re_exports::ComponentVTable, T::Inner>,
+        thread: std::thread::ThreadId,
```
tronical, line 349: I wonder if perhaps the order of these two should be swapped, so that on 32-bit systems the bigger field (64-bit thread id) comes first?

tronical (commented):
Looks good to me. (GH ui won't let me approve because that would be self-approval 🤣 )

PR #247, opened 2021-05-28
sixtyfps::Image API in C++ and Rust

ogoffart:


sixtyfps_runtime/corelib/graphics/image.rs
```diff
@@ -0,0 +1,69 @@
...
+}
+
+/// Error generated if an image cannot be loaded for any reasons.
+#[derive(Default, Debug, PartialEq)]
+pub struct LoadImageError(());
+
+/// An image type that can be displayed by the Image element
+#[repr(transparent)]
+#[derive(Default, Clone, Debug, PartialEq)]
+// FIXME: the inner should be private
```
tronical, line 45: Ah, that's public because of the interpreter/nodejs that uses `.0`?

Perhaps this could be solved using a getter on Image, like

```rust
    pub fn as_path(&self) -> Option<SharedString>
```

tronical (approved):

sixtyfps_runtime/corelib/graphics/image.rs
ogoffart: Yes, still a lot of code uses the .0
Also to create the raw data which we probably do not want public API for

tronical: As long as those are unambiguous from/into makes sense. I kind of like a function because it increases the readability of the code. Instead of "converting" an Image to a SharedString (what is that?), calling `as_path()` makes it clearer that a path is extracted (regardless of the string as return type).

PR #248, opened 2021-05-28
Fix disappearing text in vertical layouts with word-wrap and alignment

### LINKED ISSUE #246: alignment in vertical layout makes text with word wrap disappear
In the following test-case the text should be vertically centered and broken into multiple lines:

```
MainWindow := Window{
    width: 400px;
    height: 640px;
    VerticalLayout {
        alignment: center;
        Text {
            text: "This line needs to be broken into multiple lines of text.";
            font_size: 20px;
            wrap: word-wrap;
        }
    }    
}
```

tronical:
When a Text element has wrapping enabled, it should not have zero
minimum size. Otherwise the layout will give it a height of zero in the
layout in the test case and that'll make the text disappear. There are
different options but this patch goes for minimum height as if no
wrapping was enabled (so at least one line plus forced line breaks).

Fixes #246

### DIFF
--- a/sixtyfps_runtime/corelib/items/text.rs
+++ b/sixtyfps_runtime/corelib/items/text.rs
@@ -143,7 +143,7 @@ impl Item for Text {
         };
         match self.overflow() {
             TextOverflow::elide => {
-                min_size.width = font_metrics.text_size("…").width;
+                min_size.width = min_size.width.min(font_metrics.text_size("…").width);
             }
             TextOverflow::clip => {}
         }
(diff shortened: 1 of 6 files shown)


sixtyfps_runtime/corelib/items/text.rs
```diff
@@ -125,33 +125,37 @@ impl Item for Text {
...
-                min_width: min_size.width.ceil(),
-                min_height: min_size.height.ceil(),
-                preferred_width: implicit_size.width.ceil(),
-                preferred_height: implicit_size.height.ceil(),
-                ..LayoutInfo::default()
+            implicit_size
+        };
+        match self.overflow() {
+            TextOverflow::elide => {
+                min_size.width = font_metrics.text_size("…").width;
```
ogoffart, line 146:
```suggestion
                min_size.width = min_size.width.min(font_metrics.text_size("…").width);
```

ogoffart (approved):

Issue #111, opened 2020-11-23
Signal forwarding


ogoffart:
Usually, one want to just forward a signal from the component to the parent.
Suggestions:

 * Do nothing: so we have to declare a signal and forward it:

```
App := Window {
   signal button_clicked;
   Button { clicked => { root.button_clicked();  } }
}
```
Perhaps the syntax could be made easier by allowing to omit the `{ }`? 

 * Use the same syntax as for properties:

```
App := Window {
   signal button_clicked <=> but.clicked;
   but := Button {  }
}
// or
App := Window {
   signal button_clicked;
   but := Button { clicked <=> root.button_clicked  }
}
```

 * Somehow have a new syntax to automatically create a signal on the root

```
App := Window {
    Button {  
        // automatically create a signal
        forward clicked as button_clicked;
    }
}
```

jturcotte:
From a quick glance I like `signal button_clicked <=> but.clicked;`. It's consistent with the two-way properties and allows renaming the signal.

Just `clicked <=> root.button_clicked` looks a bit confusing.

One thing that makes it weird though is that the signal isn't two ways, so the correct way to express it maybe be `signal button_clicked <= but.clicked;` ?

jturcotte:
> Perhaps the syntax could be made easier by allowing to omit the { }?

About this, since you basically want to assign a function as a function property (signal handler), it could maybe be simplified further by removing parentheses too:

```
App := Window {
   signal button_clicked;
   Button { clicked: root.button_clicked }
}
```

Or maybe:
```
App := Window {
   signal button_clicked;
   Button { clicked => root.button_clicked }
}
```

If you ever allow native functions to be imported in the .60 file, the same syntax could be used to assign the native function directly as a signal handler instead of wrapping it with a pass-through lambda expression (if the signature allows it, the lambda at least allows adapting a callee with a slightly different signature).

tronical:
Subsequent discussion produced another angle. Signal forwarding here might serve the purpose of addressing a bigger problem:

    It may be desirable to write re-usable compoments in `.60` where some aspects are handled by native code. A "Piano Key" would be an example, that when pressed should use system APIs to produce the correct note. All the information to accomplish the task may be available inside the "Piano Key" and the API for producing the sound may not require any additional state, i.e. just have a "beep(frequency)" type of global function.

[...]

PR #256, opened 2021-06-07
Callback aliases

ogoffart:
Helps with #111


sixtyfps_runtime/interpreter/dynamic_component.rs
```diff
@@ -439,9 +439,38 @@ impl<'id> ComponentDescription<'id> {
...
+            .root_element
+            .borrow()
+            .property_declarations
+            .get(name)
+            .and_then(|d| d.is_alias.as_ref())
+        {
+            generativity::make_guard!(guard);
+            // Safety: we just verified that the component has the right vtable
+            let c = unsafe { InstanceRef::from_pin_ref(component, guard) };
+            generativity::make_guard!(guard);
```
tronical, line 453: Is there one guard too many?

tronical (approved):

sixtyfps_runtime/interpreter/dynamic_component.rs
ogoffart: the first guard is moved in the previous line and cannot be re-used.  I could have named them differently, but i thought it would not be necessary

PR #258, opened 2021-06-08
Add a pass to always have the top level as a Window

ogoffart:


sixtyfps_compiler/passes/ensure_window.rs
```diff
@@ -0,0 +1,99 @@
...
+use crate::expression_tree::Expression;
+use crate::langtype::Type;
+use crate::namedreference::NamedReference;
+use crate::object_tree::{Component, Element};
+use crate::typeregister::TypeRegister;
+use std::cell::RefCell;
+use std::collections::HashSet;
+use std::rc::Rc;
+
+pub fn ensure_window(component: &Rc<Component>, type_register: &TypeRegister) {
```
tronical, line 21: I had to read this function a few times :) and it makes sense to me. I’m just wondering about one detail: properties declared in new_root seem like they would not be access I through the API (interpreter/C++/Rust) since they remain in new_root. Am I missing something perhaps?

ogoffart: no, properties are not moved, so they remain in what becomes the window

tronical: Ah yes, of course they remain in win_elem

tronical (approved):

PR #261, opened 2021-06-15
split vertical and horizontal pass in layouts

ogoffart:
So that we can do 'height for width' later.

This implement the interpreter and the rust code, but the C++ is not done yet.

### DIFF
--- a/sixtyfps_compiler/passes/default_geometry.rs
+++ b/sixtyfps_compiler/passes/default_geometry.rs
@@ -20,6 +20,7 @@ use std::rc::Rc;
 
 use crate::diagnostics::BuildDiagnostics;
 use crate::langtype::DefaultSizeBinding;
+use crate::langtype::EnumerationValue;
 use crate::langtype::Type;
 use crate::object_tree::{Component, ElementRc};
 use crate::{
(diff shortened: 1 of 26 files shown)


tronical (commented):
Looks good to me

sixtyfps_compiler/passes/default_geometry.rs
```diff
@@ -199,19 +213,11 @@ fn make_default_100(elem: &ElementRc, parent_element: &ElementRc, property: &str
     });
 }
 
-fn make_default_implicit(elem: &ElementRc, property: &str) {
+fn make_default_implicit(elem: &ElementRc, property: &str, orientation: Orientation) {
```
tronical, line 216: Just a minor thought: I suppose this function could be changed to just take the orientation (and element of course) and derive the property name from the Orientation, since it's always either width or height.

sixtyfps_runtime/corelib/layout.rs
```diff
@@ -82,30 +72,24 @@ impl LayoutInfo {
...
-                other.preferred_height,
-            ),
-            horizontal_stretch: self.horizontal_stretch.min(other.horizontal_stretch),
-            vertical_stretch: self.vertical_stretch.min(other.vertical_stretch),
+            stretch: self.stretch.min(other.stretch),
         }
     }
+
+    /// Helper function to return a preferred size which is within the min/max constraints
+    pub fn preferred(&self) -> f32 {
```
tronical, line 90: Hn, so `info.preferred` and `info.preferred()` look really alike but they evaluate to something rather different. Maybe the function should be called `preferred_constrained()` or `constrained_preferred_size()` or `bounded_preferred_size()`.

tests/cases/layout/box_precentages.60
```diff
@@ -20,7 +20,6 @@ TestCase := Rectangle {
             rect1 := Rectangle {
                 background: red;
                 width: 10%;
-                height: 20%;
```
tronical, line 23: Just curious, what happened here?

tronical (approved):

tests/cases/layout/box_precentages.60
ogoffart: Before, the red box was smaller than the blue box.  
(see https://sixtyfps.io/editor/preview.html?load_url=https://raw.githubusercontent.com/sixtyfpsui/sixtyfps/master/tests/cases/layout/box_precentages.60 )
But i think that was wrong.

sixtyfps_compiler/passes/default_geometry.rs
ogoffart: there is only two call to this internal function. So i think it's fine because we would have to do a match there instead of just specifying it.

sixtyfps_runtime/corelib/layout.rs
ogoffart: I'll rename it

tronical:
The CI failure is my fault and is fixed in master. A rebase should do the trick or we merge anyway :)

tests/cases/layout/box_precentages.60
tronical: Ok :). Ah, and I see that that still works, in case somebody wants that behaviour.

PR #265, opened 2021-06-18
Add Image::size()

tronical:
This requires the image size query to be window independent and go
through the backend instead.

This implies minor changes for the Qt backend and bigger ones for the GL
backend:

* There exists now a thread-local IMAGE_CACHE, which is used by the
  backend's image_size() function as well as by the renderer for
  loading CPU side images.
* The image remain as decoded images in there (including SVG tree)
  and the window now has a texture_cache, which holds CachedImage
  with ImageData::Texture.
* Rendering an image item therefore fetches the CPU side image,
  calls upload_to_gpu() on it, which creates a new Rc<CachedImage>
  and that's stored in the texture_cache.
* The texture cache continues to be pruned when hiding the window.

### DIFF
--- a/api/sixtyfps-cpp/tests/datastructures.cpp
+++ b/api/sixtyfps-cpp/tests/datastructures.cpp
@@ -13,6 +13,7 @@ LICENSE END */
 #include "catch2/catch.hpp"
 
 #include <sixtyfps.h>
+#include <sixtyfps_image.h>
 
 SCENARIO("SharedString API")
 {
@@ -72,3 +73,24 @@ TEST_CASE("Property Tracker")
     REQUIRE(!tracker1.is_dirty());
 }
 
+TEST_CASE("Image")
+{
+    using namespace sixtyfps;
+
+    // ensure a backend exists, using private api
+    private_api::ComponentWindow wnd;
+
+    Image img;
+    {
+        auto size = img.size();
+        REQUIRE(size.width == 0.);
+        REQUIRE(size.height == 0.);
+    }
+
+    img = Image::load_from_path(SOURCE_DIR "/../../vscode_extension/extension-logo.png");
+    {
+        auto size = img.size();
+        REQUIRE(size.width == 128.);
+        REQUIRE(size.height == 128.);
+    }
+}
--- a/api/sixtyfps-cpp/include/sixtyfps_image.h
+++ b/api/sixtyfps-cpp/include/sixtyfps_image.h
@@ -15,6 +15,20 @@ LICENSE END */
 
 namespace sixtyfps {
 
+#if !defined(DOXYGEN)
+using cbindgen_private::types::Size;
+#else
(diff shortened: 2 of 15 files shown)


ogoffart:
Looks good.
There should also be a similar API in C++

tronical:
ACK. And tests for both.

tronical:
I added the C++ API and a test for it. For the Rust image a unit test breaks because of the backend selection problem.

Generally, we could make the public sixtyfps::Image API have an backend::get_or_init() type of call like we have for the component(window). In C++ that would solve the problem, and maybe with Rust we need the same, i.e. a public sixtyfps::Image in the rust API crate that wraps ImageInner (or PrivateImage) *and* can do the backend::get_or_init() dance.

That way we could get rid of the panic() and make the API overall more robust. This is naturally not an image specific problem but applies to any type in the public Rust API that we just re-export but that needs this kind of "insurance".

api/sixtyfps-cpp/tests/datastructures.cpp
```diff
@@ -13,6 +13,7 @@ LICENSE END */
 #include "catch2/catch.hpp"
 
 #include <sixtyfps.h>
+#include <sixtyfps_image.h>
```
ogoffart, line 16: Strange that this is needed, i'd expect sixtyfps.h to be sufficient

api/sixtyfps-cpp/tests/datastructures.cpp
```diff
@@ -72,3 +73,24 @@ TEST_CASE("Property Tracker")
     REQUIRE(!tracker1.is_dirty());
 }
 
+TEST_CASE("Image")
+{
+    using namespace sixtyfps;
+
+    // ensure a backend exists, using private api
+    private_api::ComponentWindow wnd;
```
ogoffart, line 81: Maybe better to call `sixtyfps::testing::init();`

api/sixtyfps-cpp/tests/datastructures.cpp
tronical: The same applies to brush I think. We can change that though. I’d do that separate from this merge request though.

api/sixtyfps-cpp/tests/datastructures.cpp
tronical: Good idea!

tronical: Looks like this will require adding the feature support to corrosion, in order to enable the testing feature.

ogoffart:
>   /home/runner/work/sixtyfps/sixtyfps/api/sixtyfps-cpp/include/sixtyfps_testing.h:17: undefined reference to `sixtyfps_testing_init_backend'

The runtime lib would need to be built with the testing feature, which I'm not sure if we want that when building with cmake in this CI step. Perhaps the previous way was better then

ogoffart (approved):

PR #269, opened 2021-06-21
C++: hide internal types

tronical:


api/sixtyfps-cpp/include/sixtyfps.h
```diff
@@ -477,7 +477,7 @@ class VectorModel : public Model<ModelData>
 template<typename C, typename ModelData>
 class Repeater
```
ogoffart, line 478: I guess the Repeater itself need to be moved.

xtask/src/cbindgen.rs
```diff
@@ -308,7 +308,7 @@ fn gen_corelib(root_dir: &Path, include_dir: &Path) -> anyhow::Result<()> {
             r"
 namespace sixtyfps {{
     namespace private_api {{ enum class VersionCheck {{ Major = {}, Minor = {}, Patch = {} }}; class ComponentWindow; }}
-    namespace cbindgen_private {{ using sixtyfps::private_api::ComponentWindow; using namespace vtable; struct KeyEvent; }}
+    namespace cbindgen_private {{ using sixtyfps::private_api::ComponentWindow; using namespace vtable; struct KeyEvent; using namespace private_api; }}
```
ogoffart, line 311: what's the reason for this change?

ogoffart (approved):

xtask/src/cbindgen.rs
tronical: For the structs for different items, cbindgen maps `width: Property<f32>` to `Property<f32> width`. Since the generated structs live in `cbindgen_private` this was an easy way for that to compile. If you prefer I can also look for an alternate way so that cbindgen generates `private_api::Property<f32>` instead.

api/sixtyfps-cpp/include/sixtyfps.h
tronical: Done, along with more types.

xtask/src/cbindgen.rs
ogoffart: I guess it's fine like that.

Maybe one want to be a bit more explicit and only do `using private_api::Property;`

ogoffart (approved):

xtask/src/cbindgen.rs
tronical: Done

PR #270, opened 2021-06-21
Fix panic in preview.html when loading images

tronical:
Binding expressions may end up calling sixtyfps::Image:size(), which
requires a graphics backend to be set.

The wasm interpreter code path circumvented the sixtyfps_corelib::backend::instance_or_init
call and went straight to the GL backend.

This patch re-introduces the usual routing through the default backend.

### DIFF
--- a/sixtyfps_runtime/interpreter/dynamic_component.rs
+++ b/sixtyfps_runtime/interpreter/dynamic_component.rs
@@ -331,7 +331,11 @@ impl<'id> ComponentDescription<'id> {
         #[cfg(not(target_arch = "wasm32"))]
         let window = sixtyfps_rendering_backend_default::backend().create_window();
         #[cfg(target_arch = "wasm32")]
-        let window = sixtyfps_rendering_backend_gl::create_gl_window_with_canvas_id(canvas_id);
+        let window = {
+            // Ensure that the backend is initialized
+            sixtyfps_rendering_backend_default::backend();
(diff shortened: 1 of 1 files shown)


ogoffart:
What i would do is:

```rust
let window = { 
     sixtyfps_rendering_backend_default::backend(); // ensure that the backend is initialized
    sixtyfps_rendering_backend_gl::create_gl_window_with_canvas_id(canvas_id)
};
```
This way there is no need to expose a new interface in the back-end itself for this.

But if we need this interface, then we can remove the explicit dependency to the gl backend in the interpreter/Cargo.toml

ogoffart:
Yet another possibility is to call `sixtyfps_corelib::backend::instance_or_init`  in `sixtyfps_rendering_backend_gl::create_gl_window_with_canvas_id`

tronical:
> What i would do is:
> 
> ```rust
> let window = { 
>      sixtyfps_rendering_backend_default::backend(); // ensure that the backend is initialized
>     sixtyfps_rendering_backend_gl::create_gl_window_with_canvas_id(canvas_id)
> };
> ```
> 
> This way there is no need to expose a new interface in the back-end itself for this.
> 
> But if we need this interface, then we can remove the explicit dependency to the gl backend in the interpreter/Cargo.toml

Okay, I'll do this.

ogoffart (approved):

PR #274, opened 2021-06-26
Just typos in code

hunger:
Just an attempt to clean out the list of issues I see when I open 60fps :-)

Sorry, the patch names are not too descriptive.

I find the Code Spell Checker very helpful while coding.

### DIFF
--- a/sixtyfps_compiler/passes/binding_analysis.rs
+++ b/sixtyfps_compiler/passes/binding_analysis.rs
@@ -210,7 +210,7 @@ fn visit_implicit_layout_info_dependencies(
         }
         "Text" => {
             vis(&NamedReference::new(&item, "text"));
-            vis(&NamedReference::new(&item, "font_familly"));
+            vis(&NamedReference::new(&item, "font_family"));
             vis(&NamedReference::new(&item, "font_size"));
             vis(&NamedReference::new(&item, "font_weight"));
             vis(&NamedReference::new(&item, "letter_spacing"));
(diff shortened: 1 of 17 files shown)


hunger:
OK, let's split this patch set: These are the code changes. The CI seems to be happy about those!

Let's move the other two patches that fix strings and comments as well as markdown warnings out.

hunger:
@tronical, @ogoffart : Your tests are happy with this, so am I:-) Please review.

sixtyfps_compiler/passes/binding_analysis.rs
```diff
@@ -210,7 +210,7 @@ fn visit_implicit_layout_info_dependencies(
         }
         "Text" => {
             vis(&NamedReference::new(&item, "text"));
-            vis(&NamedReference::new(&item, "font_familly"));
+            vis(&NamedReference::new(&item, "font_family"));
```
ogoffart, line 213: this one is interesting!!  that was a bug!

ogoffart (approved):

sixtyfps_compiler/passes/binding_analysis.rs
hunger: My first bug fix then! :-)

tronical: Slow and steady wins the race :-)

btw, did you configure the spell checker in any special way?

hunger: I did add a ton of words to its ignore/workspace list, yes.

You have quite a few apprevtions and shorthand forms in your codebase that the code spellchecker does not like too much!

tronical: I'm thinking of adding my `cspell.json` to the git repo simply. Or you could also submit yours if you have one already :-)

ogoffart:
We could have spell check as part of the CI, but i suppose it indeed need quite some white list.

sixtyfps_compiler/passes/binding_analysis.rs
hunger: In my experience the list is manageable -- if you do are comfortable to have really short (<= 3 chars) or really long (real words with camelcase/snakecase) local variables.

PR #275, opened 2021-06-27
Janitor: Fix typos in comments and user-facing strings

hunger:
No behavior change is intended!


hunger:
@ogoffart , @tronical: Your CI is happy with these now, feel free to take a look.

I did also make sure that error messages do not end in a `.`. Most did already, so I removed the few trailing dots I could find.

ogoffart (approved):
thanks

matklad:
> I did also make sure that error messages do not end in a .. Most did already, so I removed the few trailing dots I could find.

drive-by: might be a good idea to include debug_assert into the error-emitting code to check this.

ogoffart:
Thanks for the suggestion: Done in commit f5de7e0

PR #279, opened 2021-06-30
GL backend: elide Text also when word-wrapping

ogoffart:
In this case, the last line will have elipsis on the last line
when no more line can be printed

Note that the behavior is different with the Qt backend

### DIFF
--- a/sixtyfps_runtime/rendering_backends/gl/lib.rs
+++ b/sixtyfps_runtime/rendering_backends/gl/lib.rs
@@ -597,6 +597,7 @@ impl ItemRenderer for GLItemRenderer {
             .clone();
 
         let wrap = text.wrap() == TextWrap::word_wrap;
+        let elide = text.overflow() == TextOverflow::elide;
         let letter_spacing = text.letter_spacing() * self.scale_factor;
         let text_size =
             font.text_size(letter_spacing, string, if wrap { Some(max_width) } else { None });
@@ -631,9 +632,9 @@ impl ItemRenderer for GLItemRenderer {
             *y += font_metrics.height();
         };
 
-        if wrap {
-            let mut start = 0;
-            while start < string.len() && y + font_metrics.height() <= max_height {
+        let mut start = 0;
+        'lines: while start < string.len() && y + font_metrics.height() <= max_height {
+            if wrap && (!elide || y + 2. * font_metrics.height() <= max_height) {
                 let index = canvas.break_text(max_width, &string[start..], paint).unwrap();
                 if index == 0 {
                     // FIXME the word is too big to be shown, but we should still break, ideally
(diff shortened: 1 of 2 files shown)


ogoffart:
I wonder how to implement the Qt way.

What we currently do is https://github.com/sixtyfpsui/sixtyfps/blob/b5f2bbe8bb8d67a05778ca24375b1ef201a045b5/sixtyfps_runtime/rendering_backends/qt/qt_window.rs#L411-L416

Which is also broken if there are `\n` in the text.  I guess we'll have to split in lines and draw each line separately?

sixtyfps_runtime/rendering_backends/gl/lib.rs
```diff
@@ -643,15 +644,14 @@ impl ItemRenderer for GLItemRenderer {
...
                 start = index;
-            }
-        } else {
-            let elide = text.overflow() == TextOverflow::elide;
-            'lines: for line in string.lines() {
-                if y + font_metrics.height() > max_height {
-                    break;
-                }
+            } else {
+                let index = string[start..].find('\n').map_or(string.len(), |i| start + i + 1);
```
tronical, line 648: We used to call `string.lines()`, which also respects `\r\n`, but it's missing U+2028 and U+2029 (line and paragraph separator). I think we should replace all `\r\n`/`\n` and U+2029 with U+2028 and split on that.

That's slightly unrelated to this patch though, feel free to ignore :-)

ogoffart: Note that the trim() on the next line will remove the \r as well.

tronical (approved):
A manual test would be nice, like some of those .60 files we have in tests/cases that when loaded can be used to visually verify.

sixtyfps_runtime/rendering_backends/qt/qt_window.rs
```diff
@@ -402,16 +402,65 @@ impl ItemRenderer for QtItemRenderer<'_> {
...
+                while (true) {
+                    auto line = layout.createLine();
+                    if (!line.isValid()) {
+                        last_line_begin = string.size();
+                        break;
+                    }
+                    line.setLineWidth(rect.width());
+                    auto lf = QStringView(string).mid(line.textStart(), line.textLength()).indexOf('\n');
+                    if (lf >= 0) {
+                        line.setNumColumns(lf);
```
tronical, line 450: setNumColumns will do the line layout again, it would be more efficient to either call `setLineWidth` or `setNumColumns`.

I'm curious though: Why is the call to `setNumColumns` needed? Is it because you observed QTextLayout not respecting `\n`? If that is why, then I think it would be easier to upfront convert `\n` and `\r\n` to `QChar::LineSeparator` and just call `setLineWidth()` here.

sixtyfps_runtime/rendering_backends/qt/qt_window.rs
```diff
@@ -402,16 +402,65 @@ impl ItemRenderer for QtItemRenderer<'_> {
...
+                    }
+                    last_line_begin = line.textStart();
+                    last_line_size = line.textLength();
+                }
+                if (last_line_begin < string.size()) {
+                    elided = string.left(last_line_begin);
+                    QString to_elide = QStringView(string).mid(last_line_begin, last_line_size).trimmed() % QStringView(u"…");
+                    elided += fm.elidedText(to_elide, Qt::ElideRight, rect.width());
+                }
                 painter->drawText(rect, flags, elided);
```
tronical, line 464: This will do the entire work of line breaking a second time though. I think it would be more efficient to call `layout.endLayout();` and then draw it using `QTextLayout::draw()`, followed by one more call for the elision character if needed. The x position of that would I think be `QTextLine::naturalTextWidth()` of the last line.

tronical (approved):

sixtyfps_runtime/rendering_backends/qt/qt_window.rs
ogoffart: Yes, \n was not considered as a line separator. I'll try to do the change you suggest

sixtyfps_runtime/rendering_backends/qt/qt_window.rs
ogoffart: true, but knowing exactly where to cut the last line and put the elipsis is a bit of work. I'll keep it like this for now.

PR #284, opened 2021-07-02
Pr/fixes

hunger:
Let's move out the actual fixes from the clippy stuff and get those merged.

@ogoffart: I removed the drop you commented on in the clippy fixes PR. I did leave in the clones in clones the "&& clones the ref" patch.

### DIFF
--- a/xtask/src/license_headers_check.rs
+++ b/xtask/src/license_headers_check.rs
@@ -324,8 +324,8 @@ impl CargoDependency {
             toml_edit::Value::InlineTable(table) => {
                 if let (Some(path), Some(version)) = (table.get("path"), table.get("version")) {
                     Some(Self::Full {
-                        path: path.as_str().unwrap_or_default().clone().into(),
-                        version: version.as_str().unwrap_or_default().clone().into(),
+                        path: path.as_str().unwrap_or_default().to_owned(),
+                        version: version.as_str().unwrap_or_default().to_owned(),
                     })
                 } else {
                     None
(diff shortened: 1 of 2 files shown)


xtask/src/license_headers_check.rs
```diff
@@ -324,8 +324,8 @@ impl CargoDependency {
             toml_edit::Value::InlineTable(table) => {
                 if let (Some(path), Some(version)) = (table.get("path"), table.get("version")) {
                     Some(Self::Full {
-                        path: path.as_str().unwrap_or_default().clone().into(),
-                        version: version.as_str().unwrap_or_default().clone().into(),
+                        path: *(path.as_str().unwrap_or_default()).clone().into(),
```
tronical, line 327: I don’t quite understand this. I can see how a &str could be cloned and into() taking care of allocating a new string. But the patch looks like it derefs the &str and the tries the clone? How would that work? What am I missing?:-)

hunger: You are right... I was too mechanical there, fixing the clippy warning!

xtask/src/license_headers_check.rs
```diff
@@ -324,8 +324,8 @@ impl CargoDependency {
             toml_edit::Value::InlineTable(table) => {
                 if let (Some(path), Some(version)) = (table.get("path"), table.get("version")) {
                     Some(Self::Full {
-                        path: path.as_str().unwrap_or_default().clone().into(),
-                        version: version.as_str().unwrap_or_default().clone().into(),
+                        path: path.as_str().unwrap_or_default().to_owned(),
```
tronical, line 327: 👍nice

tronical (approved):

PR #290, opened 2021-07-03
Janitor: Fix clippy error about a loop that never loops

hunger:
This code abuses a loop for flow control. Split this out into a separate
function instead, using a custom enum to transfer the information back.

This is necessary since we need to transfer ownership of the
mouse_input_state back to the caller.

This change should not cause a behavior change.

### DIFF
--- a/sixtyfps_runtime/corelib/input.rs
+++ b/sixtyfps_runtime/corelib/input.rs
@@ -267,6 +267,61 @@ pub struct MouseInputState {
     grabbed: bool,
 }
 
+enum MouseGrab {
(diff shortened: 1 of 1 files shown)


tronical (commented):
I agree with clippy, your variant is easier to read. I can’t see any difference either in the behavior.

Let’s see what @ogoffart says :-)

ogoffart:
I would have used `goto` if this was possible, but it isn't so loops that break is the best that can be done.

Anyway, since simon thinks it's easier to read, let's merge that :-)

PR #291, opened 2021-07-03
Janitor: Fix clippy::if_same_then_else error

hunger:
In this case this feels at controversial, but in general I find this
clippy lint pretty helpful: It has caught several copy-paste errors
before.

No behavior change is intended with this patch.

### DIFF
--- a/sixtyfps_compiler/passes/resolving.rs
+++ b/sixtyfps_compiler/passes/resolving.rs
@@ -892,12 +892,11 @@ impl Expression {
                     (target_type, expr_ty) => {
                         if expr_ty.can_convert(&target_type) {
                             target_type
-                        } else if target_type.can_convert(&expr_ty) {
-                            expr_ty
-                        } else if expr_ty.default_unit().is_some()
-                            && matches!(target_type, Type::Float32 | Type::Int32)
+                        } else if target_type.can_convert(&expr_ty)
+                            || (expr_ty.default_unit().is_some()
+                                && matches!(target_type, Type::Float32 | Type::Int32))
                         {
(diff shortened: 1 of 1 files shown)


sixtyfps_compiler/passes/resolving.rs
```diff
@@ -892,12 +892,11 @@ impl Expression {
                     (target_type, expr_ty) => {
                         if expr_ty.can_convert(&target_type) {
                             target_type
-                        } else if target_type.can_convert(&expr_ty) {
-                            expr_ty
-                        } else if expr_ty.default_unit().is_some()
-                            && matches!(target_type, Type::Float32 | Type::Int32)
+                        } else if target_type.can_convert(&expr_ty)
+                            || (expr_ty.default_unit().is_some()
```
tronical, line 896:
```suggestion
                            || (expr_ty.default_unit().is_some() // // in case this is the '0' literal
```

sixtyfps_compiler/passes/resolving.rs
```diff
@@ -892,12 +892,11 @@ impl Expression {
...
+                                && matches!(target_type, Type::Float32 | Type::Int32))
                         {
-                            // in case this is the '0' literal
+                            // in the or case: The `0` literal.
```
tronical, line 899:
```suggestion
```

sixtyfps_compiler/passes/resolving.rs
hunger: cargo fmt moves that away all the time:-)

tronical (approved):

PR #293, opened 2021-07-04
LSP: Update Sublime Text instructions for version 4

jturcotte:
The showPreview workaround was only necessary in Sublime Text 3,
so remove it.

Version 3 is still used by some people but most future users will be
using Sublime Text 4.


ogoffart:
thanks

PR #296, opened 2021-07-06
Wip/clippy fixes

hunger:
The daily dosage of clippy fixes.

### DIFF
--- a/sixtyfps_compiler/passes/z_order.rs
+++ b/sixtyfps_compiler/passes/z_order.rs
@@ -41,9 +41,7 @@ fn reorder_children_by_zorder(
             .remove("z")
             .and_then(|e| eval_const_expr(&e.expression, "z", &e, diag));
         let z = z.or_else(|| {
-            if child_elm.borrow().repeated.is_none() {
-                return None;
-            };
+            child_elm.borrow().repeated.as_ref()?;
             if let Type::Component(c) = &child_elm.borrow().base_type {
                 c.root_element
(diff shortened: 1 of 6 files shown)


sixtyfps_compiler/passes/z_order.rs
```diff
@@ -41,9 +41,7 @@ fn reorder_children_by_zorder(
             .remove("z")
             .and_then(|e| eval_const_expr(&e.expression, "z", &e, diag));
         let z = z.or_else(|| {
-            if child_elm.borrow().repeated.is_none() {
-                return None;
-            };
+            child_elm.borrow().repeated.as_ref()?;
```
tronical, line 44: I really like every part of this patch except this one :-). Is it just me?

hunger: It definitely is canonical rust: Clippy explicitly says so:-)

The justification for this clippy lint is this: "Question mark usage is more idiomatic."

But I do not mind to just disable `clippy::question_mark`: I have never had this one before indicate some typo/copy and paste error.

tronical: I’d favor disabling it in this case, but not too strong feeling about it. @ogoffart what’s your opinion?

ogoffart: I like canonical rust ...  What's wrong with it?

tronical (approved):

PR #297, opened 2021-07-06
Janitor: CSpell fixes

hunger:
Mark up some special words in documents (like C++ lingo in rust files).

### DIFF
--- a/sixtyfps_compiler/generator/cpp.rs
+++ b/sixtyfps_compiler/generator/cpp.rs
@@ -10,6 +10,8 @@ LICENSE END */
 /*! module for the C++ code generator
 */
 
+// cSpell:ignore cstdlib cmath constexpr nullptr decltype intptr uintptr
+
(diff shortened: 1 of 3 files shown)


sixtyfps_compiler/generator/cpp.rs
```diff
@@ -10,6 +10,8 @@ LICENSE END */
 /*! module for the C++ code generator
 */
 
+// cSpell:ignore cstdlib cmath constexpr nullptr decltype intptr uintptr
```
tronical, line 13: Won’t those also show up in C++ sources? Just curious

hunger: Nope, the keywords of the language in use are ignored by cspell.

tronical: Ah, great, that's good. Thanks for the explanation :)

tronical (approved):

PR #300, opened 2021-07-07
Centralize and version highlight.js integration

tronical:
This brings the entire highlight js integration into this repo, shared across rustdoc, sphinx and mdbook and pinned as version from the upstream CDN.


ogoffart:
I wonder about the path:  `docs/html` looks like this folder would contain an html version of the docs, but that's not the case.
maybe `docs/resources`  or `docs/helper`  it could also contains images and other artifacts.

tronical:
I like `docs/resources` - will use that

ogoffart (approved):

PR #308, opened 2021-07-09
Pr/janitor

hunger:
More clippy and similar fixes.

### DIFF
--- a/sixtyfps_runtime/corelib/layout.rs
+++ b/sixtyfps_runtime/corelib/layout.rs
@@ -9,6 +9,8 @@
 LICENSE END */
 //! Runtime support for layouts.
 
+// cspell:ignore coord
+
 use crate::{slice::Slice, SharedVector};
 
 /// Vertical or Horizontal orientation
--- a/sixtyfps_runtime/corelib/layout.rs
+++ b/sixtyfps_runtime/corelib/layout.rs
@@ -249,25 +249,25 @@ mod grid_internal {
             vec![grid_internal::LayoutData { stretch: 1., ..Default::default() }; num as usize];
         let mut has_spans = false;
         for cell in data {
-            let cnstr = &cell.constraint;
-            let mut max = cnstr.max;
+            let constraint = &cell.constraint;
+            let mut max = constraint.max;
             if let Some(size) = size {
-                max = max.min(size * cnstr.max_percent / 100.);
+                max = max.min(size * constraint.max_percent / 100.);
             }
             for c in 0..(cell.span as usize) {
                 let cdata = &mut layout_data[cell.col_or_row as usize + c];
                 cdata.max = cdata.max.min(max);
             }
             if cell.span == 1 {
-                let mut min = cnstr.min;
+                let mut min = constraint.min;
                 if let Some(size) = size {
-                    min = min.max(size * cnstr.min_percent / 100.);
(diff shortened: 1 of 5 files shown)


api/sixtyfps-rs/sixtyfps-build/lib.rs
```diff
@@ -61,16 +61,17 @@ pub struct CompilerConfiguration {
     config: sixtyfps_compilerlib::CompilerConfiguration,
 }
 
-impl CompilerConfiguration {
-    /// Creates a new default configuration.
-    pub fn new() -> Self {
+impl Default for CompilerConfiguration {
```
tronical, line 64: Hmm, this changes public API in a source incompatible way. I suggest to move that out of this merge request and we discuss it separately.

sixtyfps_runtime/corelib/model.rs
```diff
@@ -263,7 +263,9 @@ impl<T> Model for ModelHandle<T> {
     }
 
     fn attach_peer(&self, peer: ModelPeer) {
-        self.0.as_ref().map(|model| model.attach_peer(peer));
+        if let Some(model) = self.0.as_ref() {
```
tronical, line 266: This comes from clippy, right?

sixtyfps_runtime/corelib/properties.rs
```diff
@@ -729,7 +729,7 @@ impl<T: Clone + InterpolatedPropertyValue + 'static> Property<T> {
             compute_animation_details: || -> AnimationDetail { None },
         };
 
-        // Safety: the AnimatedBindingCallable's type match the property type
+        // Safety: the `AnimatedBindingCallable`'s type match the property type
```
tronical, line 732: Neat workaround:)

sixtyfps_runtime/corelib/layout.rs
```diff
@@ -549,7 +549,7 @@ pub fn solve_box_layout(data: &BoxLayoutData, repeater_indexes: Slice<u32>) -> S
             current_offset += 1;
             break current_offset - 1;
         };
-        res[o * 2 + 0] = layout.pos;
```
ogoffart, line 550: That's not a "useless addition", this was to keep it aligned with the next line.
But that's true that not so important to keep aligned. (It used to be more important before when there was 4 entries)

sixtyfps_runtime/corelib/model.rs
hunger: Yes: https://rust-lang.github.io/rust-clippy/master/index.html#option_map_unit_fn

I had similar changes in other places before.

api/sixtyfps-rs/sixtyfps-build/lib.rs
hunger: I dropped this one now.

sixtyfps_runtime/corelib/layout.rs
hunger: That's what the commit message also says. Shall I drop it?

sixtyfps_runtime/corelib/model.rs
tronical: Thanks. Yeah I agree with the change, I also do find it slightly more readable:)

tronical (approved):

PR #309, opened 2021-07-09
Use `PartialEq` for `ModelHanlde<T>`

hunger:
... instead of doing a pointer comparison.

This gets rid of one of the two ptr_eq in the codebase related to the ModelHandle.


hunger:
Oh, sorry for the noise!

PR #310, opened 2021-07-10
Use `impl Default` instead of `new()` constructors

hunger:
The detault trait impl instead of `new()` patch separated out of the latest janitor patch set.

### DIFF
--- a/api/sixtyfps-node/native/lib.rs
+++ b/api/sixtyfps-node/native/lib.rs
@@ -62,7 +62,7 @@ fn load(mut cx: FunctionContext) -> JsResult<JsValue> {
         }
         None => vec![],
     };
-    let mut compiler = sixtyfps_interpreter::ComponentCompiler::new();
+    let mut compiler = sixtyfps_interpreter::ComponentCompiler::default();
     compiler.set_include_paths(include_paths);
     let c = spin_on::spin_on(compiler.build_from_path(path));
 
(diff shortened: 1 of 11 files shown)


tronical:
Singled out, this source incompatible change makes sense. I see two arguments against this though:

1. It seems more idiomatic to me to start a builder pattern with new() instead of default(). I thought Default is typically used for structs that don’t try to encapsulate anything.
2. new() has the advantage of being more extensible naming wise if the starting point of the builder pattern needs extension through an initial parameter in the future.

But I may very well be missing something about the intention of the Default trait, so no objections here - just curiosity to learn what’s idiomatic Rust API (as that’s the goal).

hunger:
Clippy is the tool that teaches most programmers what ideonmatic rust is, so I do not argue with it on this topic :-) That leaves whether or not ideomatic rust is a good thing to have in these particular places.

I like `default()` to be used consistently, simply because I can never remember where to use `new()` and where `default()` otherwise:-) Plus the `Default` trait does enable some convenient things like `Option::unwrap_or_default`, etc.

tronical:
In the standard library there are at least three builder pattern types (std::process::Command, std::fs::DirBuilder and std::thread::Builder) that don’t implement Default but use new() instead.

tronical:
I’m torn. I see the one pattern all the time (winit is another example btw and rather popular), but I see clippy on the other hand.

tronical:
@ogoffart  pointed out that we could have both, and I'm in favour of that. Implementing `Default` *and* keep the `new()` function, to get the best of both. Is that ok with you @hunger ?

hunger:
Its definitely an option, yes:-)

tronical:
Okay, great. Then I'll merge your change and re-add the `new()`.

tronical (approved):

PR #311, opened 2021-07-10
Model: Unify PartialEq

hunger:
Implement PartialEq for Rc<Model> and use that for ModelHandle as well
as Value<Model>

Wrap a Rc<Model> into a SharedModel and implement PartialEq for that.

Wrap Option<SharedModel> into the ModuleHandle as well as Value::Model.

### DIFF
--- a/sixtyfps_runtime/corelib/model.rs
+++ b/sixtyfps_runtime/corelib/model.rs
@@ -128,7 +128,7 @@ impl<T: 'static> VecModel<T> {
     where
         T: Clone,
     {
-        ModelHandle(Some(Rc::<Self>::new(slice.to_vec().into())))
+        ModelHandle::new(Rc::<Self>::new(slice.to_vec().into()))
     }
 
     /// Add a row at the end of the model
@@ -205,14 +205,49 @@ impl Model for bool {
     }
 }
 
+/// A shared model
+#[repr(transparent)]
+#[derive(derive_more::Deref, derive_more::DerefMut)]
+pub struct SharedModel<T>(pub Rc<dyn Model<Data = T>>);
+
+impl<T> std::fmt::Debug for SharedModel<T> {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        write!(f, "SharedModel(dyn Model)")
+    }
+}
+
+impl<T> Clone for SharedModel<T> {
+    fn clone(&self) -> Self {
+        Self(self.0.clone())
+    }
+}
+
+impl<T> core::cmp::PartialEq for SharedModel<T> {
+    #[allow(clippy::vtable_address_comparisons)]
+    // We think vtable_address_comparisons are safe here!
+    // Rc::ptr_eq will compare a fat pointer consisting of a `vtable` and a `self` pointer.
+    //
+    // We can get:
+    // * false positives: Two different `Model`s get considered the same, even though they are not.
+    //   For this the `self` pointer need to point to the same place. That can not happen here
+    //   since `Rc` will allocate for the refcounts, even if `T` is `()`.
+    // * false negatives: The same `Model` gets considered to be two different `Model`s.
+    //   In this case the `vtable` pointer must differ, e.g. because two crates created separate
+    //   instances of the vtable for the `Model` trait.
+    //   This will trigger some unnecessary rendering, but is benign otherwise.
+    fn eq(&self, other: &Self) -> bool {
+        Rc::ptr_eq(&self.0, &other.0)
+    }
+}
+
 /// Properties of type array in the .60 language are represented as
 /// an [`Option`] of an [`Rc`] of something implemented the [`Model`] trait
 #[derive(derive_more::Deref, derive_more::DerefMut, derive_more::From, derive_more::Into)]
-pub struct ModelHandle<T>(pub Option<Rc<dyn Model<Data = T>>>);
+pub struct ModelHandle<T>(pub Option<SharedModel<T>>);
 
 impl<T> std::fmt::Debug for ModelHandle<T> {
     fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
-        write!(f, "ModelHandle({:?})", self.0.as_ref().map(|_| "dyn Model"))
+        write!(f, "ModelHandle({:?})", self.0.as_ref())
     }
 }
 
@@ -221,6 +256,7 @@ impl<T> Clone for ModelHandle<T> {
         Self(self.0.clone())
     }
 }
+
 impl<T> Default for ModelHandle<T> {
     fn default() -> Self {
         Self(None)
@@ -231,9 +267,8 @@ impl<T> core::cmp::PartialEq for ModelHandle<T> {
     fn eq(&self, other: &Self) -> bool {
         match (&self.0, &other.0) {
             (None, None) => true,
-            (None, Some(_)) => false,
-            (Some(_), None) => false,
-            (Some(a), Some(b)) => Rc::ptr_eq(a, b),
+            (Some(a), Some(b)) => a == b,
+            _ => false,
         }
     }
 }
@@ -247,7 +282,7 @@ impl<T> From<Rc<dyn Model<Data = T>>> for ModelHandle<T> {
 impl<T> ModelHandle<T> {
     /// Create a new handle wrapping the given model
     pub fn new(model: Rc<dyn Model<Data = T>>) -> Self {
-        Self(Some(model))
+        Self(Some(SharedModel(model)))
     }
 }
 
--- a/sixtyfps_runtime/corelib/model.rs
+++ b/sixtyfps_runtime/corelib/model.rs
@@ -242,7 +242,9 @@ impl<T> core::cmp::PartialEq for SharedModel<T> {
 
 /// Properties of type array in the .60 language are represented as
 /// an [`Option`] of an [`Rc`] of something implemented the [`Model`] trait
-#[derive(derive_more::Deref, derive_more::DerefMut, derive_more::From, derive_more::Into)]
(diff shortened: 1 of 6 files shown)


sixtyfps_runtime/corelib/model.rs
```diff
@@ -205,14 +205,49 @@ impl Model for bool {
     }
 }
 
+/// Properties of type array in the .60 language are represented as
+/// an [`Option`] of an [`Rc`] of something implemented the [`Model`] trait
+#[derive(derive_more::Deref, derive_more::DerefMut)]
+pub struct SharedModel<T>(pub Rc<dyn Model<Data = T>>);
```
tronical, line 211: Should this be repr(transparent)?

hunger: Makes probably sense, yes.

sixtyfps_runtime/corelib/model.rs
```diff
@@ -231,23 +267,16 @@ impl<T> core::cmp::PartialEq for ModelHandle<T> {
...
-            (None, Some(_)) => false,
-            (Some(_), None) => false,
-            (Some(a), Some(b)) => Rc::ptr_eq(a, b),
+            (Some(a), Some(b)) => a == b,
+            _ => false,
         }
     }
 }
 
-impl<T> From<Rc<dyn Model<Data = T>>> for ModelHandle<T> {
```
tronical, line 241: Just curious, why was this removed?

hunger: I did not need it to build, but then I apparently did not build everything somehow.

At least it did build and test just fine on my local machine and fails to build here.

tronical:
I think that while this change is source incompatible, it is good - especially the centralization of the ptr_eq.

This sheds a bit of light on the fact that there are quite some ways of constructing a modelhandle :). Is that why you removed the From?

hunger:
@tronical: Yes, this definitely needs some thinking before it can go in.

Is there some documentation on the cpp_cmake parts somewhere? Looks like I broke those.

tronical:
It seems to fail deep in the interpreter api through the cmake build. I’ll try to write something about how that plays together on Monday. Sorry about this can of worms :)

tronical:
I think the heart of the error is this:

```
Error: api/sixtyfps-cpp/generated_include/sixtyfps_interpreter_internal.h:177:9: error: ‘SharedModel’ does not name a type
```

which indicates that the new `Value` variant holding `SharedModel` is failing because `SharedModel` is not a type known in C++.

tronical:
> I think the heart of the error is this:
> 
> ```
> Error: api/sixtyfps-cpp/generated_include/sixtyfps_interpreter_internal.h:177:9: error: ‘SharedModel’ does not name a type
> ```
> 
> which indicates that the new `Value` variant holding `SharedModel` is failing because `SharedModel` is not a type known in C++.

Ah, but `Value` should not be exposed in the generated headers in the first place.

hunger:
@tronical: Ok, just did a fresh push, let's see what happens.

sixtyfps_runtime/corelib/model.rs
```diff
@@ -205,14 +205,51 @@ impl Model for bool {
...
+#[derive(derive_more::Deref, derive_more::DerefMut)]
+pub struct SharedModel<T>(pub Rc<dyn Model<Data = T>>);
+
+impl<T> std::fmt::Debug for SharedModel<T> {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        write!(f, "SharedModel(dyn Model)")
+    }
+}
+
+impl<T> Clone for SharedModel<T> {
```
tronical, line 219: I suppose this can also be achieved using `#[derive(Clone)]`?

tronical: Oh, I see, same in ModelHandle, so not relevant for this change set :)

sixtyfps_runtime/corelib/model.rs
tronical: IMO is worth keeping this public API unless there’s a good reason to remove it (such as clippy complaining). IIRC initially the Handle had just the Rc but we had the add the Option, and then this impl was added to make the API easier to use.

tronical:
> @tronical: Ok, just did a fresh push, let's see what happens.

Yay, it’s green now. Only two things left IMO:

(1) typo in the commit message (Hanlde-> Handle)
(2) either break public API but mention in the commit message why (or edit changelog), or keep it - regarding the From impl

sixtyfps_runtime/corelib/model.rs
ogoffart: the problem with #[derive(Clone)] is that it adds a `where T: Clone` which we don't want.

sixtyfps_runtime/corelib/model.rs
```diff
@@ -205,14 +205,51 @@ impl Model for bool {
...
+}
+
 /// Properties of type array in the .60 language are represented as
 /// an [`Option`] of an [`Rc`] of something implemented the [`Model`] trait
-#[derive(derive_more::Deref, derive_more::DerefMut, derive_more::From, derive_more::Into)]
-pub struct ModelHandle<T>(pub Option<Rc<dyn Model<Data = T>>>);
+#[derive(
+    PartialEq, derive_more::Deref, derive_more::DerefMut, derive_more::From, derive_more::Into,
+)]
+pub struct ModelHandle<T>(pub Option<SharedModel<T>>);
```
ogoffart, line 248: i wonder if there is a way to remove the inner `pub`

Anyway, this is a public API breakage.

sixtyfps_runtime/corelib/model.rs
```diff
@@ -205,14 +205,51 @@ impl Model for bool {
...
+
+impl<T> core::cmp::PartialEq for SharedModel<T> {
+    #[allow(clippy::vtable_address_comparisons)]
+    // We think vtable_address_comparisons are safe here!
+    // Rc::ptr_eq will compare a fat pointer consisting of a `vtable` and a `self` pointer.
+    //
+    // We can get:
+    // * false positives: Two different `Model`s get considered the same, even though they are not.
+    //   For this the `self` pointer need to point to the same place. That can only happen if
+    //   `Self` is `()`, which does not make sense for a `Model`. So we ignore this case.
```
ogoffart, line 233: Self could very well be `()`.

We could imagine for example a  `impl Model for ()` that always returns 0 elements.

or a  `struct MyStaticModel;  impl Model for MyStaticModel { ... }`.

But anyway, this cannot happen that two different model are considered the same, because the Rc adds an allocation for the refcounts and always allocates. so the data pointer will always be different anyway.

sixtyfps_runtime/corelib/model.rs
```diff
@@ -205,14 +205,51 @@ impl Model for bool {
...
+    // * false negatives: The same `Model` gets considered to be two different `Model`s.
+    //   In this case the `vtable` pointer must differ, e.g. because two crates created separate
+    //   instances of the vtable for the `Model` trait.
+    //   This will trigger some unnecessary rendering, but is benign otherwise.
```
ogoffart, line 237: yes, this is correct.


(but now that i think about it, because this is Rc, a data pointer comparison would also work as intended.)

sixtyfps_runtime/interpreter/api.rs
```diff
@@ -152,7 +152,7 @@ impl PartialEq for Value {
             Value::Bool(lhs) => matches!(other, Value::Bool(rhs) if lhs == rhs),
             Value::Image(lhs) => matches!(other, Value::Image(rhs) if lhs == rhs),
             Value::Array(lhs) => matches!(other, Value::Array(rhs) if lhs == rhs),
-            Value::Model(lhs) => matches!(other, Value::Model(rhs) if Rc::ptr_eq(lhs, rhs)),
+            Value::Model(lhs) => matches!(other, Value::Model(rhs) if lhs == rhs),
```
ogoffart, line 155: I actually wonder if this was not the only reason we implemented PartialEq manually.
We could just use #[derive(PartialEq)] instead, now.

ogoffart (commented):
I've tagged the change as `breaking_change` since this change the public API.

The changes here are overall good. But we haven't decide if the next release of the `sixtyfps` crate will be 0.1.1 or 0.2.

I guess we should wait a bit to see if we have other reasons to break the source compatibility before merging this.

Also changes to public API deserve a changelog entry.

(It would be nice if there was a CI check that could tell if we are breaking semver or touching the public API)

sixtyfps_runtime/corelib/model.rs
hunger: @ogoffart: You are way deeper in this rabbit hole than me, so did I understand this correctly?
 
 But I think you are right, this lint might actually be wrong: I went ahead and filed https://github.com/rust-lang/rust-clippy/issues/7461 with my current understanding. I am pretty sure I missed something, but maybe some improvement to the explanation of that clippy lint will fall out of this:-)

sixtyfps_runtime/interpreter/api.rs
hunger: I can do that, but would prefer to do that in a separate patch.

sixtyfps_runtime/corelib/model.rs
hunger: I will add this back in in the next push.

sixtyfps_runtime/corelib/model.rs
hunger: True, I added something about this to the CHANGELOG now.

sixtyfps_runtime/corelib/model.rs
hunger: I updated the text, I hope it is a bit better now.

PR #316, opened 2021-07-13
separate license tag from authors

tronical:


hunger:
Why did you opt for such a big license header?

Looking at the [KDE recommendations](https://community.kde.org/Guidelines_and_HOWTOs/Licensing) something like this should suffice:

```rust
// SPDX-FileCopyrightText: [CURRENT YEAR] [YOUR NAME] <[YOUR MAIL ADDRESS]>
// SPDX-License-Identifier: (GPL-3.0-or-later OR LicenseRef-Proprietary)
```

(plus extra `SPDX-FileCopyrightText` lines as needed.

The dance to get your own license into SPDX is complex, but it is doable with LicenceRef/DocumentRef AFAICT.

tronical:
I think it’s because I’m used to even bigger ones, so I perceive these five lines or so as short. But really like your suggestion and will look into using that instead. Thanks!:)

ogoffart:
I'm also wondering if we really need to have license headers to every single files. 
These are especially annoying in tests cases, or examples that are meant to be small.

tronical:
Regarding this:

> // SPDX-License-Identifier: (GPL-3.0-or-later OR LicenseRef-Proprietary)

It seems that `LicenseRef-[idstring]` just requires that a file with that name (idstring) is in the `LICENSES/` directory, so for example `LicenseRef-SixtyFPS-Commercial` and then `LICENSES/SixtyfFPS-Commercial.txt`.

Does that match your understanding as well?

Then, regarding the contents of that file: I wonder if it's sufficient to let that file simply contain the current text, so

```
This software is also available under commercial licensing terms.
Please contact info@sixtyfps.io for more information.
```

What do you think?

tronical:
> I'm also wondering if we really need to have license headers to every single files.
> These are especially annoying in tests cases, or examples that are meant to be small.

Yes, perhaps there's an easier way. REUSE also suggests that this is an option: https://reuse.software/spec/#dep5

tronical:
That said, it seems even more obscure to have this in a somewhat hidden `.reuse/dep5` file.

My preference are short SPDX Identifiers - I'm not concerned about the "visual noise". If you feel that it's okay to omit this as well, how about this then:

```
// SPDX-FileCopyrightText: [CURRENT YEAR] [YOUR NAME] <[YOUR MAIL ADDRESS]>
// This file is may be used under the terms described in the file LICENSE.md, included in the packaging of this file.
```

I think that we should have *something* in the files that makes it easy to see who is the copyright holder, author and how to find out what terms of licensing are.

hunger:
The idea of SPDX is that you can automate license and compliance management. I know some OSS compliance people that will curse loudly whenever they encounter a project that does not get SPDX right;-)

So I think we should encode the license information in SPDX-compatible terms *somehow*. So leaving out the SPDX identifier is not what I would do:-)

I think those dep5 files are most suited to mark up areas of 3rd party code where you do not want to modify files. I would not use that for my own projects code: The copyright info tends to get outdated rather often when it is in the same file. Having the data in a separate file will not improve the situation.

hunger:
So can we move towards something like this then:

```
// SPDX-FileCopyrightText: [CURRENT YEAR] [YOUR NAME] <[YOUR MAIL ADDRESS]>
// SPDX-License-Identifier: (GPL-3.0-or-later OR LicenseRef-SixtyFPS)
```

with the current text in LICENSES/SixtyFPS.md?

The current approach will trigger for all your customers that use a SDPX compatible tool as that will only see "GPL" and flag that. The commercial option is totally hidden from these tools right now.

tronical:
Succeeded by #760 :)

Issue #133, opened 2020-12-14
Stack overflow on Windows in Rust Debug Builds


tronical:
Due to the inlining the generated structures tend to become rather big (the gallery for example ~13kb with the ugly style), which on debug builds with Rust on Windows tends to result in a bunch of copies of said struct. That can lead to the process running out of stack space.

ogoffart:
> Due to the inlining

I don't think this is due to "inlining",  the datastructure would have the same size or bigger if the elements where kept in different datastructure  (they would still be placed within the same component)

We would need a way in rust to allocate the structure directly on the heap without making copies on the stack.

ogoffart:
commit 51836a9 might be the fix, but we still need to test on windows

tronical:
The issue still exists today. On the printer demo with the VS debugger I get this:

backtrace:
```
 	printerdemo.exe!__chkstk() Line 109	Unknown
>	printerdemo.exe!printerdemo::sixtyfps_generated_MainWindow::InnerMainWindow::new() Line 496	Unknown
 	printerdemo.exe!printerdemo::sixtyfps_generated_MainWindow::MainWindow::new() Line 16284	Unknown
 	printerdemo.exe!printerdemo::main() Line 52	Unknown
 	printerdemo.exe!core::ops::function::FnOnce::call_once<fn(),tuple<>>(void(*)()) Line 227	Unknown
 	printerdemo.exe!std::sys_common::backtrace::__rust_begin_short_backtrace<fn(),tuple<>>(void(*)() f) Line 128	Unknown
 	printerdemo.exe!std::rt::lang_start::{{closure}}<tuple<>>(std::rt::lang_start::closure-0 *) Line 49	Unknown
 	[Inline Frame] printerdemo.exe!core::ops::function::impls::{{impl}}::call_once() Line 280	Unknown
 	[Inline Frame] printerdemo.exe!std::panicking::try::do_call() Line 379	Unknown
 	[Inline Frame] printerdemo.exe!std::panicking::try() Line 343	Unknown
 	[Inline Frame] printerdemo.exe!std::panic::catch_unwind() Line 431	Unknown
 	printerdemo.exe!std::rt::lang_start_internal() Line 34	Unknown
 	printerdemo.exe!std::rt::lang_start<tuple<>>(void(*)() main, __int64 argc, unsigned char * * argv) Line 48	Unknown
 	[External Code]	
```

and this:

[...]

PR #317, opened 2021-07-13
Attempt to fix a stack overflow on Windows

ogoffart:
Issue #133

(I haven't tested it on windows yet so I don't know if this actually solves the problem)


tronical (approved):
This makes sense and I think it should go in.

Unfortunately it doesn't fix the problem though. I'll add more details to the original issue.

ogoffart:
If it does not help, I don't think it should go in, because this code is having more unsafe operations that the code it replaces.

PR #326, opened 2021-07-15
Pr/clippy fixes

hunger:
Yet more clippy fixes... again pretty normal stuff by now.

### DIFF
--- a/sixtyfps_compiler/layout.rs
+++ b/sixtyfps_compiler/layout.rs
@@ -164,20 +164,20 @@ impl LayoutConstraints {
             *op = Some(NamedReference::new(element, prop))
         };
         let e = element.borrow();
-        e.bindings.get("height").map(|s| {
+        if let Some(s) = e.bindings.get("height") {
             constraints.fixed_height = true;
             apply_size_constraint("height", s, &mut constraints.min_height);
             apply_size_constraint("height", s, &mut constraints.max_height);
-        });
-        e.bindings.get("width").map(|s| {
+        }
+        if let Some(s) = e.bindings.get("width") {
             if s.expression.ty() == Type::Percent {
                 apply_size_constraint("width", s, &mut constraints.min_width);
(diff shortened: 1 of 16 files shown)


sixtyfps_compiler/layout.rs
```diff
@@ -227,14 +227,30 @@ impl LayoutConstraints {
...
 
     pub fn visit_named_references(&mut self, visitor: &mut impl FnMut(&mut NamedReference)) {
-        self.max_width.as_mut().map(|e| visitor(&mut *e));
-        self.min_width.as_mut().map(|e| visitor(&mut *e));
-        self.max_height.as_mut().map(|e| visitor(&mut *e));
-        self.min_height.as_mut().map(|e| visitor(&mut *e));
-        self.preferred_width.as_mut().map(|e| visitor(&mut *e));
-        self.preferred_height.as_mut().map(|e| visitor(&mut *e));
-        self.horizontal_stretch.as_mut().map(|e| visitor(&mut *e));
-        self.vertical_stretch.as_mut().map(|e| visitor(&mut *e));
```
ogoffart, line 237: /me liked that code much better before

ogoffart (approved):
looks ok, although i really dislike that clippy don't like me to use Option::map that returns ()

If it was just me, i'd disable that lint globaly

sixtyfps_compiler/layout.rs
hunger: Yeap, that one indeed got more ugly!

I am actually surprised how little difference this makes in most places though! I had expected this particular lint to cause more ugliness in your code base (and I have never seen this `foo.map(||->() {...}` in any other project before.

hunger:
So what about this PR? Should I drop the `map()` to `if let Some(...)` change and replace it with a global override?

I have never seen this trick in the wild, so it at least is not very common and might cause some users to wonder:-)

PR #314, opened 2021-07-12
Scripts: Generator for skeletons of weekly reports

hunger:
A bare bodes script that will just generate a html file (dated tomorrow)
containing the commits of the last seven days.

Use this as a skeleton for the "this week in sixtyfps" reports.

PR #329, opened 2021-07-18
Scripts: Generator for skeletons of weekly reports

hunger:
A bare bodes script that will just generate a html file (dated tomorrow)
containing the commits of the last seven days.

Use this as a skeleton for the "this week in sixtyfps" reports.

This is a continuation of #314, I just can not push updates anymore there since I was lazy and used the master branch:-(

### DIFF
--- a/scripts/sixtyfps_weekly.py
+++ b/scripts/sixtyfps_weekly.py
@@ -1,0 +1,222 @@
+#!/usr/bin/python3
+
+# LICENSE BEGIN
+# Copyright (c) 2021 Tobias Hunger <tobias.hunger@gmail.com>
+#
+# SPDX-License-Identifier: GPL-3.0-only
+# This file is also available under commercial licensing terms.
+# Please contact info@sixtyfps.io for more information.
+# LICENSE END
+
+import datetime
+from re import sub
+import subprocess
+import typing
+
+
+def parse_message(input: str) -> typing.Tuple[str, str]:
+    subject = ""
+    message = ""
+
+    for l in input.rstrip().split("\n"):
+        if not l.startswith("    "):
+            continue
+        if subject == "":
+            subject = l[4:]
+            continue
+
+        message += f"{l[4:]}\n"
+
+    return (subject, message.strip())
+
+
+def get_git_log(
+    start_date: datetime.datetime, end_date: datetime.datetime
+) -> typing.Sequence[typing.Tuple[str, str, str, str]]:
+    result = subprocess.run(
+        [
+            "/usr/bin/git",
+            "log",
+            f"--since={start_date.isoformat()}",
+            f"--until={end_date.isoformat()}",
+            "--pretty=format:%w(0)Subject: %s%nAuthor: %ae%nHash: %H%nBody:%n%w(80,4,4)%b%n%w(0)END_OF_BODY%n%n",
+        ],
+        capture_output=True,
+        check=True,
+    )
+
+    return _parse_git_log(result.stdout.decode("utf-8"))
+
+
+def _parse_git_log(log_data: str) -> typing.Sequence[typing.Tuple[str, str, str, str]]:
+    result: typing.List[typing.Tuple[str, str, str, str]] = []
+
+    author = ""
+    subject = ""
+    commit = ""
+    message = ""
+
+    for l in log_data.split("\n"):
+        if l == "":
+            message += "\n"
+        elif l.startswith("Subject: "):
+            subject = l[9:].strip()
+        elif l.startswith("Author: "):
+            author = l[8:].strip()
+        elif l.startswith("Hash: "):
+            commit = l[6:].strip()
+        elif l.startswith("Body:"):
+            message = ""
+        elif l.startswith("    "):
+            message = f"{message}{l[4:]}\n"
+        elif l.startswith("END_OF_BODY"):
+            message = message.strip()
+            message_str = message
+            message_str.replace("\n", "\\\n")
+
+            assert commit and author and subject
+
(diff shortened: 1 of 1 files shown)


scripts/sixtyfps_weekly.py
```diff
@@ -0,0 +1,141 @@
...
+    return "</body>"
+
+
+end_date = datetime.datetime.now().replace(hour=23, minute=59, second=59, microsecond=999999) 
+start_date = (end_date - datetime.timedelta(days=7)).replace(hour=0, minute=0, second=0, microsecond=0)
+
+result = subprocess.run(
+    [
+        "/usr/bin/git",
+        "log",
```
ogoffart, line 126: I'm not sure if git log is meant to be parsed.  But there are options to make it with a given format
In fact, you could almost use the `--format` to reformat the whole thing with something like

`--pretty 'format:<li>%s<a href="https://github.com/sixtyfpsui/sixtyfps/commit/%H">%h</a><p>%b</p></li>'`

Well, we'll have to convert the commit to html still, but you get the idea.
important though is the link to the commit.

ogoffart (commented):
I wonder if this should not be in the website repository instead.
We might want to aggregate the commit from different repository (for example cargo-ui)

Long term we might want to look more at pull requests instead

scripts/sixtyfps_weekly.py
hunger: I'll add that next time round:-)

hunger:
Newest version uses `git log --pretty=...` for easier parsing.

It also adds links to the individual commits and "linkifies" #BUGIDs and turns `\`` into `<code>`-sections.

matklad:
drive-by comment: in rust-analyzer, we [do this stuff in Rust](https://github.com/rust-analyzer/rust-analyzer/blob/0bee7cb71634800a4ce66120d9dea63aa52e3f81/xtask/src/release/changelog.rs). The overall experiences is that using Rust for such things is a maintenance win long term -- having only one language reduces accidental complexity. 

OTOH, 60fps is by design cross-language, and you will grow python bindings at some stage anyway, so there's no additional tax for adding an extra language in this case.

hunger:
@matklad Thank you for your feedback! I might actually steal some of that:-)

I have to admit, that I just defaulted to python since that is what I use a lot at this time and it felt like a good fit for the task. Switching over to rust mid-term does make a lot of sense. I'll probably want to stick with python for a while longer though: So far I am not entirely sure what this report should look like and python feels "simpler" to just hack something together.

matklad:
Couple of more links then, just to give you inspiration!

* https://github.com/matklad/cargo-xtask
* https://github.com/rust-analyzer/rust-analyzer/blob/0bee7cb71634800a4ce66120d9dea63aa52e3f81/xtask/src/main.rs

tronical:
I love the xtask concept as well as the xshell crate - super easy to use and great for "scripting" with a type system behind.

We use `cargo xtask` for similar things, such as C++ docs, nodejs packaging, other "scripty" things. But it's a matter of taste, of course :-)

scripts/sixtyfps_weekly.py
```diff
@@ -0,0 +1,195 @@
...
+            subject = ""
+            message = ""
+        else:
+            print(f'ERROR: Unknown line "{l}".')
+
+    return result
+
+
+def generate_header(start_date: datetime.datetime, end_date: datetime.datetime) -> str:
+    start_str = f"{start_date.month}.{start_date.day}.{start_date.year}"
```
ogoffart, line 91: mm.dd.yyyy  id not a valid date format in any country.
We used `from ddth of mmmmm to ddth of mmmmm yyyy`

scripts/sixtyfps_weekly.py
```diff
@@ -0,0 +1,195 @@
...
+    timestamp_str = f"{end_date.year}-{end_date.month:02}-{end_date.day:02}T{end_date.hour:02}:{end_date.minute:02}:{end_date.second:02}+01:00"
+    return f"""<!DOCTYPE html>
+<html lang="en">
+
+<head>
+    <meta name="description" content="Changelog for {end_str}" />
+    <title>{start_str} to {end_str}</title>
+</head>
+<!--
+    <date>{timestamp_str}</date>
```
ogoffart, line 102: this should be the monday, actually

hunger: Done.

scripts/sixtyfps_weekly.py
hunger: Yes, you are right. It should look a bit better now.

scripts/sixtyfps_weekly.py
```diff
@@ -0,0 +1,222 @@
...
+    n = int(n)
+    suffix = ['th', 'st', 'nd', 'rd', 'th'][min(n % 10, 4)]
+    if 11 <= (n % 100) <= 13:
+        suffix = 'th'
+    return str(n) + suffix
+
+
+def date_string(date: datetime.datetime) -> str:
+    month_names = [ "INVALID", "January", "February", "March", "April", "May", "June", "July", "August", "September", "October", "November", "December" ]
+    return "{0} of {1} of {2}".format(make_ordinal(date.day), month_names[date.month], date.year)
```
ogoffart, line 114:
```suggestion
    return "{0} of {1} {2}".format(make_ordinal(date.day), month_names[date.month], date.year)
```

Also the start date don't include the year.  One should be consistent with what we already done

ogoffart (commented):
I'd rather put this script in the website repo.

hunger:
I pushed this into the website repo now, so I will close this PR.

PR #336, opened 2021-07-20
Add a as_any to the Model trait to allow getting a reference to the original model

ogoffart:
This should help to get back the original model from the component.

For example, it will remove the need in this hack to use a thread local 

https://github.com/sixtyfpsui/cargo-ui/blob/cefb0b2bca658290c0b27d9cff48e365c3dfbebc/src/main.rs#L101

because we can just get the model back from the component and downcast it to the right value.

### DIFF
--- a/sixtyfps_runtime/corelib/model.rs
+++ b/sixtyfps_runtime/corelib/model.rs
@@ -85,6 +85,30 @@ pub trait Model {
     {
         ModelIterator { model: self, row: 0 }
     }
+
+    /// Return something that can be downcast'ed (typically self)
+    ///
+    /// This is useful to get back to the actual model from a ModelHandle stored
+    /// in a component.
+    ///
+    /// ```
+    /// # use sixtyfps_corelib::model::*;
+    /// # use std::rc::Rc;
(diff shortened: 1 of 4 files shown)


sixtyfps_runtime/corelib/model.rs
```diff
@@ -85,6 +85,27 @@ pub trait Model {
...
+    /// # use std::rc::Rc;
+    /// let vec_model = Rc::new(VecModel::from(vec![1i32, 2, 3]));
+    /// let handle = ModelHandle::from(vec_model as Rc<dyn Model<Data = i32>>);
+    /// // later:
+    /// handle.as_any().downcast_ref::<VecModel<i32>>().unwrap().push(4);
+    /// assert_eq!(handle.row_data(3), 4);
+    /// ```
+    ///
+    /// Note: the default implementation return nothing and this method should be implemented
+    /// by model implementation to return something usefull
```
tronical, line 105:
```suggestion
    /// Note: the default implementation returns nothing and this method should be implemented
    /// by model implementation to return something useful
```

That said, is it possible to give more concrete advice here, i.e. suggest the typical body that people should insert?

tronical (approved):

PR #337, opened 2021-07-20
updated erroneous link

mwaitzman:
The link to the blog post previously pointed to the root of the repo's docs folder.


tronical (approved):
Ooops, well spotted - thanks :)

I guess this link could also be written plain (without the `[foo](bar)`) and GitHub would automatically turn it into a link. But this certainly works.

ogoffart:
The official markdown way to do it is to have the link within `<https://...>`.  Github also hightlight plain link.  (but that's not the case for rustdoc for example)

PR #339, opened 2021-07-20
Added icon property to the Window element

ogoffart:


sixtyfps_runtime/rendering_backends/gl/graphics_window.rs
```diff
@@ -466,10 +467,20 @@ impl PlatformWindow for GraphicsWindow {
             GraphicsWindowBackendState::Unmapped => {}
             GraphicsWindowBackendState::Mapped(window) => {
                 let title = window_item.title();
+                let icon = window_item.icon();
                 let mut size: LogicalSize<f64> = {
                     let backend = window.backend.borrow();
                     let winit_window = backend.window();
                     winit_window.set_title(&title);
+                    if let Some(img) = crate::images::CachedImage::new_from_resource((&icon).into())
+                        .and_then(|i| i.into_image())
```
tronical, line 476: This would always load the image from disk when showing the window. Maybe this could use the cache that we have (untested!):

```suggestion
                    crate::IMAGE_CACHE.with(|image_cache| {
                        image_cache
                        .borrow_mut()
                        .load_image_resource(&icon).into())
                    }).and_then(|i| i.into_image())
```

docs/builtin_elements.md
```diff
@@ -41,6 +41,7 @@ or smaller. The initial width can be controlled with the `preferred-width` prope
 ### Properties
 
 * **`title`** (*string*): The window title that is shown in the title bar.
+* **`icon`** (*image*): The window icon shown in the title bar and the task bar.
```
tronical, line 44:
```suggestion
* **`icon`** (*image*): The window icon shown in the title bar and the task bar. This is only available on Windows and the X11 and Wayland windowing systems.
```

sixtyfps_runtime/rendering_backends/gl/graphics_window.rs
ogoffart: the problem is that the image cache turns it into a femtovg texture, does it not?  What i really want is a RGB thing. 
It only reloads it when the title or the width or height becomes dirty

tronical: `crate::IMAGE_CACHE` will only create decoded images on the CPU side. femtovg textures require a window, GL context, etc., which is why the `GraphicsWindow` has a `texture_cache` member.

tronical (approved):
(of course my suggestions aren't in the way of submitting this, forgot to press approval :)

docs/builtin_elements.md
ogoffart: The property is available everywhere, but it only is shown on supported system

tronical: Ah yes, I'll edit the suggestion to reflect that. My point is that I think that it's worth mentioning limitations like that.

PR #340, opened 2021-07-20
GL backend: Use the cache for the Window icon

ogoffart:
I had to change a bit  the internal image API to avoid copying too much.


tronical (approved):
Nice :)

Issue #333, opened 2021-07-20
Expose more window API


ogoffart:
Introduce a `sixtyfps::Window` which expose some of the window API from the backend (`winit::Window` or `QWindow`)
Some features of the window that relates to the UI itself can be done on the .60 `Window` element in a declarative way:

 - [x] Window title  (already done with `title` property)
 - [ ] Window icon
 - [ ] Scale factor (read only?)
 - [ ] cursor (not a window property, more like an element property)

The more imperative UI can be done via API:
One would need to expose what is currently the `ComponentWindow` into a public `sixtyfps::Window` which has function to do a bunch of things.

This include:
 - [ ] Setting the window state: maximized, minimized, fullscreen, normal (#325)
 - [ ] Setting some windows flag (e.g stay_always_on_top (#324), or request_attention, or border_less) 
 - [ ] Setting the geometry (#323)
 - [ ] Show/hide which are currently on the component, but also allow setting a callback when the window is requesting to be closed in order to show another dialog and cancel and a getter to know if it is currently shown or hidden.

The `items::Window` need to be renamed `WindowItem` to avoid duplicated name (just like `ImageItem`)

PR #341, opened 2021-07-20
Rename the implementation of the `Window` item to `WindowItem`

tronical:
If we were to add `sixtyfps::window::Window` to the re_exports, then
this clashes. We might rename the former, but this is a cleaner naming
in any case.

Relates to #333


ogoffart (approved):
(looks like you need to do cargo fmt for the CI)

PR #342, opened 2021-07-20
window api cleanup

tronical:
This reduces the API surface of ComponentWindow, so that we can rename it to `sixtyfps::Window` for example.

### DIFF
--- a/api/sixtyfps-cpp/lib.rs
+++ b/api/sixtyfps-cpp/lib.rs
@@ -29,7 +29,7 @@ pub unsafe extern "C" fn sixtyfps_component_window_init(out: *mut ComponentWindo
         core::mem::size_of::<ComponentWindow>(),
         core::mem::size_of::<ComponentWindowOpaque>()
     );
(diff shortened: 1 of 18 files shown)


tronical:
The path forward I was thinking of is to basically remove `ComponentWindow`, re-use `Rc<corelib::window::Window>` inside the corelib, backend and interpreter crates and have a real API facing `sixtyfps::Window` in the Rust API crate. Only in *that* location we have access to the `Rc<corelib::window::Window>` as well as the backend factory functions to create them, so that for the public `sixtyfps::Window` there can be a `Default` impl, for example.

ogoffart (approved):

ogoffart:
How about the C++ API ?

tronical:
> How about the C++ API ?

It would be the same there. So `sixtyfps::Window` could become the new public type, which wraps a `WindowRcOpaque` (currently `ComponentWindowOpaque`).

I'll merge this and hopefully in the next request it'll become easier to see.

PR #346, opened 2021-07-22
Start a new sixtyfps::Window API for Rust, C++, the interpreters and JS

tronical:
The generated component now provides access to a Window type
via the window() accessor function.

This is part of #333

### DIFF
--- a/sixtyfps_compiler/generator/cpp.rs
+++ b/sixtyfps_compiler/generator/cpp.rs
@@ -900,8 +900,9 @@ fn generate_component(
         component_struct.members.push((
             Access::Private,
             Declaration::Var(Var {
-                ty: "sixtyfps::private_api::WindowRc".into(),
-                name: "window_rc".into(),
+                ty: "sixtyfps::Window".into(),
+                name: "m_window".into(),
+                init: Some("sixtyfps::Window{sixtyfps::private_api::WindowRc()}".into()),
                 ..Var::default()
             }),
         ));
@@ -909,8 +910,9 @@ fn generate_component(
         component_struct.members.push((
             Access::Public, // FIXME: many of the different component bindings need to access this
             Declaration::Var(Var {
-                ty: "sixtyfps::private_api::WindowRc".into(),
-                name: "window_rc".into(),
+                ty: "sixtyfps::Window".into(),
+                name: "m_window".into(),
+                init: Some("sixtyfps::Window{sixtyfps::private_api::WindowRc()}".into()),
                 ..Var::default()
             }),
         ));
@@ -932,7 +934,7 @@ fn generate_component(
             Declaration::Function(Function {
                 name: "show".into(),
                 signature: "() const".into(),
-                statements: Some(vec!["window_rc.show();".into()]),
+                statements: Some(vec!["m_window.show();".into()]),
                 ..Default::default()
             }),
         ));
@@ -942,7 +944,7 @@ fn generate_component(
             Declaration::Function(Function {
                 name: "hide".into(),
                 signature: "() const".into(),
-                statements: Some(vec!["window_rc.hide();".into()]),
+                statements: Some(vec!["m_window.hide();".into()]),
                 ..Default::default()
             }),
         ));
@@ -951,8 +953,8 @@ fn generate_component(
             Access::Public,
             Declaration::Function(Function {
                 name: "window".into(),
-                signature: "() const -> sixtyfps::Window".into(),
-                statements: Some(vec!["return sixtyfps::Window(window_rc);".into()]),
+                signature: "() const -> const sixtyfps::Window&".into(),
+                statements: Some(vec!["return m_window;".into()]),
                 ..Default::default()
             }),
         ));
@@ -971,7 +973,7 @@ fn generate_component(
             }),
         ));
 
-        init.push("self->window_rc.init_items(this, item_tree());".into());
+        init.push("self->m_window.window_handle().init_items(this, item_tree());".into());
 
         component_struct.friends.push("sixtyfps::private_api::WindowRc".into());
     }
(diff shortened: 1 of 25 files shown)


tronical:
Does this look like an okay direction? A value based type, with explicit sharing behind the scenes, and not constructible but only accessible via `component.window()`.

sixtyfps_runtime/corelib/window.rs
```diff
@@ -339,6 +339,44 @@ impl WindowHandleAccess for WindowRc {
...
+
+        /// De-registers the window from the windowing system, therefore hiding it.
+        pub fn hide(&self) {
+            use super::WindowHandleAccess;
+            self.0.window_handle().hide();
+        }
+    }
+
+    #[doc(hidden)]
+    impl super::WindowHandleAccess for Window {
```
ogoffart, line 373: if this is not meant to be in the api, maybe move this impl to the `super` module

sixtyfps_compiler/generator/cpp.rs
```diff
@@ -896,7 +896,7 @@ fn generate_component(
             Access::Private,
             Declaration::Var(Var {
                 ty: "sixtyfps::private_api::WindowRc".into(),
-                name: "window".into(),
+                name: "window_".into(),
```
ogoffart, line 899: I assume you can't call it `window` because of the new `window()` function.  
But maybe it can have a more descriptive name: `window_rc`, `m_window`, `component_window`, `window_handle`, `window_inner`
What do you think?

ogoffart:
>  A value based type, with explicit sharing behind the scenes

Yes, that's the right thing to do.  
But maybe `Window` is not the right name, since `window.clone()`  does not create another window.
Maybe `WindowHandle` ? 
Or manage to only  return a `&Window`  that can't be cloned?  Is there a reason why one would want to keep a handle around?

sixtyfps_runtime/corelib/window.rs
tronical: You're right, I'll move it.

sixtyfps_compiler/generator/cpp.rs
tronical: Good point. I thought of `m_window` but realised we don't use the `m_` prefix, but for some reason (lack of coffee?) I couldn't see the other options. I'll go with `window_rc`.

tronical:
> > A value based type, with explicit sharing behind the scenes
> 
> Yes, that's the right thing to do.
> But maybe `Window` is not the right name, since `window.clone()` does not create another window.
> Maybe `WindowHandle` ?
> Or manage to only return a `&Window` that can't be cloned? Is there a reason why one would want to keep a handle around?

I don't think there's a reason to keep the window around that's returned by an instance.

In the future I like the idea of `Window` as type-name to facilitate the use-case of re-using a window by being able to write:

```rust
let window = sixtyfps::Window::default();
// ...
let instance = component_compiler.create_with_window(window);
window.show();

// reload
let new_instance - component_compiler.create_with_window(window);
```

There's a bit more work behind that though - among others it will require moving the `Window` out of that `api` module and into the public API, which in turn makes it inaccessible for the interpreter.

I'll try and see if there's a way to return a `&Window`.

sixtyfps_compiler/generator/cpp.rs
```diff
@@ -942,7 +944,17 @@ fn generate_component(
...
+                statements: Some(vec!["m_window.hide();".into()]),
+                ..Default::default()
+            }),
+        ));
+
+        component_struct.members.push((
+            Access::Public,
+            Declaration::Function(Function {
+                name: "window".into(),
+                signature: "() const -> const sixtyfps::Window&".into(),
```
tronical, line 956: This is a little awkward, in the sense that the window accessor is `const` and returns a `const &`, which means that functions like `set_position` would also have to be `const`.

I think a more idiomatic C++ API would be actually a `std::shared_ptr<sixtyfps::Window>`. So "clone" like before, but it's visible in the type name that it's explicitly shared and thus cloning won't create a new window on the screen. What do you think?

ogoffart: how about returning a  `sixtyfps::Window&`  (not const)

ogoffart: Or a  ` sixtyfps::Window*`

tronical: Yes, returning a mutable reference is what I'm going to do now.

sixtyfps_compiler/generator/cpp.rs
```diff
@@ -953,7 +953,7 @@ fn generate_component(
             Access::Public,
             Declaration::Function(Function {
                 name: "window".into(),
-                signature: "() const -> const sixtyfps::Window&".into(),
+                signature: "() -> sixtyfps::Window&".into(),
```
ogoffart, line 956: I think it's ok to return a non-const reference from a `const` function.
In this case, i'd say that the window function should be const because it does not change the component.

The same could be said about hide and show, they don't change the component, they could be const.

tronical: Ah yes, you're right, I'll change the accessor to be `const` again. I'd like to keep `show()`/`hide()` and `run()` as non-const unless you object. `show()` and `hide()` change the visibility state (quite... visibly :-) and `run()` is all about changing the state.

ogoffart: You're missing the `const_cast` 
(or the `mutable`)

tronical: 🤦‍♂️ You're totally right (as usual). My C++ is becoming rusty :-). Added as commit to this series now.

ogoffart (approved):

PR #347, opened 2021-07-22
Pr/janitor

hunger:
More janitor patches

### DIFF
--- a/sixtyfps_runtime/rendering_backends/gl/lib.rs
+++ b/sixtyfps_runtime/rendering_backends/gl/lib.rs
@@ -1360,7 +1360,7 @@ impl GLItemRenderer {
                                         .map(Rc::new)
                                 })
                         })
-                        .map(|image| ItemGraphicsCacheEntry::Image(image))
+                        .map(ItemGraphicsCacheEntry::Image)
                         .map(|cache_entry| self.colorize_image(cache_entry, colorize_property))
                 });
 
--- a/sixtyfps_runtime/rendering_backends/gl/lib.rs
+++ b/sixtyfps_runtime/rendering_backends/gl/lib.rs
@@ -43,7 +43,7 @@ mod fonts;
 
 type CanvasRc = Rc<RefCell<femtovg::Canvas<femtovg::renderer::OpenGl>>>;
 
-const KAPPA90: f32 = 0.5522847493;
+const KAPPA90: f32 = 0.55228;
 
 #[derive(Clone)]
 enum ItemGraphicsCacheEntry {
--- a/sixtyfps_runtime/rendering_backends/gl/lib.rs
+++ b/sixtyfps_runtime/rendering_backends/gl/lib.rs
@@ -20,6 +20,7 @@ use std::cell::RefCell;
 use std::pin::Pin;
 use std::rc::Rc;
 
+use euclid::approxeq::ApproxEq;
 use sixtyfps_corelib::graphics::{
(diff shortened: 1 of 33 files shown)


sixtyfps_runtime/rendering_backends/qt/qt_window.rs
```diff
@@ -889,8 +889,8 @@ impl QtItemRenderer<'_> {
                 rect.is_valid()
                     && (rect.x != 0.
                         || rect.y != 0.
-                        || rect.width != target_width
-                        || rect.height != target_height)
+                        || (rect.width - target_width).abs() > 100. * f64::EPSILON
+                        || (rect.height - target_height).abs() > 100. * f64::EPSILON)
```
tronical, line 893: I wonder if we should start using euclid for these comparisons as we start introducing different factors. Concretely I'm thinking of https://docs.rs/euclid/0.22.6/euclid/approxeq/trait.ApproxEq.html , so like this:

```suggestion
                        || !rect.width.approx_eq(&target_width)
                        || !rect.height.approx_eq(&target_height))
```

(this already compiles since we depend on euclid)

What do you think?

sixtyfps_runtime/rendering_backends/gl/lib.rs
```diff
@@ -43,7 +43,7 @@ mod fonts;
 
 type CanvasRc = Rc<RefCell<femtovg::Canvas<femtovg::renderer::OpenGl>>>;
 
-const KAPPA90: f32 = 0.5522847493;
+const KAPPA90: f32 = 0.55228;
```
ogoffart, line 46: on the other hand, it doesn't hurt to have it, does it?

ogoffart (approved):

sixtyfps_runtime/rendering_backends/qt/qt_window.rs
hunger: Sounds great! I'll update that.

sixtyfps_runtime/rendering_backends/gl/lib.rs
hunger: Not really, it just makes people assume that the results will be more precise than they really are.

sixtyfps_runtime/rendering_backends/qt/qt_window.rs
hunger: Done in the latest push: I went in and replaced all the `EPSILON`s I added before:-)

I did so in one patch on top of what I had before.

tools/lsp/main.rs
```diff
@@ -271,7 +272,7 @@ fn handle_request(
         // representation.
         let requested_color = params.color;
 
-        let color_literal = if (requested_color.alpha - 1.).abs() > 10.0 * f32::EPSILON {
+        let color_literal = if requested_color.alpha.approx_eq(&1.) {
```
tronical, line 275: This was originally `!= 1`.

```suggestion
        let color_literal = if !requested_color.alpha.approx_eq(&1.) {
```

ogoffart: I'd go as far as simply say 
```suggestion
        let color_literal = if requested_color.alpha < 1. {
```

hunger: Good catch!

I'll go with @tronical s approach: We want that path to be taken, even if we just have 0.9999999 :-)

These values are potentially the result of a calculation with some rounding error.

PR #351, opened 2021-07-24
Fix float comparison in qt_window.rs

hunger:
Sorry, I broke the float comparison fix earlier. I only noticed when going over the patch again once you had spotted the same issue in a different place already (and fixed it there before merging).

Amends 02bdcbf6fffa38a822ecebcc84e16ad1be92c46e


tronical (approved):

tronical:
Thanks :)

PR #353, opened 2021-07-26
Fix angles conversion and allow angle and duration in the public API

ogoffart:
One of the commit fix the angle conversion.

The other sets of commit make public than angle and duration in the public API.
Rationale for making the duration public is that we document that it is converted to i64 and that it is the number of millisecond
An argument against making it public is that one would want to use some Rust or C++ duration type instead.
Similarly, should we make angle properties public?

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -23,6 +23,7 @@ All notable changes to this project will be documented in this file.
  - Default height of element in a ListView no longer defaults to 100%
  - Support of `*=` and `/=` on types with unit such as length.
  - Proper compilation error when using a self assignment operator on an invalid type instead of a panic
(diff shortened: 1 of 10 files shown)


CHANGELOG.md
```diff
@@ -20,6 +21,7 @@ All notable changes to this project will be documented in this file.
  - Default height of element in a ListView no longer defaults to 100%
  - Support of `*=` and `/=` on types with unit such as length.
  - Proper compilation error when using a self assignment operator on an invalid type instead of a panic
+ - Angle conversion for value not in degree
```
tronical, line 24:
```suggestion
 - Angle conversion for values specified in radians, gradians and turns
```

tronical (approved):

PR #354, opened 2021-07-27
Add support for icons in the Button widget

tronical:
Funny anecdote: When trying to implement the Qt part I probably spent a good twenty minutes debugging why on earth the icon isn't drawn, until I noticed that on macOS buttons don't have icons. On Linux it worked :-)


sixtyfps_runtime/rendering_backends/qt/widgets.rs
```diff
@@ -177,15 +178,23 @@ impl Item for NativeButton {
         _window: &WindowRc,
     ) -> LayoutInfo {
         let mut text: qttypes::QString = self.text().as_str().into();
+        let icon = crate::qt_window::load_image_from_resource(
```
ogoffart, line 181:
```suggestion
        let icon : qttypes::QPixmap = crate::qt_window::load_image_from_resource(
```

I like to have the type specified near the cpp! macro because the cpp macro does only minimal type checking (same size, basically) so settinh the type explicitly make sure that one can quickly see that this is correct.

sixtyfps_runtime/rendering_backends/qt/widgets.rs
```diff
@@ -246,18 +255,26 @@ impl Item for NativeButton {
     fn_render! { this dpr size painter =>
         let down: bool = this.pressed();
         let text: qttypes::QString = this.text().as_str().into();
+        let icon = crate::qt_window::load_image_from_resource(
```
ogoffart, line 258:
```suggestion
        let icon : qttypes::QPixmap = crate::qt_window::load_image_from_resource(
```

ogoffart (approved):

Issue #83, opened 2020-10-12
Select the default style


ogoffart:
Currently the way to select the style is by setting the `SLINT_STYLE` environment variable while compiling (when running the slint-compiler)

 * `SLINT_STYLE=native`  for the native desktop style using Qt 
 * `SLINT_STYLE=fluent`  for our current builtin style
 * `SLINT_STYLE=material`  for the builtin material style

We would like the native style to be the default on desktop, but only if Qt was found while compiling the runtime lib. But since we don't know that when running the compiler, the experience would be quite bad.

The native style require Qt, see https://github.com/slint-ui/slint/blob/master/docs/install_qt.md 

Update: When using the rust integration or the interpreter, the default style is Qt if Qt was found

Update: some extra documentation in https://github.com/slint-ui/slint/blob/master/docs/widgets.md#selecting-a-widget-style

flukejones:
Given that there are two main native styles (QT and GTK), will you be supporting GTK style at all?

ogoffart:
> Given that there are two main native styles (QT and GTK), will you be supporting GTK style at all?

In a GTK-based desktop environment, Qt itself is using the GTK style begind the scene. Using the Qt framework as an abstraction to draw native widget allow us to be looking native on every platform with less work for us at the moment.
(This means we have SixtyFPS -> Qt -> GTK.)

Creating a GTK backend is totally possible, and we might consider doing that, but this is not a priority given that Qt provide a decent native look even in a GTK environment right now.

PR #355, opened 2021-07-28
Rust: defaults to the `native` style when Qt is available

ogoffart:
This requires some gymnastics to get right as the information
need to be passed to the compiler despite having no direct dependency
between the compiler and the runtime or backends.
So use a file in the build directory to tell the default style

cc: #83

### DIFF
--- a/api/sixtyfps-rs/sixtyfps-build/lib.rs
+++ b/api/sixtyfps-rs/sixtyfps-build/lib.rs
@@ -203,6 +203,7 @@ pub fn compile_with_config(
     };
     if std::env::var_os("SIXTYFPS_STYLE").is_none() && compiler_config.style.is_none() {
         compiler_config.style = std::env::var_os("OUT_DIR").and_then(|path| {
+            // Same logic as in sixtyfps-rendering-backend-default's build script to get the path
             let path = Path::new(&path).parent()?.parent()?.join("SIXTYFPS_DEFAULT_STYLE.txt");
             let style = std::fs::read_to_string(path).ok()?;
             Some(style.trim().into())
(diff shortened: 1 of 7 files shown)


sixtyfps_runtime/rendering_backends/qt/build.rs
```diff
@@ -49,4 +49,5 @@ fn main() {
     println!("cargo:rerun-if-changed=qt_window.rs");
     println!("cargo:rerun-if-changed=widgets.rs");
     println!("cargo:rerun-if-changed=lib.rs");
+    println!("cargo:HAS_NATIVE_STYLE=1");
```
tronical, line 52: Maybe it should be called `SUPPORTS_NATIVE_STYLE`, but feel free to ignore

api/sixtyfps-rs/sixtyfps-build/lib.rs
```diff
@@ -201,6 +201,13 @@ pub fn compile_with_config(
             compiler_config.embed_resources = true;
         }
     };
+    if std::env::var_os("SIXTYFPS_STYLE").is_none() && compiler_config.style.is_none() {
+        compiler_config.style = std::env::var_os("OUT_DIR").and_then(|path| {
+            let path = Path::new(&path).parent()?.parent()?.join("SIXTYFPS_DEFAULT_STYLE.txt");
```
tronical, line 206:
```suggestion
            // out_dir is something like
            // <target_dir>/build/$package/out
            // and we want locate the common directory, where the sixtyfps-rendering-backend-default
            // package has written the suggested default
            let path = Path::new(&path).parent()?.parent()?.join("SIXTYFPS_DEFAULT_STYLE.txt");
```

tronical (approved):

tronical:
I think this approach of getting to the common build directory is okay - spoken as a person without shame for such hacks :-)

PR #361, opened 2021-07-29
clippy generated code

tronical:


sixtyfps_compiler/generator/rust.rs
```diff
@@ -119,6 +119,8 @@ pub fn generate(doc: &Document, diag: &mut BuildDiagnostics) -> Option<TokenStre
         .collect::<Vec<_>>();
     Some(quote! {
         #[allow(non_snake_case)]
+        #[allow(clippy::match_single_binding)]
+        #[allow(clippy::deref_addrof)]
```
ogoffart, line 123: I'd go away and exclude a broad category  of lints that don't make sense for generated code:

```suggestion
        #[allow(clippy::style)]
        #[allow(clippy::complexity)]
```

Or do you think we would benefit from some of these lints?

ogoffart (approved):
Eventually, we should test that the generated code produce no lints.

Unfortunately, the current rust driver uses the macro, and the warnings seems suppressed when coming from a macro expansion.

Maybe there is a way to make sure they fire?
Otherwise. We could consider having a second rust driver for rust that tests the code generated and writen to a file.

sixtyfps_compiler/generator/rust.rs
tronical: Excellent idea. TIL clippy categories. That makes a whole lot more sense to me.

tronical:
> Eventually, we should test that the generated code produce no lints.
> 
> Unfortunately, the current rust driver uses the macro, and the warnings seems suppressed when coming from a macro expansion.
> 
> Maybe there is a way to make sure they fire?
> Otherwise. We could consider having a second rust driver for rust that tests the code generated and writen to a file.

I've been wondering about the same, how can we test that we don't accidentally introduce new clippy warnings/errors in generated code?

I don't know if there's a way to make the warnings surface, unless perhaps they can be turned into errors?

hunger:
This does remove a lot of noise.

It might make sense to allow `clippy::approx_constant`. It does things like these errors:

```
error: approximate value of `f64::consts::SQRT_2` found. Consider using it directly
    --> /home/dev/src/sixtyfps/target/debug/build/slide_puzzle-42ede3577183b114/out/slide_puzzle.rs:2625:133
     |
2625 | ...   + r#Rectangle :: FIELD_OFFSETS . r#width) . apply_pin (_self) . get () as f64) * (2f64 as f64)) as f64) * 
(1.4142f64 as f64)) }
     |                                                                                                                 
 ^^^^^^^^^
     |
     = note: `#[deny(clippy::approx_constant)]` on by default
     = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#approx_constant
```

which are probably OK in our use case.

ogoffart:
> What do you think about allowing clippy::approx_constant

We could. We could also write `sqrt(2)` in the slide puzzle code. (This did not exist at the time, I think)

tronical:
> > What do you think about allowing clippy::approx_constant
> 
> We could. We could also write `sqrt(2)` in the slide puzzle code. (This did not exist at the time, I think)

Yeah, that's an issue I see as well with enabling these. We allow arithmetic expressions in .60 that we translate directly to Rust code. If the result produces a clippy warning in generated code, it's may not be obvious to the users where the corresponding `.60` code is - at which point it's more of an annoyance than help.

PR #362, opened 2021-07-29
Implement the visible property

### LINKED ISSUE #141: Implement the `visible` property
This property is assignable but it does nothing for now and could come in handy when developing more complex UIs. Until then I guess that transparent colors can be a workaround.

I can see other properties in the `reserved_property` function. It could be useful to have a warning printed when a reserved but non-implemented property is assigned.

ogoffart:
This is done by adding an empty Clip Item in the item tree, and clipping it when the visible property is false

Closes #141

### DIFF
--- a/sixtyfps_compiler/passes/visible.rs
+++ b/sixtyfps_compiler/passes/visible.rs
@@ -1,0 +1,88 @@
+/* LICENSE BEGIN
+    This file is part of the SixtyFPS Project -- https://sixtyfps.io
+    Copyright (c) 2021 Olivier Goffart <olivier.goffart@sixtyfps.io>
+    Copyright (c) 2021 Simon Hausmann <simon.hausmann@sixtyfps.io>
+
+    SPDX-License-Identifier: GPL-3.0-only
+    This file is also available under commercial licensing terms.
+    Please contact info@sixtyfps.io for more information.
+LICENSE END */
+//! Pass that lowers synthetic `visible` properties to Clip element
+
+use std::cell::RefCell;
+use std::rc::Rc;
+
+use crate::expression_tree::{Expression, NamedReference};
+use crate::langtype::{NativeClass, Type};
+use crate::object_tree::{self, Component, Element, ElementRc};
+use crate::typeregister::TypeRegister;
+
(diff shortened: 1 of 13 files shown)


sixtyfps_compiler/passes/visible.rs
```diff
@@ -0,0 +1,87 @@
...
+
+pub fn handle_visible(component: &Rc<Component>, type_register: &TypeRegister) {
+    let native_clip = type_register.lookup("Clip").as_builtin().native_class.clone();
+
+    crate::object_tree::recurse_elem_including_sub_components(
+        component,
+        &(),
+        &mut |elem: &ElementRc, _| {
+            if elem.borrow().native_class().map_or(false, |n| {
+                Rc::ptr_eq(&n, &native_clip) && elem.borrow().id.ends_with("_visibility")
```
tronical, line 28:
```suggestion
                // Avoid native clips and previously created, synthetic visibility elements
                Rc::ptr_eq(&n, &native_clip) && elem.borrow().id.ends_with("_visibility")
```

sixtyfps_compiler/passes/visible.rs
```diff
@@ -0,0 +1,87 @@
...
+            };
+
+            let has_visible_binding = |e: &ElementRc| {
+                e.borrow().base_type.lookup_property("visible").property_type != Type::Invalid
+                    && (e.borrow().bindings.contains_key("visible")
+                        || e.borrow()
+                            .property_analysis
+                            .borrow()
+                            .get("visible")
+                            .map_or(false, |a| a.is_set))
```
tronical, line 46: Just curious to see if I understand this correctly: Is this condition true when for example another element creates just a two-way binding to "visible"?

Since this is used in two places at least, maybe it's worth a helper function on Element?

tronical (approved):
Great idea to re-use the clip :)

sixtyfps_compiler/passes/visible.rs
ogoffart: > Is this condition true when for example another element creates just a two-way binding to "visible"?

Actually it's not, which is a bug. but i didn't really want to fix that bug now

PR #365, opened 2021-07-29
Pr/janitor

hunger:
Janitor: Fix clippy::map_entry might be a bit controversial since it introduces a string clone that was not necessary before.

### DIFF
--- a/sixtyfps_compiler/passes/lower_layout.rs
+++ b/sixtyfps_compiler/passes/lower_layout.rs
@@ -202,9 +202,8 @@ impl GridLayout {
         if let Some(layout_item) = create_layout_item(item_element, diag) {
             if layout_item.repeater_index.is_some() {
                 diag.push_error(
-                    format!(
-                        "'if' or 'for' expressions are not currently supported in grid layouts"
-                    ),
+                    "'if' or 'for' expressions are not currently supported in grid layouts"
+                        .to_string(),
                     &*item_element.borrow(),
                 );
                 return;
--- a/sixtyfps_compiler/passes/lower_layout.rs
+++ b/sixtyfps_compiler/passes/lower_layout.rs
@@ -9,6 +9,8 @@
 LICENSE END */
 //! Passe that compute the layout constraint
 
+use lyon_path::geom::euclid::approxeq::ApproxEq;
+
 use crate::diagnostics::BuildDiagnostics;
 use crate::diagnostics::Spanned;
 use crate::expression_tree::*;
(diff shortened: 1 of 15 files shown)


sixtyfps_compiler/passes/lower_layout.rs
```diff
@@ -502,7 +504,7 @@ fn eval_const_expr(
 ) -> Option<u16> {
     match expression {
         Expression::NumberLiteral(v, Unit::None) => {
-            if *v < 0. || *v > u16::MAX as f64 || v.trunc() != *v {
+            if *v < 0. || *v > u16::MAX as f64 || !v.trunc().approx_eq(v) {
```
ogoffart, line 507: I've seen this one before. Hasn't this already been done? Or is that duplicate code?

ogoffart (approved):
Thanks

sixtyfps_compiler/passes/lower_layout.rs
```diff
@@ -9,6 +9,8 @@
 LICENSE END */
 //! Passe that compute the layout constraint
 
+use lyon_path::geom::euclid::approxeq::ApproxEq;
```
tronical, line 12: I'd like to suggest to use euclid directly, instead of going through a re-export - unless there's a good reason to prefer the indirection?

sixtyfps_compiler/passes/lower_layout.rs
tronical: Commit 2cdc5848a2b77a52b4c8611410fbfb79c3743ea7 changed this 

> I changed the code to decide whether a number is a positive integer to
> make do without a comparison along the way.

```patch
diff --git a/sixtyfps_compiler/passes/lower_layout.rs b/sixtyfps_compiler/passes/lower_layout.rs
index 6d538546..92b94475 100644
--- a/sixtyfps_compiler/passes/lower_layout.rs
+++ b/sixtyfps_compiler/passes/lower_layout.rs
@@ -493,12 +493,11 @@ fn eval_const_expr(
 ) -> Option<u16> {
     match expression {
         Expression::NumberLiteral(v, Unit::None) => {
-            let r = *v as u16;
-            if (r as f32 - *v as f32).abs() > f32::EPSILON {
+            if *v < 0. || *v > u16::MAX as f64 || v.trunc() != *v {
                 diag.push_error(format!("'{}' must be a positive integer", name), span);
                 None
```

and accidentally introduced this comparison :).

tronical (approved):

PR #370, opened 2021-07-30
Fix cursor rendering with the Qt backend for non-ascii text

### LINKED ISSUE #363: The cursor position is wrong when there are CJK characters in TextInput
cjk-input

Environment: Linux on X11
Font: Noto Snas CJK SC

In fact, there is more than one problem on CJK. I did not find any documentation about CJK. Is there any plan to support CJK in the future?

I am looking forward to the development of this library! 👍

tronical:
We calculate the cursor position as byte offset in the utf-8 encoded string,
while Qt expects an index in the utf-16 encoded QString.
When those differ, the cursor is rendered at the wrong location.

Fixes #363

### DIFF
--- a/sixtyfps_runtime/rendering_backends/qt/qt_window.rs
+++ b/sixtyfps_runtime/rendering_backends/qt/qt_window.rs
@@ -472,7 +472,8 @@ impl ItemRenderer for QtItemRenderer<'_> {
         let selection_background_color: u32 =
             text_input.selection_background_color().as_argb_encoded();
 
-        let string: qttypes::QString = text_input.text().as_str().into();
+        let text = text_input.text();
+        let string: qttypes::QString = text.as_str().into();
         let font: QFont =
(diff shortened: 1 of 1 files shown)


sixtyfps_runtime/rendering_backends/qt/qt_window.rs
```diff
@@ -1504,3 +1521,34 @@ pub(crate) mod ffi {
...
+
+#[test]
+fn test_utf8_byte_offset_to_utf16_units() {
+    assert_eq!(utf8_byte_offset_to_utf16_units("Hello", 2), 2);
+
+    {
+        let test_str = "a🚀🍌";
+        assert_eq!(test_str.encode_utf16().count(), 5);
+
+        let banana_offset = test_str.char_indices().skip(2).next().unwrap().0;
```
ogoffart, line 1546: i love that variable name :-)

ogoffart (approved):

PR #374, opened 2021-07-31
Mention forward-focus in the FocusScope example

jturcotte:
This makes the example a bit more useful out of the box,
at least to let the user know that this property exists.

### DIFF
--- a/docs/builtin_elements.md
+++ b/docs/builtin_elements.md
@@ -463,7 +463,8 @@ The FocusScope exposes callback to intercept the pressed key when it has focus.
 
 ```60
 Example := Window {
-    FocusScope {
+    forward-focus: my_key_handler;
+    my_key_handler := FocusScope {
         key-pressed(event) => {
             debug(event.text);
             if (event.modifiers.control) {


jturcotte:
My first reflex was to try `has_focus: true;`, but it seems like this property is read-only? Or not at least setting it has no effect?

If so, it would be nice to know what property is read only or not in the documentation or as a compilation error when trying to write it.

docs/builtin_elements.md
```diff
@@ -463,7 +463,8 @@ The FocusScope exposes callback to intercept the pressed key when it has focus.
 
 ```60
 Example := Window {
-    FocusScope {
+    forward-focus: MyKeyHandler;
+    MyKeyHandler := FocusScope {
```
ogoffart, line 467:
```suggestion
    forward-focus: my_key_handler;
    my_key_handler := FocusScope {
```

Our current coding style kind of use lower case for id.  (but we didn't really define that anywhere though)

tronical, line 467: Good point. Maybe we should make that automatic: If one of element's children is focusable (because it's a FocusScope or uses forward-focus itself), then this should perhaps imply forwarding of focus to that child.

ogoffart: Maybe indeed.

Or at least the error message when trying to setting the focus to it could hint at the forward-focus property

jturcotte: Done

jturcotte: Could you make it recursive?

Also if an element has more than one focusable child, would it take the first one?
Else adding a new focusable element would have the unexpected side effect of removing the automatic focus if the previously sole focusable element. If this happens deep in the tree, this could be an hard-to-catch regression.

jturcotte: But then what would happen if I call `focus()` on the second element or one of its descendants while the first sibling has an implicit or explicit forward-focus?

PR #375, opened 2021-08-02
Minor reduction in generated Rust code (llvm lines)

tronical:
When setting a constant, we only need to use a closure for the
expression if it contains a return statement.

This saves about ~14000 llvm lines for a debug build of the printer
demo.

### DIFF
--- a/sixtyfps_compiler/generator/rust.rs
+++ b/sixtyfps_compiler/generator/rust.rs
@@ -214,7 +214,19 @@ fn handle_property_binding(
         let tokens_for_expression = compile_expression(binding_expression, component);
         init.push(if is_constant {
             let t = rust_type(&prop_type).unwrap_or(quote!(_));
-            quote! { #rust_property.set((||-> #t { (#tokens_for_expression) as #t })()); }
+
+            let mut uses_return = false;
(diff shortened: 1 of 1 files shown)


sixtyfps_compiler/generator/rust.rs
```diff
@@ -214,7 +214,19 @@ fn handle_property_binding(
         let tokens_for_expression = compile_expression(binding_expression, component);
         init.push(if is_constant {
             let t = rust_type(&prop_type).unwrap_or(quote!(_));
-            quote! { #rust_property.set((||-> #t { (#tokens_for_expression) as #t })()); }
+
+            let mut uses_return = false;
```
ogoffart, line 218:
```suggestion
            // When there is a `return` statement, we must use a lambda expression in the generated code so that the 
            // generated code can have an actual return in it. We only want to do that if necessary because otherwise
            // this would slow down the rust compilation
            let mut uses_return = false;
```

I just think we can add a comment saying this is about the rust compilation, because otherwise one might "optimize" this code by reverting the patch again.  Feel free to change my wording.

ogoffart (approved):

PR #378, opened 2021-08-02
Fix disappearing elements with implicit width/height

### LINKED ISSUE #376: Declaring `height` to be animated causes `Rectangle` to have zero height
In the test-case below the inner Rectangle's height becomes zero when it is declared to be animated:

```
export TestCase := Window {
    preferred-width: 100px;
    preferred-height: 100px;
    r := Rectangle {
        background: blue;
        animate height { duration: 1s; }
    }
    TouchArea {
        clicked => {
            debug(r.height)
        }
    }
}
```

This is also visible in the generated code. With the `animate {}` declaration, no binding for `height` is initialised - otherwise the expected `100%` binding is set up.

tronical:
When an element gets its width and height from the parent through an implicit 100% binding,
those bindings were missing when an animation was pre-defined.

The provided new-type wrapper offers a function to deal with replacing just
binding expression, instead of the
expression *and* the animation.

Fixes #376

### DIFF
--- a/sixtyfps_compiler/expression_tree.rs
+++ b/sixtyfps_compiler/expression_tree.rs
@@ -13,7 +13,7 @@ use crate::layout::Orientation;
 use crate::object_tree::*;
 use crate::parser::{NodeOrToken, SyntaxNode};
 use core::cell::RefCell;
-use std::collections::{BTreeMap, HashMap};
+use std::collections::HashMap;
 use std::rc::{Rc, Weak};
 
 // FIXME remove the pub
@@ -1177,7 +1177,7 @@ pub enum Path {
 #[derive(Debug, Clone)]
 pub struct PathElement {
     pub element_type: Rc<BuiltinElement>,
-    pub bindings: BTreeMap<String, BindingExpression>,
+    pub bindings: BindingsMap,
 }
 
 #[derive(Clone, Debug)]
--- a/sixtyfps_compiler/object_tree.rs
+++ b/sixtyfps_compiler/object_tree.rs
@@ -23,7 +23,9 @@ use crate::parser::{identifier_text, syntax_nodes, SyntaxKind, SyntaxNode};
 use crate::typeloader::ImportedTypes;
 use crate::typeregister::TypeRegister;
 use std::cell::RefCell;
+use std::collections::btree_map::Entry;
 use std::collections::{BTreeMap, HashMap};
+use std::iter::FromIterator;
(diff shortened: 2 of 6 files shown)


ogoffart (approved):

ogoffart:
maybe the change in lower_state in commit e0fad7fbdfdbfa7acc23bcb1e070b29075ac45cf could use a helper function in the binding map to see if the binding has been set.

tronical:
> maybe the change in lower_state in commit [e0fad7f](https://github.com/sixtyfpsui/sixtyfps/commit/e0fad7fbdfdbfa7acc23bcb1e070b29075ac45cf) could use a helper function in the binding map to see if the binding has been set.

Not sure in fully understand (but happy to make changes, of course). Do you mean something like this?

```patch
diff --git a/sixtyfps_compiler/object_tree.rs b/sixtyfps_compiler/object_tree.rs
index 0398ccdb..6323a95e 100644
--- a/sixtyfps_compiler/object_tree.rs
+++ b/sixtyfps_compiler/object_tree.rs
@@ -371,6 +371,16 @@ pub fn set_binding_if_not_set(&mut self, property_name: String, expression: Bind
             }
         };
     }
+
+    pub fn get_binding_expression(&self, property_name: &str) -> Option<&Expression> {
+        self.0.get(property_name).and_then(|binding| {
+            if !matches!(binding.expression, Expression::Invalid) {
+                Some(&binding.expression)
+            } else {
+                None
+            }
+        })
+    }
 }
 
 impl IntoIterator for BindingsMap {
diff --git a/sixtyfps_compiler/passes/lower_states.rs b/sixtyfps_compiler/passes/lower_states.rs
index c5f92121..2fac7ded 100644
--- a/sixtyfps_compiler/passes/lower_states.rs
+++ b/sixtyfps_compiler/passes/lower_states.rs
@@ -181,10 +181,8 @@ fn compute_state_property_name(root_element: &ElementRc) -> String {
 fn expression_for_property(element: &ElementRc, name: &str) -> Expression {
     let mut element_it = Some(element.clone());
     while let Some(element) = element_it {
-        if let Some(e) = element.borrow().bindings.get(name) {
-            if !matches!(e.expression, Expression::Invalid) {
-                return e.expression.clone();
-            }
+        if let Some(e) = element.borrow().bindings.get_binding_expression(name) {
+            return e.clone();
         }
         element_it = if let Type::Component(base) = &element.borrow().base_type {
             Some(base.root_element.clone())
```

PR #384, opened 2021-08-03
GL backend: Avoid run-time opengl errors with clipped zero width or h…

### LINKED ISSUE #377: Clipping on a zero width/height Rectangle causes message to appear in GL renderer
Applying a clip to a Rectangle that has either a zero width or height (or both) causes a message to appear on the console when using the GL renderer:

```
export TestCase := Window {
    preferred-width: 400px;
    preferred-height: 400px;
    Rectangle {  
        clip: true;
        border-radius: 1px;
        height: 0px;
    }            
}
```

```
Error on set_uniforms uniforms - Invalid framebuffer operation
Error on render done - Invalid framebuffer operation
```

This appears to be caused by trying to create a zero width/height FBO (needed for clipping).

tronical:
…eight rectangles

If a Rectangle has a border-radius and clipping, we use an FBO to render
the children and then use femtovg's stencil clipping. If the Rectangle
has a zero width or height, we would end up trying to create a texture
with such dimensions, which produces run-time opengl errors.

We can detect this situation and avoid it early on. The same might happen for shadows.

Fixes #377

### DIFF
--- a/sixtyfps_runtime/rendering_backends/gl/lib.rs
+++ b/sixtyfps_runtime/rendering_backends/gl/lib.rs
@@ -934,7 +934,7 @@ impl ItemRenderer for GLItemRenderer {
             return;
         }
 
-        let cache_entry = self
+        let cache_entry = match self
             .graphics_window
             .graphics_cache
             .borrow_mut()
@@ -961,7 +961,7 @@ impl ItemRenderer for GLItemRenderer {
                         &self.shared_data.canvas,
                         shadow_image_width,
                         shadow_image_height,
-                    );
+                    )?;
 
                     {
                         let mut canvas = self.shared_data.canvas.borrow_mut();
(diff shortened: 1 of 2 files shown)


sixtyfps_runtime/rendering_backends/gl/lib.rs
```diff
@@ -934,7 +934,7 @@ impl ItemRenderer for GLItemRenderer {
             return;
         }
 
-        let cache_entry = self
+        let cache_entry = match self
```
ogoffart, line 937: style: this expression is a bit big and the match arm are far away from the match.  Maybe consider using an intermediate variable.

ogoffart (approved):

sixtyfps_runtime/rendering_backends/gl/lib.rs
tronical: That's a very good point - thanks. I'll change that with an intermediate variable.

PR #386, opened 2021-08-03
Fix regression of two-way binding on default geom'ed properties causi…

### LINKED ISSUE #385: "Recursion detected" panic when using two-way binding on Rectangle's width
The following reduced test-case produces a "Recursion detected" panic at run-time:

```
export TestCase := Rectangle {
    width: 100px;
    Rectangle {
        width <=> root.width;
    }
}
```

tronical:
…ng binding loop

Commit 064c39d625395b9ac74961e4ded7606cdbfe16ba introduced the regression that
if a two-way binding was set on a property that we'd also set a default geometry
on, we'd end up applying that on the two-way binding, causing a binding loop.

set_binding_if_not_set needs to only set the binding if... there's really none yet.

Fixes #385

### DIFF
--- a/sixtyfps_compiler/object_tree.rs
+++ b/sixtyfps_compiler/object_tree.rs
@@ -361,14 +361,21 @@ impl Clone for PropertyAnimation {
 pub struct BindingsMap(BTreeMap<String, BindingExpression>);
 
 impl BindingsMap {
-    pub fn set_binding_if_not_set(&mut self, property_name: String, expression: BindingExpression) {
+    pub fn set_binding_if_not_set(
+        &mut self,
+        property_name: String,
+        expression_fn: impl FnOnce() -> BindingExpression,
+    ) {
         match self.0.entry(property_name) {
             Entry::Vacant(vacant_entry) => {
-                vacant_entry.insert(expression);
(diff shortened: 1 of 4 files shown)


ogoffart (approved):

ogoffart:
Maybe we actually still want to create a default binding (when not actually creating a loop)

tronical:
What would be the use-case for that binding? Something like this?

```
Rectangle {
    Rectangle {
        width: 500px;
        height: 500px;
        Rectangle {
            width <=> syncme.width;
            height <=> syncme.height;
        }
    }
    syncme := Rectangle {
        x: 500px;
    }
}
```

ogoffart:
yes, exactly.  Or just aliased to some other property which is read for whatever reason.
But maybe that's a bit far fetched

tronical:
Maybe it's a fair reason though, if we say that the following two should be equivalent:

```
first := Rectangle {
    width <=> second.width;
}
second := Rectangle {
}
```

and

```
first := Rectangle {
}
second := Rectangle {
    width <=> first.width;
}
```

I mean: I do find it difficult to argue that the first example shall be different than the second one.

PR #388, opened 2021-08-04
Fix clippy errors in generated code of sliding puzzle

hunger:
Disable all the clippy checks that are registered as errors in
the sliding puzzle example.

The following warnings are allowed in this change:

* clippy::approx_constant: This is probably the right thing to do
  as we do not want to get errors for all possible constants defined in
  `.60` files
* clippy::float_cmp: This is probably wrong! We should do the proper
  float comparisson dance for these values.
* clippy::if_same_then_else: Probably ok
* clippy::erasing_op: This is about multiplying indices with 0 in some
  places, which seems to be ok for generated code.

### DIFF
--- a/sixtyfps_compiler/generator/rust.rs
+++ b/sixtyfps_compiler/generator/rust.rs
@@ -126,6 +126,10 @@ pub fn generate(doc: &Document, diag: &mut BuildDiagnostics) -> Option<TokenStre
         #[allow(non_snake_case)]
         #[allow(clippy::style)]
         #[allow(clippy::complexity)]
+        #[allow(clippy::approx_constant)]
+        #[allow(clippy::float_cmp)]
+        #[allow(clippy::if_same_then_else)]
+        #[allow(clippy::erasing_op)]
         mod #compo_module {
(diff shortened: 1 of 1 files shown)


hunger:
Just pushing this to get the discussion started on how to address these problems.

hunger:
These additional `allow`s fix all the generated warnings, too:

```diff
+        #[allow(clippy::redundant_clone)]
+        #[allow(clippy::cmp_owned)]
```

These seem safe to me and hard to avoid in generic generated code.

ogoffart:
> clippy::float_cmp: This is probably wrong! We should do the proper
float comparisson dance for these values.

yes we should.
Look in that file for `Expression::BinaryExpression`  and when op is '=', we should not use quote!(==) as we do now, but the right thing

ogoffart:
> clippy::erasing_op: This is about multiplying indices with 0 in some
places, which seems to be ok for generated code.

Yes, that's ok for now.  Long term we will want to do this kind of optimization also in our compiler because we can possibly remove more expensive call that the compiler can't remover

PR #389, opened 2021-08-04
Pr/janitor

hunger:
More janitor commits, most are pretty small and simple.

`Remove unnecessary clone() calls` is the only not automatic change in the batch: It removes some code that seems unnecessary after earlier janitorial changes. The clone() is marked as useless, removing that makes the variable unnecessary.

With this applied and #388 applied and ignoring `clippy::vtable_address_comparisons` I got down to 97 warnings (72 of them being unique) in all of sixtyfps. The remaining ones are a bit harder to fix than this set.

### DIFF
--- a/api/sixtyfps-rs/sixtyfps-build/lib.rs
+++ b/api/sixtyfps-rs/sixtyfps-build/lib.rs
@@ -233,7 +233,7 @@ pub fn compile_with_config(
         .join(
             path.file_stem()
                 .map(Path::new)
-                .unwrap_or(Path::new("sixtyfps_out"))
+                .unwrap_or_else(|| Path::new("sixtyfps_out"))
                 .with_extension("rs"),
         );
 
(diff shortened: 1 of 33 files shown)


tools/compiler/main.rs
```diff
@@ -80,7 +80,7 @@ fn main() -> std::io::Result<()> {
             write!(f, " {}", resource)?;
         }
 
-        writeln!(f, "")?;
+        writeln!(f)?;
```
ogoffart, line 83: I did not even know this was possible

api/sixtyfps-rs/sixtyfps-build/lib.rs
```diff
@@ -233,7 +233,7 @@ pub fn compile_with_config(
         .join(
             path.file_stem()
                 .map(Path::new)
-                .unwrap_or(Path::new("sixtyfps_out"))
+                .unwrap_or_else(|| Path::new("sixtyfps_out"))
```
ogoffart, line 236: I thought path::new was a no-op.

ogoffart (approved):
Thanks!

PR #391, opened 2021-08-05
Pr/janitor

hunger:
These are the last instances of `map(|| ())` I can find.

### DIFF
--- a/sixtyfps_compiler/passes/visible.rs
+++ b/sixtyfps_compiler/passes/visible.rs
@@ -24,9 +24,11 @@ pub fn handle_visible(component: &Rc<Component>, type_register: &TypeRegister) {
         component,
         &(),
         &mut |elem: &ElementRc, _| {
-            if elem.borrow().native_class().map_or(false, |n| {
-                Rc::ptr_eq(&n, &native_clip) && elem.borrow().id.ends_with("_visibility")
-            }) {
+            let is_lowered_from_visible_property =
+                elem.borrow().native_class().map_or(false, |n| {
+                    Rc::ptr_eq(&n, &native_clip) && elem.borrow().id.ends_with("_visibility")
+                });
+            if is_lowered_from_visible_property {
                 // This is the element we just created. Skip it.
                 return;
             }
--- a/sixtyfps_runtime/corelib/flickable.rs
+++ b/sixtyfps_runtime/corelib/flickable.rs
@@ -75,13 +75,13 @@ impl FlickableData {
                 }
             }
             MouseEvent::MouseMoved { pos } => {
-                if inner.capture_events
+                let do_intercept = inner.capture_events
                     || inner.pressed_time.map_or(false, |pressed_time| {
                         crate::animations::current_tick() - pressed_time < DURATION_THRESHOLD
                             && (pos - inner.pressed_pos).square_length()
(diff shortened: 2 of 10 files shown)


sixtyfps_compiler/passes/visible.rs
```diff
@@ -24,9 +24,10 @@ pub fn handle_visible(component: &Rc<Component>, type_register: &TypeRegister) {
         component,
         &(),
         &mut |elem: &ElementRc, _| {
-            if elem.borrow().native_class().map_or(false, |n| {
+            let has_native_class = elem.borrow().native_class().map_or(false, |n| {
```
ogoffart, line 27: that's not a right name for it.

Maybe `is_visibility_clip_element`  or `is_lowered_from_visible_property` ?

sixtyfps_runtime/corelib/flickable.rs
```diff
@@ -75,13 +75,12 @@ impl FlickableData {
...
             }
             MouseEvent::MouseMoved { pos } => {
-                if inner.capture_events
-                    || inner.pressed_time.map_or(false, |pressed_time| {
-                        crate::animations::current_tick() - pressed_time < DURATION_THRESHOLD
-                            && (pos - inner.pressed_pos).square_length()
-                                > DISTANCE_THRESHOLD * DISTANCE_THRESHOLD
-                    })
-                {
+                let outside_of_thresholds = inner.pressed_time.map_or(false, |pressed_time| {
```
ogoffart, line 78: We don't want to call current_tick() if we don't have too. So we should keep the capture_events condition before.

sixtyfps_runtime/corelib/model.rs
```diff
@@ -538,7 +538,8 @@ impl<C: RepeatedComponent + 'static> Repeater<C> {
         let listview_geometry_tracker =
             unsafe { self.map_unchecked(|s| &s.listview_geometry_tracker) };
         let init = &init;
-        if listview_geometry_tracker
+
+        let has_no_geometry = listview_geometry_tracker
```
ogoffart, line 542: again wrong variable name.
Maybe `geometry_changed` ?

sixtyfps_runtime/corelib/properties.rs
```diff
@@ -405,12 +405,14 @@ impl PropertyHandle {
 
     /// Implementation of Self::set_binding.
     fn set_binding_impl(&self, binding: *mut BindingHolder) {
-        if self.access(|b| {
+        let has_no_previous_binding = self.access(|b| {
```
ogoffart, line 408: `previous_binding_intercepted`

sixtyfps_runtime/corelib/properties.rs
```diff
@@ -608,21 +610,24 @@ impl<T: Clone> Property<T> {
     where
         T: PartialEq,
     {
-        if !self.handle.access(|b| {
+        let has_no_previous_binding = self.handle.access(|b| {
```
ogoffart, line 613: `previous_binding_intercepted`

ogoffart (commented):
naming is hard. But it shows that the code can be more readable once named because the name you guessed where sometimes the opposite of what it actually is

ogoffart (changes requested):
(see inline comments)

ogoffart (approved):

PR #392, opened 2021-08-05
Allow creating `sixtyfps::Image` from image buffers, in Rust

tronical:


tronical:
Note to self: avoid the entire QImage allocation for image_size() in the Qt backend!

.github/workflows/ci.yaml
```diff
@@ -605,8 +605,9 @@ jobs:
       with:
           command: test
           # skip example_printerdemo_old because that was not in 0.0.5
+          # skip plitters because that was not in 0.1.0
```
ogoffart, line 608:
```suggestion
          # skip plotters because that was not in 0.1.0
```

ogoffart (commented):
Somehow, this needs to be better documented.  The `sixtyfps::Image` documentation should have some examples.   
`imgref::Img<SharedVector<Pixel>>`   should probably re-exported by the sixtyfps crate as well so user code don't have to add imgref in their own Cargo.toml

Also this deserve a changelog entry.

tronical:
> Somehow, this needs to be better documented. The `sixtyfps::Image` documentation should have some examples.
> `imgref::Img<SharedVector<Pixel>>` should probably re-exported by the sixtyfps crate as well so user code don't have to add imgref in their own Cargo.toml

Very good points. The docs especially need indeed examples. I think showing how to interoperate with the image crate is quite essential. I'll make a little image filter example and then derive example code for the docs from that.

> Also this deserve a changelog entry.

Fully agreed.

tronical:
This is going to need a bit more work, but I'm getting there. Marking as WIP.

sixtyfps_runtime/corelib/graphics/image.rs
```diff
@@ -11,6 +11,226 @@ LICENSE END */
...
+    pub fn height(&self) -> usize {
+        match self {
+            Self::RGB8(buffer) => buffer.height(),
+            Self::RGBA8(buffer) => buffer.height(),
+            Self::RGBA8Premultiplied(buffer) => buffer.height(),
+        }
+    }
+}
+
+impl PartialEq for SharedImageBuffer {
```
ogoffart, line 218: Why Is the built-in derive not working?

examples/plotter/Cargo.toml
```diff
@@ -0,0 +1,35 @@
...
+[[bin]]
+path = "main.rs"
+name = "plotter"
+
+[dependencies]
+sixtyfps = { path = "../../api/sixtyfps-rs" }
+#plotters = { version = "0.3.1", default-features = false, features = ["bitmap"] }
+plotters = { version = "0.3.1", default-features = false, features = ["bitmap_backend", "surface_series"] }
+imgref = "1"
+rgb = "0.8.27"
```
ogoffart, line 18: Would be nice if our users didn't have to import that. Could we not re-export it

ogoffart (commented):
Why is there two type: a Image and a SharedImageBuffer.  Can't we merge these in the public API?

tronical:
> Why is there two type: a Image and a SharedImageBuffer. Can't we merge these in the public API?

That’s very tempting. How would this look like?

Image currently wraps ImageInner, the enum that holds either Path, embedded encoded image or embedded decoded image.

Folded, that would mean all those enum variants become public, there would be Image::Path, Image::EncodedImageData and then the variants for the decoded images.

Is that what you had in mind?

sixtyfps_runtime/corelib/graphics/image.rs
tronical: Initially that type contained an `ImgVec`, which didn't implement `PartialEq`. Now we *could* use `PartialEq` from the built-in derive, but it would mean `SharedVector`'s `PartialEq`, which will compare pixel by pixel.

For a `property <image>` that's rather expensive, hence this implementation that is guaranteed to return false if the images differ, but it won't return true for all equal images.

examples/plotter/Cargo.toml
tronical: We could do that, yes. It's only needed for https://docs.rs/rgb/0.8.27/rgb/trait.ComponentBytes.html#method.as_bytes right now.

What's the best way of doing that?

Another option that doesn't require a re-export:

Currently the use is to get a (mutable) reference to a `[u8]` slice via these two methods. So it looks like this:

```rust
pixel_buffer
    .as_slice() // we provide that, to give access to the SharedVector as slice
    .as_bytes(); // provided by rgb's impl of ComponentBytes for the [PixelType]
```

So we could simply offer `as_bytes()` and `as_bytes_mut()` in the API and that's it:

```rust
pixel_buffer
    .as_bytes();
```

ogoffart:
The idea would be to keep the enum internal, but have method to get slices into the buffer.

tronical:
Okay, yes, that's doable. So API wise that's favouring `clone_from_slice_rgb8/clone_from_slice_rgba8/clone_from_slice_rgba8_premultiplied` over a public enum. I'll make that change and we can see how it looks like.

examples/plotter/Cargo.toml
tronical: I'll go ahead and also do that as a commit on top.

sixtyfps_runtime/corelib/graphics/image.rs
```diff
@@ -54,39 +270,80 @@ impl Image {
     pub fn load_from_path(path: &std::path::Path) -> Result<Self, LoadImageError> {
         Ok(Image(ImageInner::AbsoluteFilePath(path.to_str().ok_or(LoadImageError(()))?.into())))
     }
-    /*
-    /// Create an image from argb pixels.
+
+    /// Creates a new Image from the specified shared pixel buffer, where each pixel has three color
+    /// channels (red, green and blue) encoded as u8.
+    pub fn new_rgb8(buffer: SharedPixelBuffer<RGB8Pixel>) -> Self {
```
ogoffart, line 276: should this function and other maybe be called  `from_rgb8`   ?

tronical:
This is pretty minimal now, the only public dependency we have now is the rgb crate, where we re-export the `RGB<u8>` and `RGBA<u8>` pixel types and use the `ComponentBytes`, `Pod` and `AsPixels` traits in the public `SharedPixelBuffer` API. The latter is used for casting the pixel vector data to an array of raw bytes, for interop when creating buffers and asking somebody else to feed data into it.

For feeding image data into `sixtyfps::Image` there's `clone_from_slice`:

```
    pub fn clone_from_slice<SourcePixelType>(
        pixel_slice: &[SourcePixelType],
        width: usize,
        height: usize,
    ) -> Self
    where
        [SourcePixelType]: rgb::AsPixels<Pixel>,
```

sixtyfps_runtime/corelib/graphics/image.rs
```diff
@@ -44,7 +215,52 @@ impl<'a> From<&'a Image> for &'a ImageInner {
 #[derive(Default, Debug, PartialEq)]
 pub struct LoadImageError(());
 
-/// An image type that can be displayed by the Image element
+/// An image type that can be displayed by the Image element. You can construct
+/// Image objects from a path to an image file on disk, using [`Self::load_from_path`].
+///
+/// Another typical use-case is to have an external crate for rendering some content as an image.
```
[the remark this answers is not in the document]
tronical: Okay

sixtyfps_runtime/corelib/graphics/image.rs
tronical: Okay, I'll rename them. It makes sense, as without the `Premultiplied` distinction we'd use the `From` trait I think.

.github/workflows/ci.yaml
```diff
@@ -589,6 +589,13 @@ jobs:
           git checkout v${{ matrix.from_version }} --no-overlay -- tests/cases
           # Remove printerdemo_old from Cargo.toml because this directory was not there in 0.0.5
           sed -i "/printerdemo_old/d" Cargo.toml
+    - name: Remove newer examples
```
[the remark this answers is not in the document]
tronical: I've included the image filter in ci.yaml. The plotter needs more work as it uses the system fonts to draw text on the axes.

examples/imagefilter/Cargo.toml
```diff
@@ -0,0 +1,32 @@
...
+
+#wasm# [lib]
+#wasm# path = "main.rs"
+#wasm# crate-type = ["cdylib"]
+
+#wasm# [target.'cfg(target_arch = "wasm32")'.dependencies]
+#wasm# wasm-bindgen = { version = "0.2" }
+#wasm# web-sys = { version = "0.3", features=["console"] }
+#wasm# console_error_panic_hook = "0.1.5"
+#wasm# getrandom = { version = "0.2.2", features = ["js"] }
```
ogoffart, line 32: do we need getrandom or is that just copy/paste?

ogoffart (commented):
> The plotter needs more work as it uses the system fonts to draw text on the axes.

Does it, it looks like plotter should be well supported on wasm

examples/imagefilter/Cargo.toml
tronical: Well spotted, that's copy & pasted. I had removed it locally many times but never committed because I would have accidentally committed without the `#` :).

ogoffart (approved):

api/sixtyfps-rs/lib.rs
```diff
@@ -185,7 +185,9 @@ struct MyStruct {
 
 pub use sixtyfps_macros::sixtyfps;
 
-pub use sixtyfps_corelib::graphics::{Brush, Color, Image, LoadImageError, RgbaColor};
+pub use sixtyfps_corelib::graphics::{
+    Brush, Color, Image, LoadImageError, RGB8Pixel, RGBA8Pixel, RgbaColor, SharedPixelBuffer,
```
ogoffart, line 189: Now we have both RGBA8Pixel and RgbaColor in the public API

Note the difference in the uppercasing of rgba.
Also do we still need RgbaColor?

sixtyfps_runtime/corelib/graphics/image.rs
```diff
@@ -11,6 +11,175 @@ LICENSE END */
 use crate::slice::Slice;
 use crate::{SharedString, SharedVector};
 
+/// SharedPixelBuffer is a container for storing image data as pixels, backed by
+/// [`SharedVector`]. That means it is internally reference counted and cheap
+/// to clone.
+///
+/// You can construct new a new empty shared pixel buffer with [`SharedPixelBuffer::new`],
```
ogoffart, line 18:
```suggestion
/// You can construct new a new shared pixel buffer with [`SharedPixelBuffer::new`],
```

Empty would mean the image is empty and that's not so useful. I wanted to replace empty by transparent but that's not right either

tronical:
> > The plotter needs more work as it uses the system fonts to draw text on the axes.
> 
> Does it, it looks like plotter should be well supported on wasm

Apologies, I should have elaborated more on what I mean with "more work".

It is very well supported on wasm, but when run against a HTML canvas element: See https://docs.rs/plotters-canvas/0.3.0/plotters_canvas/struct.CanvasBackend.html#method.with_canvas_object .

In order for us to use that, we'd have to add support for creating a `sixtyfps::Image` from a `HtmlImageElement`, which is doable. And then the plotters example would have to convert the `HtmlCanvasElement` to a `HtmlImageElement`.

Alternatively, we could patch plotters to work with a supplied font. Or we could implement a plotters rendering backend. Or use a graph that doesn't use text and hope that we don't hit the corresponding code path, or remove the axis description.

sixtyfps_runtime/corelib/graphics/image.rs
```diff
@@ -11,6 +11,175 @@ LICENSE END */
...
+impl<Pixel> SharedPixelBuffer<Pixel> {
+    /// Return a slice to the pixel data.
+    pub fn as_slice(&self) -> &[Pixel] {
+        self.data.as_slice()
+    }
+}
+
+impl<Pixel: Clone + Default> SharedPixelBuffer<Pixel> {
+    /// Creates a new SharedPixelBuffer with the given width and height. Each pixel will be initialized with the value
+    /// that `[Default::default()`] returns for the Pixel type.
```
ogoffart, line 83:
```suggestion
    /// that [`Default::default()`] returns for the Pixel type.
```

sixtyfps_runtime/corelib/graphics/image.rs
```diff
@@ -11,6 +11,175 @@ LICENSE END */
 use crate::slice::Slice;
 use crate::{SharedString, SharedVector};
 
+/// SharedPixelBuffer is a container for storing image data as pixels, backed by
+/// [`SharedVector`]. That means it is internally reference counted and cheap
```
ogoffart, line 15: But we don't expose the inner vector at any point, right? So that's an implementation detail

sixtyfps_runtime/corelib/graphics/image.rs
```diff
@@ -11,6 +11,175 @@ LICENSE END */
...
+
+    /// Returns the number of pixels per line.
+    pub fn stride(&self) -> usize {
+        self.stride
+    }
+}
+
+impl<Pixel: Clone> SharedPixelBuffer<Pixel> {
+    /// Return a mutable slice to the pixel data. If the SharedPixelBuffer was shared, this will make a copy of the buffer.
+    pub fn as_mut_slice(&mut self) -> &mut [Pixel] {
```
ogoffart, line 52: Wondering if make_mut_slice would not be a better name. What do you think?

sixtyfps_runtime/corelib/graphics/image.rs
```diff
@@ -54,39 +270,80 @@ impl Image {
...
-        width: u32,
-        height: u32,
-        data: SharedVector<u32>,
-    ) -> Result<Self, LoadImageError> {
-        if data.len() != width as usize * height as usize {
-            Err(LoadImageError(()))
-        } else {
-            Ok(Image(ImageReference::EmbeddedRgbaImage { width, height, data }))
-        }
+    /// Only construct an Image with this function if you know that your pixels are encoded this way.
```
ogoffart, line 290: What if it isn't?  Is it UB? Then the function should be unsafe. Or is it just that were gonna get rendering artifacts? Or should we check it and panic if it isn't?

sixtyfps_runtime/corelib/graphics/image.rs
```diff
@@ -54,39 +270,80 @@ impl Image {
...
 pub(crate) mod ffi {
     #![allow(unsafe_code)]
 
     use super::super::Size;
     use super::*;
 
+    /// Expand RGB8Pixel so that cbindgen can see it. (is in fact rgb::RGB<u8>)
+    #[cfg(cbindgen)]
+    #[repr(C)]
+    struct RGB8Pixel {
```
ogoffart, line 332: Alternatively, one could declare it in C++ directly in one of our internal headers

ogoffart (approved):

sixtyfps_runtime/corelib/graphics/image.rs
tronical: I think it's a good idea to mention that the image will be empty - it visually is. Otherwise one might wonder if the pixel data is undefined/uninitialized.

Or does "empty" for you mean zero width and height (despite `new` taking those parameters)?

api/sixtyfps-rs/lib.rs
tronical: I think we should keep the `Pixel` types, in contrast to the `Color` types.

The casing is a good point though, that is an inconsistency. I'll change it to `Rgb8Pixel` and `Rgba8Pixel`.

sixtyfps_runtime/corelib/graphics/image.rs
tronical: I think it's good to mention that to make it easier to understand why it's cheap to clone. But OTOH it's not strictly needed - I'll remove it. We can add it in the unlikely event that somebody asks :)

sixtyfps_runtime/corelib/graphics/image.rs
tronical: As discussed, yes ver good point, that's a better name. I'll make the corresponding change to this and SharedVector after merging.

sixtyfps_runtime/corelib/graphics/image.rs
tronical: There is no way of programmatically checking if the alpha is pre-multiplied or not. If you use this function with an RGBA image that has the alpha not pre-multiplied, then you'll get a visually differently rendered image (it'll have a kind of tint) - and vice versa: If you get an Rgba-premultiplied image and use it with `from_rgba8` you'll see that the colors are off.

sixtyfps_runtime/corelib/graphics/image.rs
tronical: True, and that might become the way to go if we make this image-from-data API public in C++.

PR #393, opened 2021-08-06
Pr/janitor

hunger:
More janitor stuff... but we are getting close to the finish line: We are down to 24 warnings with this applied (and ignoring the vtable pointer comparison error).

### DIFF
--- a/sixtyfps_compiler/generator/cpp.rs
+++ b/sixtyfps_compiler/generator/cpp.rs
@@ -2128,7 +2128,7 @@ fn compile_path(path: &crate::expression_tree::Path, component: &Rc<Component>) 
     }
 }
 
-fn compile_path_events(events: &crate::expression_tree::PathEvents) -> (Vec<String>, Vec<String>) {
+fn compile_path_events(events: &[crate::expression_tree::PathEvent]) -> (Vec<String>, Vec<String>) {
     use lyon_path::Event;
 
     let mut coordinates = Vec::new();
--- a/sixtyfps_compiler/generator/rust.rs
+++ b/sixtyfps_compiler/generator/rust.rs
@@ -1891,7 +1891,7 @@ fn get_layout_info(
     }
 }
 
-fn compile_path_events(events: &crate::expression_tree::PathEvents) -> TokenStream {
+fn compile_path_events(events: &[crate::expression_tree::PathEvent]) -> TokenStream {
     use lyon_path::Event;
 
     let mut coordinates = Vec::new();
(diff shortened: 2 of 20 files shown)


sixtyfps_compiler/object_tree.rs
```diff
@@ -396,7 +396,7 @@ impl<'a> IntoIterator for &'a BindingsMap {
     type IntoIter = std::collections::btree_map::Iter<'a, String, BindingExpression>;
 
     fn into_iter(self) -> Self::IntoIter {
-        (&self.0).into_iter()
+        (&self.0).iter()
```
ogoffart, line 399: not sure what was wrong with that?
then maybe we don't need to take a reference anymore?

sixtyfps_compiler/generator/cpp.rs
```diff
@@ -2128,7 +2128,7 @@ fn compile_path(path: &crate::expression_tree::Path, component: &Rc<Component>)
     }
 }
 
-fn compile_path_events(events: &crate::expression_tree::PathEvents) -> (Vec<String>, Vec<String>) {
+fn compile_path_events(events: &[crate::expression_tree::PathEvent]) -> (Vec<String>, Vec<String>) {
```
ogoffart, line 2131: How is that better?

sixtyfps_compiler/generator/rust.rs
```diff
@@ -126,6 +126,12 @@ pub fn generate(doc: &Document, diag: &mut BuildDiagnostics) -> Option<TokenStre
         #[allow(non_snake_case)]
         #[allow(clippy::style)]
         #[allow(clippy::complexity)]
+        #[allow(clippy::approx_constant)]
+        #[allow(clippy::float_cmp)]
+        #[allow(clippy::if_same_then_else)]
+        #[allow(clippy::erasing_op)]
+        #[allow(clippy::redundant_clone)]
+        #[allow(clippy::cmp_owned)]
```
ogoffart, line 134: Was than meat in this commit?

sixtyfps_compiler/generator/cpp.rs
ogoffart: Ah yes, because it was a Vec before

sixtyfps_compiler/generator/rust.rs
hunger: Oh, sorry! that needs to go!

sixtyfps_compiler/object_tree.rs
hunger: This is why this is bad (according to the clippy help):

> Readability. Calling into_iter on a reference will not move out its content into the resulting iterator, which is confusing. It is better just call iter or iter_mut directly.

sixtyfps_compiler/generator/cpp.rs
hunger: This accepts anything that can be turned into a slice, not just a Vec.

It also needs less derefs to get to the actual data, so it might even be a little bit faster.

ogoffart (approved):

ogoffart:
there is merge conflicts :-/

PR #394, opened 2021-08-06
Fix SharedVector creation from iterators with default size hints

tronical:
* In capacity_for_grow, don't compare the number of elements
  (current_cap) with the size of an element in bytes.
  That fixes the capacity not growing.
* When re-allocating due to growth and copying elements from the old
  inner to the new inner, make sure to copy all old elements from the beginning,
  not only the last element repeatedly.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -29,6 +29,7 @@ All notable changes to this project will be documented in this file.
  - Support of `*=` and `/=` on types with unit such as length.
  - Proper compilation error when using a self assignment operator on an invalid type instead of a panic
  - Angle conversion for values specified in radians, gradians and turns
+ - SharedVector was sometimes not allocating big enough storage.
 
 ## [0.1.0] - 2021-06-30
 
--- a/sixtyfps_runtime/corelib/sharedvector.rs
+++ b/sixtyfps_runtime/corelib/sharedvector.rs
@@ -65,7 +65,7 @@ fn alloc_with_capacity<T>(capacity: usize) -> NonNull<SharedVectorInner<T>> {
 /// Return a new capacity suitable for this vector
 /// Loosely based on alloc::raw_vec::RawVec::grow_amortized.
 fn capacity_for_grow(current_cap: usize, required_cap: usize, elem_size: usize) -> usize {
-    if current_cap >= elem_size {
+    if current_cap >= required_cap {
         return current_cap;
     }
     let cap = core::cmp::max(current_cap * 2, required_cap);
@@ -315,7 +315,7 @@ impl<T> FromIterator<T> for SharedVector<T> {
                         while *begin < size {
                             unsafe {
                                 core::ptr::write(
-                                    result.inner.as_mut().data.as_mut_ptr().add(size),
+                                    result.inner.as_mut().data.as_mut_ptr().add(*begin),
                                     core::ptr::read(old_inner.as_ref().data.as_ptr().add(*begin)),
                                 );
                                 *begin += 1;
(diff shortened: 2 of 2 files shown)


sixtyfps_runtime/corelib/sharedvector.rs
```diff
@@ -65,9 +65,6 @@ fn alloc_with_capacity<T>(capacity: usize) -> NonNull<SharedVectorInner<T>> {
 /// Return a new capacity suitable for this vector
 /// Loosely based on alloc::raw_vec::RawVec::grow_amortized.
 fn capacity_for_grow(current_cap: usize, required_cap: usize, elem_size: usize) -> usize {
-    if current_cap >= elem_size {
```
ogoffart, line 68: maybe this was meant to be required_cap.

So we do not re-allocates when it it already big enough.

But I guess the caller of capacity_for_grow already takes care of that?

sixtyfps_runtime/corelib/sharedvector.rs
```diff
@@ -483,6 +480,36 @@ fn invalid_capacity_test() {
...
 
+#[test]
+fn collect_from_iter_with_no_size_hint() {
+    struct NoSizeHintIter {
+        data: Vec<i32>,
+        i: usize,
+    }
+
+    impl Iterator for NoSizeHintIter {
+        type Item = i32;
```
ogoffart, line 491: For test, I like using more complex types like String or something else that has a drop so that double drop or wrong initialization would be more visible.

CHANGELOG.md
```diff
@@ -29,6 +29,7 @@ All notable changes to this project will be documented in this file.
  - Support of `*=` and `/=` on types with unit such as length.
  - Proper compilation error when using a self assignment operator on an invalid type instead of a panic
  - Angle conversion for values specified in radians, gradians and turns
+ - SharedVector now works when collecting from an iterator with a default (zero) size hint.
```
ogoffart, line 32:
```suggestion
 - SharedVector now was sometimes not allocating big enough storage.
```

A zero size iterator was just one of the way to reproduce this bug.

ogoffart (approved):

sixtyfps_runtime/corelib/sharedvector.rs
tronical: You're right, that makes more sense and the call site `push()` doesn't check, so that would have been a regression!

sixtyfps_runtime/corelib/sharedvector.rs
tronical: I'll cover the capacity issue in a separate test function (via `push`), but I'd like to retain this test function to cover the issue in `from_iter` specifically, when we start out with a capacity that's smaller than the final result. That's why I used this wrapper, to trigger the code path where we copy elements specifically. Is there another type from the standard library that would be better to use instead that does *not* start with the exact capacity?

I was thinking of an iterator with `filter()` or so applied, but I'm not sure that's worth it - I'd find this easier to read. What do you think?

ogoffart: I meant using String instead of i32 as the Item, because i32 is Copy and don't need to be drop'ed.

tronical: Ahh, got it - thanks! Will change.

ogoffart (approved):

tronical:
I'll merge this as I need it in another branch, but let me know if you have any other suggestions for changes :)

PR #396, opened 2021-08-06
Janitor: Replace `'` with `´` where that makes sense

hunger:
Replace strings to use `´` for elisions instead of `'`. Sorry, this
is a pet peeve of mine;-)

Remove many of the elisions altogether by spelling them out. That looks
IMHO more professional in the generated documentation (where many of
these strings might show up).

On the plus side this makes cspell a bit happier, since that recognizes
that the part after the `´` can be ignored, which it does not
consistently do for the part after `'`. This allows to remove one extra
entry in the word list.


hunger:
Not sure merging this makes too much sense:-)

tronical:
I'm curious, where does the use of `´` instead of `'` come from? Is the `´` meant to be visually closer to the typesetter's apostrophe than the typewriter apostrophe we are currently using?

I'm inclined to say that in software development the typewriter's variant dominates, and it is also clearly the easiest to reach in keyboard layouts. So if we decide to go for this, everyone will have to change their habits and we might need a way to "enforce" this, otherwise it'll be very quickly a mixture (as opposed to now).

I wonder if there's a way to teach CSpell to accept the typewriter's variant?

ogoffart:
Yes, If anything, it should use the apostrophe `’` , not the quote `´`

hunger:
You both seem to be comfortable with "you're", "let's" and the rest in the docs, so I am not sure this patch is worth the interruption it introduces. 

I had also expected more impact on cspell: This entire patch gets rid of only a single word in the wordlist, so for that alone it is definitely not worth doing either.

So how about I extract the bigger changes and submit only those?

tronical:
I remain curious about the motivation behind this change. Is it typesetter vs. typewriter? What’s the origin of favoring one over the other?

hunger:
Typesetting making a distinction between the different uses of the character was the original motivation, coupled with the idea that it will fix a couple of cspell issues.

But now that @ogoffart  pointed out that I have been obsession over the wrong character, I feel thoroughly stupid and should probably just retract the patch entirely -- it does fix some few cspell problems, but that is probably not worthwhile having such a big patch.

PR #398, opened 2021-08-09
Error when trying to use internal builtin items

ogoffart:


sixtyfps_compiler/builtins.60
```diff
@@ -73,7 +73,6 @@ export Rotate := _ {
     property <length> width;
     property <length> height;
     //-default_size_binding:expands_to_parent_geometry
-    //-is_internal
```
tronical, line 76: I'd prefer if this remained internal, as it is also not documented and we're not using it in any of our examples.

I recall adding this for the attempt of doing the coffee machine demo, but so far only a test is using it. Perhaps we can allow the use of internal elements in tests?

tronical (approved):

sixtyfps_compiler/builtins.60
ogoffart: > Perhaps we can allow the use of internal elements in tests?

This would be hard as we would somehow pass the information that it is a test from each driver to the document loader.
Keeping it internal is basically removing the feature anyway, so maybe we can just remove the rotation test
(renaming it to `rotate.60.disabled`  or something)

tronical: You're right, there's no point in keeping the test around then either. Especially since it's effectively a "manual test" anyway.

tronical (approved):
Thanks :)

PR #401, opened 2021-08-10
Fix links to rgb crate in the sixtyfps.io hosted online docs

tronical:
Use the external docs rustdoc feature to link to the rgb docs on docs.rs.

### DIFF
--- a/sixtyfps_runtime/corelib/Cargo.toml
+++ b/sixtyfps_runtime/corelib/Cargo.toml
@@ -41,6 +41,7 @@ weak-table =  "0.3"
 scopeguard = "1.1.0"
 cfg-if = "1"
 unicode-segmentation = "1.8.0"
+# Note: When bumping this, make sure to adjust the rgb base url in ci.yaml for the nightly docs build
 rgb = "0.8.27"
 
 [target.'cfg(target_arch = "wasm32")'.dependencies]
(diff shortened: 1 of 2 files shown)


sixtyfps_runtime/corelib/Cargo.toml
```diff
@@ -41,6 +41,7 @@ weak-table =  "0.3"
 scopeguard = "1.1.0"
 cfg-if = "1"
 unicode-segmentation = "1.8.0"
+# Note: When bumping this, make sure to adjust the rgb base url in ci.yaml for the nightly docs build
 rgb = "0.8.27"
```
ogoffart, line 45: Could the CI extract this number with a simple shell script:
```
export RUSTDOCFLAGS="--extern-html-root-url rgb=https://docs.rs/rgb/`grep 'rgb = '  sixtyfps_runtime/corelib/Cargo.toml | sed 's/^.*"\(.*\)"/\1/' `/"
```

Or not worth it?

ogoffart (approved):
That's great if it works
I guess it links to all dependent crate already on docs.rs

sixtyfps_runtime/corelib/Cargo.toml
tronical: Nice idea! I'll try that (after merging)

PR #403, opened 2021-08-11
sharedvector clear

tronical:


sixtyfps_runtime/corelib/sharedvector.rs
```diff
@@ -251,6 +245,32 @@ impl<T: Clone> SharedVector<T> {
...
+            // Safety: The array was of size inner.header.size, so there should be an element there
+            unsafe {
+                core::ptr::drop_in_place(inner.data.as_mut_ptr().add(inner.header.size));
+            }
+        }
+    }
+
+    /// Clears the vector and removes all elements. The capacity remains unaffected.
+    pub fn clear(&mut self) {
+        self.detach(self.capacity());
```
ogoffart, line 271: But detach will do a deep clone of everything inside, we don't want that when we want to clear. That's quite unintuitive.

sixtyfps_runtime/corelib/sharedvector.rs
```diff
@@ -251,6 +245,32 @@ impl<T: Clone> SharedVector<T> {
...
+
+    fn shrink(&mut self, new_len: usize) {
+        if self.len() == new_len {
+            return;
+        }
+
+        debug_assert!(
+            unsafe { self.inner.as_ref().header.refcount.load(atomic::Ordering::Relaxed) } == 1
+        );
+        // Safety: caller (and above debug_assert) must ensure that the array is not shared.
```
ogoffart, line 257: that should be an `assert`  or the function need to be marked `unsafe`

In fact, if it was public, i'd say that shrink should in fact detach if it is shared, but only clone new_len element to the final vector (and only allocates (new_len)

ogoffart (changes requested):
What's really the use case of going to great length to keep the capacity. In fact, you most likely want to clear just to say: "i want the vector to be empty now", so equivelent to `vec = SharedVector::default()`.
For comparison, that's what the QString::clear does (does not keep allocated memory).  QList::clear does what you do in Qt6, but it didn't do it always in Qt5.

Rust's `Vec::clear` or C++'s `std::vector::clear`  keeps the capacity, but that's easier to do in a non-shared vector.
I think i'd see a more usefull behavior as
 - keep the capacity if not shared
 - set to empty vector without capacity otherwise.

tronical:
> I think i'd see a more usefull behavior as
> 
> * keep the capacity if not shared
> * set to empty vector without capacity otherwise.

Sounds good to me.

api/sixtyfps-cpp/include/sixtyfps_sharedvector.h
```diff
@@ -127,6 +127,20 @@ struct SharedVector
...
+    /// Clears the vector and removes all elements. The capacity remains unaffected.
+    void clear()
+    {
+        if (inner->refcount != 1) {
+            *this = SharedVector();
+        } else {
+            auto b = begin(), e = end();
+            inner->size = 0;
+            for (auto it = b; it < e; ++it) {
+                it->~T();
```
ogoffart, line 139: (there is a leak when the destructor throws an exception, but let's ignore that :-)

sixtyfps_runtime/corelib/sharedvector.rs
```diff
@@ -251,6 +245,42 @@ impl<T: Clone> SharedVector<T> {
...
+    pub fn clear(&mut self) {
+        let is_shared =
+            unsafe { self.inner.as_ref().header.refcount.load(atomic::Ordering::Relaxed) } != 1;
+        if is_shared {
+            unsafe {
+                if self.inner.as_ref().header.refcount.fetch_sub(1, atomic::Ordering::SeqCst) == 1 {
+                    drop_inner(self.inner)
+                }
+            }
+            self.inner = NonNull::from(&SHARED_NULL).cast();
```
ogoffart, line 279:
```suggestion
            *self = SharedVector::default();
```

ogoffart (approved):

PR #404, opened 2021-08-11
Pr/pre commit

hunger:
Another proposal for more tools:-)

The first commit has the configuration and  instructions on how to set this up. The second one has the result of applying all hooks on all files in the repo.

Most changes are about missing newlines at the end of files and there is a chunk for applying the `black` formater on the python file. The rest is basically some stripped whitespace at the end of lines.


tronical:
+1 from me. This is good for people who would like to use this opt-in hygiene and the changes to the files look fine to me.

(This is intended as opt-in, right?)

hunger:
You can not force clients to run git hooks, so this has to be opt-in.

It might make sense to run some of these on the server side as well though. E.g. this includes a check for filenames that cause trouble on case-insensitive filesystems.

hunger:
@tronical: Is there a way to not have your license checker complain about the config file for pre-commit? I see no need to add a license header into that, it is in its entirety copied from the example configs on the relevant test repositories that are used.

hunger:
Sorry, I totally forgot about this one: All should be green now that @tronical's change is in that prevents the copyright program to go red...

hunger:
Added some short section about pre-commit to the CONTRIBUTING.md file while at it:-)

tronical (approved):

PR #405, opened 2021-08-12
Add a simple loading spinner animation for the plotter wasm demo

tronical:
I'd like to apply this to all the wasm examples, but this one is here first for review.

The demos take a while to download, especially over slower mobile connections.

https://user-images.githubusercontent.com/1486/129153614-f9d49d0a-77cd-40af-9689-d0dbf7332693.mov

### DIFF
--- a/examples/plotter/index.html
+++ b/examples/plotter/index.html
@@ -8,10 +8,37 @@
 
 <head>
     <meta content="text/html;charset=utf-8" http-equiv="Content-Type" />
+    <style>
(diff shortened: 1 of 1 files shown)


ogoffart:
Why only the plotter and not all demos?

tronical:
Yeah, it's a bit unfortunate - this was about the smallest I could find on the web after tweaking it a little and vs. I've shrunk it a little by disabling the formatter. Will merge and apply to the other examples.

PR #413, opened 2021-08-17
vscode extension: add an option to set LSP parameter

ogoffart:


ogoffart:
Currently just one settings to set all the arguments.

Or should we insteads have several settings to set the style, the backend, the include path, ...  separately.

In the config, it looks like this:

Screenshot_20210817_161451

tronical (approved):

PR #415, opened 2021-08-17
Interpreter API: preserve the dashes and underscore when listing properties

ogoffart:


tronical (approved):
Nice! That's fairly minimal.

PR #421, opened 2021-08-19
add missing opening html tag in the rust tutorial

marcograss:
in the section to run the tutorial in the browser with wasm, the <html> opening tag is missing for the stub webpage


tronical:
Ooops, thank you! :)

PR #424, opened 2021-08-20
Fix panic when calling focus() on a repeated element

### LINKED ISSUE #422: Panic when calling `focus()` on conditional element
Noticed this when upgrading cargo-ui, the following test-case causes a panic due to out-of-bounds item-tree access:

```
export Test := Window {
    if (true): combo := FocusScope {
        TouchArea {
            clicked => {
                combo.focus();
            }
        }
    }
}
```

This is triggered in cargo-ui through the use of the ComboBox in the package selection, that itself is a conditional focus scope.

tronical:
When trying to reference an instance of a repeated item, for use with
ItemRc or VRef<Item>, the item_index (or id) of the ElementRc is not
directly what we want.

For the Rust and C++ code generators, the self/_self already refers
correctly to the repeated instance, but item_index must be zero to refer
to the root.

For the interpreter the self equivalent is the
local_context.component_instance is already the repeated instance that
we can use, and we're just looking for the root element of that, in
order to create the ItemRc/VRef.

Fixes #422

### DIFF
--- a/sixtyfps_compiler/expression_tree.rs
+++ b/sixtyfps_compiler/expression_tree.rs
@@ -776,6 +776,12 @@ impl Expression {
         self.visit(|e| e.visit_recursive(visitor));
     }
 
+    /// Visit itself and each sub expression recursively
+    pub fn visit_recursive_mut(&mut self, visitor: &mut dyn FnMut(&mut Self)) {
+        visitor(self);
+        self.visit_mut(|e| e.visit_recursive_mut(visitor));
+    }
+
     pub fn is_constant(&self) -> bool {
         match self {
             Expression::Invalid => true,
--- a/sixtyfps_compiler/generator/cpp.rs
+++ b/sixtyfps_compiler/generator/cpp.rs
@@ -1413,6 +1413,34 @@ fn access_named_reference(
     access_member(&nr.element(), nr.name(), component, component_cpp)
 }
 
+/// Returns the code that can access the component of the given element
+fn access_element_component<'a>(
+    element: &ElementRc,
+    current_component: &Rc<Component>,
+    component_cpp: &'a str,
+) -> Cow<'a, str> {
+    let e = element.borrow();
+    let enclosing_component = e.enclosing_component.upgrade().unwrap();
+    if Rc::ptr_eq(current_component, &enclosing_component) {
+        component_cpp.into()
+    } else {
+        access_element_component(
+            element,
+            &current_component
+                .parent_element
+                .upgrade()
+                .unwrap()
+                .borrow()
+                .enclosing_component
+                .upgrade()
+                .unwrap(),
+            &format!("{}->parent", component_cpp),
+        )
+        .to_string()
+        .into()
+    }
+}
+
 fn compile_expression(
     expr: &crate::expression_tree::Expression,
     component: &Rc<Component>,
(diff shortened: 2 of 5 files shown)


ogoffart:
Does that work with several level of repeater?

eg, the example slightly modified: 

```
export TestCase := Window {
    width: 100phx;
    height: 100phx;

    property<bool> combo_has_focus;

    if (true): combo := FocusScope {
        if (true): TouchArea {
            clicked => {
                combo.focus();
                root.combo_has_focus = combo.has-focus;
            }
        }
    }
}
```

sixtyfps_compiler/generator/rust.rs
```diff
@@ -1247,7 +1247,13 @@ fn compile_expression(expr: &Expression, component: &Rc<Component>) -> TokenStre
                     if let Expression::ElementReference(focus_item) = &arguments[0] {
                         let focus_item = focus_item.upgrade().unwrap();
                         let focus_item = focus_item.borrow();
-                        let item_index = focus_item.item_index.get().unwrap();
+                        let item_index = if focus_item.repeated.is_some() {
+                            // We want the repeated instance itself; _self already points to it, and the root element
```
ogoffart, line 1251: i'm concern that _self does not always points to it as it could be a deeper instance.

ogoffart:
I'm wondering: what does the focus_item in `Expression::ElementReference(focus_item)` points to, the "inner" or the "outer" ?
I believe this should already points to the "inner" element. If not, i think the bug is there, and we should not be trying to fix it in the codegen, but rather when creating the ElementReference

sixtyfps_compiler/generator/rust.rs
tronical: You're right, `self` is incorrect in this nested case.

tronical:
You're right, the real bug seems to be that `focus_item` is pointing to the outer element. That would probably make it work right away in the interpreter and the C++/Rust code generators still need code to traverse "up" from `self` if needed - similar to `access_member`.

sixtyfps_compiler/generator/cpp.rs
```diff
@@ -1413,6 +1413,34 @@ fn access_named_reference(
...
+                .unwrap()
+                .borrow()
+                .enclosing_component
+                .upgrade()
+                .unwrap(),
+            &format!("{}->parent", component_cpp),
+        )
+        .to_string()
+        .into()
+    }
```
tronical, line 1441: This doesn't *have* to be a recursion, I could also write it as a simple loop. Do you have a preference?

sixtyfps_compiler/passes/repeater_component.rs
```diff
@@ -93,4 +93,18 @@ fn adjust_references(comp: &Rc<Component>) {
             };
         }
     });
+    // Transform any references to the repeated element to refer to the root of each instance.
+    visit_all_expressions(comp, |expr, _| {
```
tronical, line 97: This is quite "aggressive" and I'm not entirely certain if this needs any additional checks ("access only from inside?").

sixtyfps_compiler/expression_tree.rs
```diff
@@ -776,6 +776,12 @@ impl Expression {
         self.visit(|e| e.visit_recursive(visitor));
     }
 
+    /// Visit itself and each sub expression recursively
```
ogoffart, line 779: i really wish there was some way to not have to duplicate everything for mut and non mut

sixtyfps_compiler/generator/cpp.rs
ogoffart: i don't have a preference.

tronical:
I've implemented a different approach now that tries to solve this before the code gen. I have the sinking feeling that I might be missing something though.

sixtyfps_compiler/passes/repeater_component.rs
ogoffart: A few line earlier, we do something similar.  so i guess that's ok.

ogoffart (approved):

tronical:
Thx :)

PR #425, opened 2021-08-20
Refactor the way the two-ways biding are represented internaly

ogoffart:
Don't put them in a fake expression.
This simplifies a bit the expression handling, and will make
possible to fix analysis that needs a vew into the aliases


tronical (approved):
That's a nice simplification!

PR #430, opened 2021-08-23
Fix default text color when using Qt/native style in dark mode on macOS

tronical:
We apply a default window background from the palette to all `Window`
elements, and likewise we need to apply the default text color to Text
elements to ensure a readable contrast.

### DIFF
--- a/sixtyfps_runtime/rendering_backends/qt/qt_widgets.rs
+++ b/sixtyfps_runtime/rendering_backends/qt/qt_widgets.rs
@@ -2269,6 +2269,7 @@ pub struct NativeStyleMetrics {
     pub layout_padding: Property<f32>,
     pub text_cursor_width: Property<f32>,
     pub window_background: Property<Color>,
+    pub default_text_color: Property<Color>,
 }
 
 impl Default for NativeStyleMetrics {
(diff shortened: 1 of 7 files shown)


ogoffart (approved):
Looks good.  :+1:

The commit message is a bit misleading because, while the fix is a good side effect of the change, this is not really what the change do.

(This is not only in dark mode on macOS)

sixtyfps_runtime/rendering_backends/qt/qt_widgets.rs
```diff
@@ -2269,6 +2269,7 @@ pub struct NativeStyleMetrics {
     pub layout_padding: Property<f32>,
     pub text_cursor_width: Property<f32>,
     pub window_background: Property<Color>,
+    pub default_text_color: Property<Brush>,
```
ogoffart, line 2272: I guess this should be a Color,  not a Brush

tronical: You're right, maybe this should be the minimal type here and we do the conversion in the style application. I'll fix that.

tronical:
> Looks good. 👍
> 
> The commit message is a bit misleading because, while the fix is a good side effect of the change, this is not really what the change do.
> 
> (This is not only in dark mode on macOS)

True true, I got a bit too focused on the effect that I see :). I'll mention the macOS bit as an extra sentence and change the wording to also match the change log more closely.

PR #443, opened 2021-08-30
Fix missing compiler warning about duplicated ids when using id more …

tronical:
…than twice

Don't remove the id of a duplicate from the seen set, as it may appear
again.


ogoffart (approved):
When this was an error, i thought this was fine because it would have had to be fixed anyway and then you'd get another error.
But this is better to show a diagnostic for every id

tronical:
Just ran into this as I was reading the code to do similar checks with export duplicates :)

PR #444, opened 2021-08-31
Produce errors when exporting a name multiple times

tronical:


sixtyfps_compiler/object_tree.rs
```diff
@@ -1612,63 +1612,109 @@ impl Exports {
...
-        ));
+            }),
+        );
+
+        struct SeenExport {
+            name_location: SyntaxNode,
+            warned: bool,
+        }
+        let mut seen_exports: HashMap<String, SeenExport> = HashMap::new();
+        let mut export_diagnostics = Vec::new();
```
tronical, line 1682: This is needed because `diag` can't be captured mutably by multiple closures. I'd be happy to re-arrange the code if that's desired. This seemed like an okay compromise to me to have the filter condition in one spot.

ogoffart (approved):

PR #445, opened 2021-08-31
Print an warning in stderr when modifying read-only model

ogoffart:
... and improve the documentation of Model

### DIFF
--- a/api/sixtyfps-cpp/include/sixtyfps.h
+++ b/api/sixtyfps-cpp/include/sixtyfps.h
@@ -431,11 +431,17 @@ public:
     /// Returns the data for a particular row. This function should be called with `row <
     /// row_count()`.
     virtual ModelData row_data(int i) const = 0;
-    /// Sets the data for a particular row. This function should be called with `row < row_count()`.
-    /// If the model cannot support data changes, then it is ok to do nothing (default
-    /// implementation). If the model can update the data, the implementation should also call
-    /// row_changed.
-    virtual void set_row_data(int, const ModelData &) {};
+    /// Sets the data for a particular row.
+    ///
+    /// This function should only be called with `row < row_count()`.
+    ///
+    /// If the model cannot support data changes, then it is ok to do nothing.
+    /// The default implementation will print a warning to stderr.
+    ///
+    /// If the model can update the data, it should also call `row_changed`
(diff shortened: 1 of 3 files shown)


api/sixtyfps-cpp/include/sixtyfps.h
```diff
@@ -431,11 +431,17 @@ class Model
...
-    /// If the model cannot support data changes, then it is ok to do nothing (default
-    /// implementation). If the model can update the data, the implementation should also call
-    /// row_changed.
-    virtual void set_row_data(int, const ModelData &) {};
+    /// Sets the data for a particular row.
+    ///
+    /// This function should only be called with `row < row_count()`.
+    ///
+    /// If the model cannot support data changes, then it is ok to do nothing.
+    /// The default implementation will print a warning to stderr.
```
tronical, line 439: We could also consider using a return value to indicate to the caller that the insertion/mutation failed. But that's a breaking change, not for now :)

sixtyfps_runtime/corelib/model.rs
```diff
@@ -64,18 +64,92 @@ impl ModelNotify {
...
+    ///
+    /// This function should be called with `row < row_count()`, otherwise the implementation can panic.
+    ///
+    /// If the model cannot support data changes, then it is ok to do nothing.
+    /// The default implementation will print a warning to stderr.
+    ///
+    /// If the model can update the data, it should also call [`ModelNofity::row_changed`] on its
+    /// internal [`ModelNotify`].
+    fn set_row_data(&self, _row: usize, _data: Self::Data) {
+        eprintln!(
```
tronical, line 146: Maybe this could use debug_log! so that in wasm build it prints on console.log? Just a thought, doesn't have to been this change either.

tronical (approved):

api/sixtyfps-cpp/include/sixtyfps.h
ogoffart: But then who would print report the error?  The generated code that call that function?

tronical: Yes, when the caller is generated code (or we'll collect it in general binding error code that may need in the future). If the caller is the user, then they have a way of acting. Could even be annotated with a `__attribute__ ((__warn_unused_result__))`.

PR #447, opened 2021-08-31
Add support for global aliases

tronical:
When exporting an global multiple times under different names, make sure
that they alias in the generated code.

As a consequence, the compiler maintains the original unique name and in
Rust and C++ makes only the exported names public. In the interpreter
the internal name is theoretically still accessible from the outside.

### DIFF
--- a/sixtyfps_compiler/generator/rust.rs
+++ b/sixtyfps_compiler/generator/rust.rs
@@ -132,13 +132,20 @@ pub fn generate(doc: &Document, diag: &mut BuildDiagnostics) -> Option<TokenStre
         .globals
         .iter()
         .filter_map(|glob| {
-            (glob.visible_in_public_api() && glob.requires_code_generation())
-                .then(|| public_component_id(glob))
+            (glob.visible_in_public_api() && glob.requires_code_generation()).then(|| {
+                glob.exported_global_names
+                    .borrow()
+                    .iter()
+                    .map(|name| ident(&name))
+                    .collect::<Vec<_>>() // Would prefer not to collect here, but borrow() requires
+            })
         })
+        .flatten()
         .collect::<Vec<_>>();
 
     Some(quote! {
         #[allow(non_snake_case)]
+        #[allow(non_camel_case_types)]
         #[allow(clippy::style)]
         #[allow(clippy::complexity)]
         mod #compo_module {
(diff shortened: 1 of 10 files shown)


tronical:
This is semi-ready, there is at least one open question that may affect this change set:

This patch makes the previously internal name of globals publicly visible. In the interpreter by name, in Rust only when accessing the module directly and in C++ all types are in the same namespace anyway. Should I change the patch again to do the renaming again and thus never have an "internal" name?

ogoffart:
> Should I change the patch again to do the renaming again and thus never have an "internal" name

I think that'd be better, yes.  Otherwise it will be in the auto-completion and stuff.

ogoffart:
I think we should only be type aliases for the rare case when this happen.  In the normal case, there should be no typedef or type alias.
This will also be shown in the doc if the user generate the documentation for the generated code.

sixtyfps_compiler/generator/rust.rs
```diff
@@ -910,12 +917,21 @@ fn generate_component(
             }
         ))
     } else if component.is_global() && component.visible_in_public_api() {
+        let aliases = component
+            .exported_global_names
+            .borrow()
+            .iter()
+            .flat_map(|name| (name != &component.root_element.borrow().id).then(|| ident(&name)))
```
ogoffart, line 924: did you mean filter_map?

ogoffart: or  `filter(...).map(...)`

sixtyfps_compiler/passes/unique_id.rs
```diff
@@ -32,30 +30,17 @@ pub fn assign_unique_id(component: &Rc<Component>, exports: &[(String, Type)]) {
...
-        .collect::<HashMap<_, _>>();
+fn rename_globals(component: &Rc<Component>, mut count: u32) {
     for g in &component.used_types.borrow().globals {
         count += 1;
         let mut root = g.root_element.borrow_mut();
         if matches!(&root.base_type, Type::Builtin(_)) {
             // builtin global keeps its name
             root.id = g.id.clone();
-        } else if let Some(s) = exports.get(&ByAddress(g.clone())) {
-            root.id = s.to_string();
```
ogoffart, line 58: we can keep the "first" exported name.

tronical:
> > Should I change the patch again to do the renaming again and thus never have an "internal" name
> 
> I think that'd be better, yes. Otherwise it will be in the auto-completion and stuff.

Thanks a very good point! Thanks - I'll change that.

sixtyfps_compiler/generator/rust.rs
tronical: Indeed, that's easier to read than `flat_map` here.

sixtyfps_compiler/object_tree.rs
```diff
@@ -224,7 +224,7 @@ pub struct Component {
 
     /// Set to true if the component is a global that's exported, thus requires
     /// visibility in publicly generated API.
-    pub exported_global: Cell<bool>,
+    pub exported_global_names: RefCell<Vec<String>>,
```
ogoffart, line 227: The comment is out of date.

ogoffart (approved):
just one comment

sixtyfps_compiler/object_tree.rs
tronical: Oops. Thanks. Fixed :)

PR #449, opened 2021-09-01
Basic support for Multi-line TextInput

ogoffart:


sixtyfps_runtime/rendering_backends/gl/lib.rs
```diff
@@ -677,120 +608,124 @@ impl ItemRenderer for GLItemRenderer {
...
-        let vertical_alignment = text_input.vertical_alignment();
-        let metrics = self.draw_text_impl(
-            width,
-            height,
-            sixtyfps_corelib::items::TextInput::FIELD_OFFSETS.text.apply_pin(text_input),
+        let (min_select, max_select) = text_input.selection_anchor_and_cursor();
+        let cursor_pos = text_input.cursor_position();
+        let cursor_visible = cursor_pos >= 0 && text_input.cursor_visible();
+        let mut canvas = self.shared_data.canvas.borrow_mut();
+        let font_height = canvas.measure_font(paint).unwrap().height();
```
tronical, line 624: Just a suggestion, but since the same metric is also used in `layout_text_lines` perhaps it should be called `line_height`? 

The variable could also probably be declared and initialised inside the closure from the `metrics` parameter, unless I'm missing something?

sixtyfps_runtime/rendering_backends/gl/lib.rs
```diff
@@ -677,120 +608,124 @@ impl ItemRenderer for GLItemRenderer {
...
+                    canvas.fill_path(
+                        &mut rect_to_path(selection_rect),
+                        femtovg::Paint::color(to_femtovg_color(
+                            &text_input.selection_background_color(),
+                        )),
+                    );
+                    let mut selected_paint = paint;
+                    selected_paint
+                        .set_color(to_femtovg_color(&text_input.selection_foreground_color()));
+                    canvas
```
tronical, line 675: Like the old code had a comment about the paint order, I think this code deserves a comment that it needs fixing to work with text selection in ligatures. Or maybe a reminder that this all should go into femtovg.

tronical (approved):
Nice :)

sixtyfps_runtime/rendering_backends/gl/lib.rs
tronical: You're right, it is not the same.

sixtyfps_runtime/corelib/items/text.rs
```diff
@@ -364,14 +364,16 @@ impl Item for TextInput {
...
+                    } else if keycode == InternalKeyCode::Return && self.single_line() {
                         Self::FIELD_OFFSETS.accepted.apply_pin(self).call(&());
                         return KeyEventResult::EventAccepted;
                     }
                 }
 
                 // Only insert/interpreter non-control character strings
-                if event.text.is_empty() || event.text.as_str().chars().any(|ch| ch.is_control()) {
+                if event.text.is_empty()
+                    || event.text.as_str().chars().any(|ch| ch.is_control() && ch != '\n')
```
tronical, line 375: Should the `\n` check be tied to the `single_line()` property as well?

sixtyfps_runtime/corelib/items/text.rs
```diff
@@ -597,7 +599,11 @@ impl TextInput {
         self.delete_selection();
         let mut text: String = self.text().into();
         let cursor_pos = self.selection_anchor_and_cursor().1;
-        text.insert_str(cursor_pos, text_to_insert);
+        if text_to_insert.contains('\n') && self.single_line() {
+            text.insert_str(cursor_pos, &text_to_insert.replace('\n', " "));
```
tronical, line 603: I think that maybe this replacement should be done only at rendering time. If somebody sets a text with newline characters on a `TextInput` element that has `single-line: true` then it should also work, not only when editing text (where this code is called).

tronical (approved):

sixtyfps_runtime/corelib/items/text.rs
ogoffart: yeah, it works already with Qt. But without the actual replacement, we would need to do that replacement in a few more place to get it work with femtovg because it renders new line with a box, and nothing after it.

PR #450, opened 2021-09-01
Window abstraction cleanups

tronical:
I need to do more testing (results are looking okay at first glance), but code wise I think that this is ready for review.

### DIFF
--- a/sixtyfps_runtime/corelib/window.rs
+++ b/sixtyfps_runtime/corelib/window.rs
@@ -322,6 +322,21 @@ impl Window {
     pub fn set_scale_factor(&self, factor: f32) {
         self.scale_factor.as_ref().set(factor)
     }
+
+    /// Returns the font properties that are set on the root item if it's a Window item.
+    pub fn default_font_properties(&self) -> crate::graphics::FontRequest {
+        self.try_component()
+            .and_then(|component_rc| {
+                let component = ComponentRc::borrow_pin(&component_rc);
+                let root_item = component.as_ref().get_item_ref(0);
+                ItemRef::downcast_pin(root_item).map(
+                    |window_item: Pin<&crate::items::WindowItem>| {
+                        window_item.default_font_properties()
+                    },
+                )
+            })
+            .unwrap_or_default()
+    }
 }
 
 impl core::ops::Deref for Window {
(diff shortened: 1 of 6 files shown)


sixtyfps_runtime/corelib/window.rs
```diff
@@ -103,6 +105,23 @@ impl crate::properties::PropertyChangeHandler for WindowRedrawTracker {
     }
 }
 
+/// This enum describes the different ways a popup can be rendered by the back-end.
+pub enum PopupWindowLocation {
+    /// The popup is rendered in its own top-level window that is know to the windowing system.
+    TopLevel(Rc<Window>),
+    /// The popup is rendered as an embedded child window at the given position.
+    ChildWindow(Point),
```
ogoffart, line 113: I'm a bit torn there.
Even if the backend don't render as a separate window (which the GL backend eventually should do anyway), should there not be, conceptually, a different Rc<Window> for the popup anyway?

sixtyfps_runtime/rendering_backends/gl/graphics_window.rs
```diff
@@ -218,15 +124,21 @@ impl GraphicsWindow {
             }
         };
 
-        let backend = self.window_factory.as_ref()(window_builder);
+        let backend = crate::GLRenderer::new(
```
ogoffart, line 127: Is `backend` the right name here?

Do we even need GLRenderer at all

sixtyfps_runtime/corelib/window.rs
tronical: In torn, too. It’s tempting but otoh it means there will be a circular dependency. So if the application code decides to drop its ref to the window while a pop up is active, then I think the window will leak?

sixtyfps_runtime/rendering_backends/gl/graphics_window.rs
tronical: We do not, that’s right. I wanted to continue to with that but the patch become a bit large :-)

ogoffart (approved):

sixtyfps_runtime/corelib/window.rs
```diff
@@ -103,6 +105,23 @@ impl crate::properties::PropertyChangeHandler for WindowRedrawTracker {
...
+pub struct PopupWindow {
+    /// The location defines where the pop up is rendered.
+    pub location: PopupWindowLocation,
+    /// The component that is responsible for providing the popup content.
+    pub component: ComponentRc,
+}
+
 /// Structure that represent a Window in the runtime
 pub struct Window {
     /// FIXME! use Box instead;
```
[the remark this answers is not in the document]
tronical: I'll look into that.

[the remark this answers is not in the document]
tronical: I tried that locally and it's possible to remove the Rc from the platform window. Some parts become a little messier, some cleaner.

PR #451, opened 2021-09-01
C++: add blocking_invoke_from_event_loop

ogoffart:
A blocking version of invoke_from_event_loop

### DIFF
--- a/api/sixtyfps-cpp/include/sixtyfps.h
+++ b/api/sixtyfps-cpp/include/sixtyfps.h
@@ -21,6 +21,8 @@ LICENSE END */
 #include <chrono>
 #include <optional>
 #include <thread>
+#include <mutex>
+#include <condition_variable>
 
 namespace sixtyfps::cbindgen_private {
 // Workaround https://github.com/eqrion/cbindgen/issues/43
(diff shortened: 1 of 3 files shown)


ogoffart:
The test is failling on windows:

```
 3: Test command: D:\a\sixtyfps\cppbuild\api\sixtyfps-cpp\test_eventloop.exe
3: Test timeout computed to be: 1500
3: thread '<unnamed>' panicked at 'Initializing the event loop outside of the main thread is a significant cross-platform compatibility hazard. If you really, absolutely need to create an EventLoop on a different thread, please use the `EventLoopExtWindows::new_any_thread` function.', C:\Users\runneradmin\.cargo\registry\src\github.com-1ecc6299db9ec823\winit-0.25.0\src\platform_impl\windows\event_loop.rs:139:9
```

I wonder why. Does catch22 run the test in threads?

api/sixtyfps-cpp/include/sixtyfps.h
```diff
@@ -798,6 +802,55 @@ void invoke_from_event_loop(Functor f)
...
+/// ```
+/// #include "my_application_ui.h"
+/// #include <thread>
+///
+/// int main(int argc, char **argv)
+/// {
+///     auto ui = MyApplicationUI::create();
+///     ui->set_status_label("Pending");
+///
+///     std::thread worker_thread([&]{
```
tronical, line 826: I wonder if this should capture ui by value instead of reference. Wouldn’t that be safer?

ogoffart:
> I think the bug is that quit_event_loop does not work when called from another thread, because it accessed the thread local event loop (and ends up creating it).

aha..  i was under the impression that quit_event_loop could be called from a thread, but apparently not.

tronical (approved):

PR #452, opened 2021-09-02
Clean up GL backend code

tronical:
Remove GLRenderer and GLRendererData. Most of the fields fold nicely
into the MappedWindow.

### DIFF
--- a/sixtyfps_runtime/rendering_backends/gl/lib.rs
+++ b/sixtyfps_runtime/rendering_backends/gl/lib.rs
@@ -284,109 +284,6 @@ impl OpenGLContext {
     }
 }
 
-struct GLRendererData {
-    canvas: CanvasRc,
-
-    opengl_context: OpenGLContext,
-
-    // Layers that were scheduled for rendering where we can't delete the femtovg::ImageId yet
-    // because that can only happen after calling `flush`. Otherwise femtovg ends up processing
-    // `set_render_target` commands with image ids that have been deleted.
-    layer_images_to_delete_after_flush: RefCell<Vec<CachedImage>>,
-}
-
-pub struct GLRenderer {
-    pub(crate) shared_data: Rc<GLRendererData>,
-}
-
-impl GLRenderer {
-    pub(crate) fn new(
-        window_builder: winit::window::WindowBuilder,
-        #[cfg(target_arch = "wasm32")] canvas_id: &str,
-    ) -> GLRenderer {
-        #[cfg(target_arch = "wasm32")]
-        let (opengl_context, renderer) =
-            OpenGLContext::new_context_and_renderer(window_builder, canvas_id);
-        #[cfg(not(target_arch = "wasm32"))]
-        let (opengl_context, renderer) = OpenGLContext::new_context_and_renderer(window_builder);
-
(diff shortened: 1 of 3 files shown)


sixtyfps_runtime/corelib/item_rendering.rs
```diff
@@ -38,7 +38,7 @@ impl CachedRenderingData {
     pub fn get_or_update<T: Clone>(
         &self,
         cache: &RefCell<RenderingCache<T>>,
-        update_fn: impl FnOnce() -> T,
+        update_fn: impl FnMut() -> T,
```
ogoffart, line 41: Why the change?  Isn't  FnOnce stronger than FnMut ?

sixtyfps_runtime/rendering_backends/gl/lib.rs
```diff
@@ -876,13 +780,13 @@ impl ItemRenderer for GLItemRenderer {
                     let shadow_image_height = shadow_rect.max_y().ceil() as usize;
 
                     let shadow_image = CachedImage::new_empty_on_gpu(
-                        &self.shared_data.canvas,
+                        &self.canvas,
```
ogoffart, line 783: There is also a local `canvas` variable. Should that one not be used? Or what's the difference

ogoffart (approved):
after fixing the warning

sixtyfps_runtime/rendering_backends/gl/lib.rs
tronical: Well spotted, yes, that should be used instead.

sixtyfps_runtime/rendering_backends/gl/lib.rs
```diff
@@ -1029,20 +926,17 @@ impl ItemRenderer for GLItemRenderer {
         {
             let paint = layer_to_restore.image.as_paint();
 
-            self.shared_data
-                .layer_images_to_delete_after_flush
-                .borrow_mut()
-                .push(layer_to_restore.image);
+            self.layer_images_to_delete_after_flush.push(layer_to_restore.image);
```
tronical, line 929: Removing this use of interior mutability is what triggers the change from `FnOnce` to `FnMut`. If you prefer then I can keep that vector in a RefCell instead and stay with `FnOnce`.

sixtyfps_runtime/corelib/item_rendering.rs
tronical: See comment in `lib.rs`, there this allows replacing a use of interior mutability. But maybe that's not worth it and it's better to keep this `FnOnce`?

tronical: I think I spoke too soon. Not sure why, but if I change it back now it works :)

sixtyfps_runtime/rendering_backends/gl/lib.rs
tronical: In the forest of warnings I overlooked the one that really caused it, which was an immutable borrow of self outside the closure, that didn't go well with the mixture of immutable and mutable borrows inside. But fortunately that was easy to solve :)

ogoffart (approved):

PR #453, opened 2021-09-02
Internal cleanup: Remove canvas dependency from text layout code

tronical:
We can use the non-canvas bound TextContext and need the canvas only
in the closure when painting.


ogoffart (commented):
Great! :+1:  i wanted to do that as well

btw, will we have a problem when there are transformations (rotation, scale, ...)

ogoffart (approved):

PR #454, opened 2021-09-02
LineEdit is now able to scroll to keep the cursor visible

ogoffart:


sixtyfps_runtime/corelib/items/text.rs
```diff
@@ -275,23 +276,39 @@ impl Item for TextInput {
...
                 &|| self.unresolved_font_request(),
-                "********************",
-                None,
+                if text.is_empty() { "*" } else { text.as_str() },
+                max_width,
             )
-            .ceil();
+        };
+
+        // Stretch uses `round_layout` to explicitly align the top left and bottom right of layout nodes
```
tronical, line 289: Unrelated, but I wonder if we still need this :)

sixtyfps_runtime/rendering_backends/gl/graphics_window.rs
```diff
@@ -557,6 +557,65 @@ impl PlatformWindow for GraphicsWindow {
...
+
+        let mut result = Point::default();
+
+        let width = text_input.width() * scale_factor;
+        let height = text_input.height() * scale_factor;
+        if width <= 0. || height <= 0. {
+            return result;
+        }
+
+        let font = text_input
```
tronical, line 576: Also unrelated, but this is starting to look like an opportunity for moving the code into a shared function for determining the font.

tronical (approved):

PR #458, opened 2021-09-03
Allow selecting features in the CMake integration

tronical:
Provide the same features as in the Rust API crate, as CMake
options.


ogoffart (approved):
Please add an entry in the changelog

tronical:
Rebased to resolve conflict in changelog

PR #459, opened 2021-09-03
Optimize away Rectangle that serve no purpose

ogoffart:
Rectangles which do not draw anything and have no x or y don't need to be in
the item tree, we can just remove them.


sixtyfps_compiler/layout.rs
```diff
@@ -454,3 +454,34 @@ pub fn layout_info_type() -> Type {
         node: None,
     }
 }
+
+/// Get the implicit layout info of a particular element
+pub fn implicit_layout_info_call(elem: &ElementRc, orientation: Orientation) -> Expression {
+    if matches!(&elem.borrow().base_type ,
+                    Type::Builtin(base_type) if base_type.name == "Rectangle")
```
tronical, line 461: We have this pattern in a few places, might become worth a helper function in Element

tronical (approved):
Nice :)

tronical (approved):

sixtyfps_compiler/passes/optimize_useless_rectangles.rs
```diff
@@ -0,0 +1,71 @@
...
+//!
+//! Rectangles which do not draw anything and have no x or y don't need to be in
+//! the item tree, we can just remove them.
+
+use crate::{langtype::Type, object_tree::*};
+use std::rc::Rc;
+
+pub fn optimize_useless_rectangles(root_component: &Rc<Component>) {
+    let mut item_count = 0;
+    let mut optim_count = 0;
```
tronical, line 20: I suppose that this is unused?:)

sixtyfps_compiler/passes/optimize_useless_rectangles.rs
```diff
@@ -0,0 +1,65 @@
...
+    if e.is_flickable_viewport || e.child_of_layout {
+        return false;
+    };
+
+    let base_type = match &e.base_type {
+        Type::Builtin(base_type) if base_type.name == "Rectangle" => base_type,
+        _ => return false,
+    };
+
+    // Check that no Rectangle property other than height and width are set
```
tronical, line 54: What happens for example if a rectangle sets just width and height and a child binds to the parent’s width/height?

ogoffart: in that case, the height and width properties will be materialized by the move_declarations pass (via move_declarations::simplify_optimized_items )

tronical: Ah yes, that makes sense

PR #465, opened 2021-09-04
Add image-scaling property to Images to control how they are scaled

ryban:
Like CSS image-rendering it has "smooth" and "pixelated" options. Only OpenGL has been tested right now, have not tested WASM or Qt. Right now the first instance of a `@image-url()` will set the scaling for that specific image. The same image used from memory on the otherhand can all have a different scaling property.

This example shows scaling a 8x8 pixel image

Still need to update documentation, but I wanted to get other backends working, possibly fix the image-url issue, and ask if there is a better way to implement this first.

### DIFF
--- a/tests/cases/examples/image_rendering.60
+++ b/tests/cases/examples/image_rendering.60
@@ -7,7 +7,6 @@
     This file is also available under commercial licensing terms.
     Please contact info@sixtyfps.io for more information.
 LICENSE END */
-import { HorizontalLayout } from "sixtyfps_widgets.60";
 
 export MainWindow := Window {
     title: "SixtyFPS Image Scaling Example";
--- a/tests/cases/examples/image_rendering.60
+++ b/tests/cases/examples/image_rendering.60
@@ -1,0 +1,27 @@
+/* LICENSE BEGIN
+    This file is part of the SixtyFPS Project -- https://sixtyfps.io
+    Copyright (c) 2021 Olivier Goffart <olivier.goffart@sixtyfps.io>
+    Copyright (c) 2021 Simon Hausmann <simon.hausmann@sixtyfps.io>
+
+    SPDX-License-Identifier: GPL-3.0-only
+    This file is also available under commercial licensing terms.
+    Please contact info@sixtyfps.io for more information.
+LICENSE END */
+import { HorizontalLayout } from "sixtyfps_widgets.60";
+
+export MainWindow := Window {
+    title: "SixtyFPS Image Scaling Example";
+    preferred-width: 800px;
+    preferred-height: 600px;
+
+    HorizontalLayout {
+        Image {
+            source: @image-url("../../../examples/printerdemo/ui/images/cat_preview_round.png");
+            image-rendering: smooth;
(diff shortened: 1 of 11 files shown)


ogoffart:
Thanks for your contribution, this is very welcome.

It looks good to me, I just have a few minor point:
 - The CI is failing because you added the example in Cargo.toml, but you forgot to git add the examples. I don't think there should be a rust example for that, but if it is just a .60 file it can be added in tests/cases/examples
 - it is missing the Qt backend  implementation. It should be trivial to add as well (just add the right flag to the painter) But i'd understand if you don't have Qt and i can just add this myself
 - As you said, it's missing documentation and changelog entry.

Since the wasm backend is the same as the GL backend, i imagine it should work out of the box.

Then, there is the question of  naming.
With CSS, it is called  `image-rendering`  https://developer.mozilla.org/en-US/docs/Web/CSS/image-rendering
I think it might make more sense to keep the same name as CSS.  Maybe @tronical will have an opinion the name as well.

ryban:
Yeah, I forgot to remove the example from Cargo.toml because the example was not really interesting. I don't have Qt so I hadn't implemented it, but if someone else can implement it that would help. I did get wasm to compile after working through a perl issue and it does work as expected.

I chose `image-scaling` instead of `image-rendering` to be more direct on what the setting is doing, but thinking about it `image-scaling` could be misinterpreted as a scaling factor. Plus there is value in keeping with what people know.

tronical:
I very much like the name as proposed in the latest patch set :-) (image-rendering). Nice addition!

tests/cases/examples/image_rendering.60
```diff
@@ -0,0 +1,27 @@
+/* LICENSE BEGIN
+    This file is part of the SixtyFPS Project -- https://sixtyfps.io
+    Copyright (c) 2021 Olivier Goffart <olivier.goffart@sixtyfps.io>
+    Copyright (c) 2021 Simon Hausmann <simon.hausmann@sixtyfps.io>
+
+    SPDX-License-Identifier: GPL-3.0-only
+    This file is also available under commercial licensing terms.
+    Please contact info@sixtyfps.io for more information.
+LICENSE END */
+import { HorizontalLayout } from "sixtyfps_widgets.60";
```
ogoffart, line 10:
```suggestion
```

Not needed because HorizontalLayout is a builtin element

ogoffart (commented):
Looks good to me :+1: 

Only thing is to fix CI because the test has an error, and one should do `cargo fmt`

ogoffart:
Thank you!

I'll do the Qt backend changes now

ogoffart:
I done the Qt backend.

Btw, i noticed a bug: the image-rendering property is not properly applied if both image are the same file with the GL backend, because the cached texture contains the image-rendering flag, but it is not part of the cache key.

ryban:
Thanks for merging and adding Qt support. Yeah, which ever Image instance references a file first "locks in" the image-rendering setting for that image. I wasn't sure how to change the cache key.

PR #468, opened 2021-09-06
Improve borrow granularity

tronical:
This fixes the panic when using auto-reload with the viewer and the GL backend

### DIFF
--- a/sixtyfps_runtime/rendering_backends/gl/graphics_window.rs
+++ b/sixtyfps_runtime/rendering_backends/gl/graphics_window.rs
@@ -180,9 +180,7 @@ impl GraphicsWindow {
     pub fn default_font_properties(&self) -> FontRequest {
         self.self_weak.upgrade().unwrap().default_font_properties()
     }
-}
 
-impl GraphicsWindow {
     /// Draw the items of the specified `component` in the given window.
     pub fn draw(self: Rc<Self>) {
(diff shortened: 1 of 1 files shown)


sixtyfps_runtime/rendering_backends/gl/graphics_window.rs
```diff
@@ -355,56 +353,64 @@ impl PlatformWindow for GraphicsWindow {
...
-                if w > 0. {
-                    size.width = w as _;
-                }
-                if h > 0. {
-                    size.height = h as _;
+        if matches!(&*self.map_state.borrow(), GraphicsWindowBackendState::Unmapped) {
+            return;
+        }
+
+        let borrow_window = || {
```
ogoffart, line 360: note that we have this match in many place. I wonder if it is not worth haveing that as an associated method.

ogoffart (approved):

sixtyfps_runtime/rendering_backends/gl/graphics_window.rs
tronical: Yeah, I noticed the same. Started doing that to realise that they all differ slightly in what they need - this needs a mutable borrow, the other one doesn't and the third one doesn't need the granularity. It felt a bit like a "noise" change, but let me post it again and then we can look at it independent from this :)

PR #470, opened 2021-09-06
GL backend: clean up of window map state handling

tronical:
Use std::cell:Ref/RefMut to reduce the indentation and make it easier to
see where the borrow occurs.


tronical:
This is subjective, not entirely sure it's worth it :)

ogoffart:
Looks sensible, but i'll let you decide if it is worth it.

tronical:
I'll submit it, but I'll wrap the return value in an `Option` for a slightly safer API.

PR #472, opened 2021-09-06
Allow global singletons to have default callback handlers

### LINKED ISSUE #467: set default handler for globals in .60
This is currently not supported, but it should be possible:

```
global Foo := {
  callback bar(int)-> int;
  bar(x) => { return x + 1; }
}
```

tronical:
It's already working, we just need to remove the error handling :-)

Fixes #467


ogoffart (approved):
That was easy ;)

PR #476, opened 2021-09-07
small text polish fixes

tronical:


tronical:
Oh, I suppose that I should add a test for the ctrl+a short-cut :)

sixtyfps_compiler/widgets/common/common.60
```diff
@@ -43,7 +43,7 @@ export LineEditInner := Rectangle {
     placeholder := Text {
         height: 100%; width: 100%;
         vertical-alignment: center;
-        text: root.text == "" ? root.placeholder-text : "";
+        text: (root.text == "" && !root.has-focus) ? root.placeholder-text : "";
```
ogoffart, line 46: Why are you doing that?
Maybe it should depends on the style.
Here in my browser all the placeholder stay until i press  a character.

The Qt logic is a bit more complicated and depends on the alignment, i don't really understand why, but in general it still show the text until you enter something.
https://code.woboq.org/qt5/qtbase/src/widgets/widgets/qlineedit_p.h.html#_ZNK16QLineEditPrivate25shouldShowPlaceholderTextEv

tronical: You're right - I mistakenly thought this made more sense (and it looked better to me), but it's inconsistent. Maybe it was the lack of the cursor that confused, I'll fix that instead :)

ogoffart (approved):

PR #478, opened 2021-09-08
Fix panic when trying to access layout cache of destroyed items

ogoffart:
This can be reproduced by deleting the last item of the printer queue in the
printer demo.
It is a regression showing up because we now emit the MouseExit event after
the mouse grab as released.
The problem is that we upgrade the weak item, and call geometry() on it.
Calling geometry will re-evaluate the layout cache which will re-evaluate
the model which will result in the component being removed and the cache
entry having less item than expected.

It is ok to simply return 0. for these layout location since the item will
disapear anyway.


tronical (approved):
Maybe changelog worthy? :)

ogoffart:
> Maybe changelog worthy? :)

I don't think so, since that's a regression

PR #479, opened 2021-09-08
Make it possible to disable the GL backend in the CMake build

tronical:
Even if SIXTYFPS_BACKEND_GL is set to OFF, we would default to enabling
X11, which translates to enabling the x11 feature, which in turn
implicitly activates the GL backend. In addition, the interpreter Rust
API crate also enables the GL backend in its defeaults.

In order to make it possible to turn off the GL backend, three changes are necessary:

* The default backend no more selects a default :-). Instead the Rust
  API crate, Rust interpreter API crate and the CMake project define
  the same features and defaults.
* The C++ crate depends on the interpreter without its defaults (which enable GL)
* In CMake the X11 and Wayland features become dependent options, that
  are only show in the CMake configure UI if the GL backend is enabled.
  This way a GL-disabled build won't also pass --features x11.

### DIFF
--- a/sixtyfps_runtime/rendering_backends/default/Cargo.toml
+++ b/sixtyfps_runtime/rendering_backends/default/Cargo.toml
@@ -14,7 +14,6 @@ path = "lib.rs"
 [features]
 x11 = ["sixtyfps-rendering-backend-gl/x11"]
 wayland = ["sixtyfps-rendering-backend-gl/wayland"]
-default = ["sixtyfps-rendering-backend-gl"]
 
 [dependencies]
 sixtyfps-corelib = { version = "=0.1.2", path = "../../corelib" }
--- a/api/sixtyfps-cpp/CMakeLists.txt
+++ b/api/sixtyfps-cpp/CMakeLists.txt
@@ -10,6 +10,8 @@
 cmake_minimum_required(VERSION 3.16)
 project(SixtyFPS LANGUAGES CXX)
 
+include(CMakeDependentOption)
+
 include(FetchContent)
 FetchContent_Declare(
     Corrosion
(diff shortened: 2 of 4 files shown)


sixtyfps_runtime/rendering_backends/default/Cargo.toml
```diff
@@ -14,7 +14,6 @@ path = "lib.rs"
 [features]
 x11 = ["sixtyfps-rendering-backend-gl/x11"]
 wayland = ["sixtyfps-rendering-backend-gl/wayland"]
-default = ["sixtyfps-rendering-backend-gl"]
```
ogoffart, line 17: without that, i'm afraid that the gl backend might not be build by default for the rust crate

ogoffart: for example the LSP might also need that, and the viewer

tronical: Oh? Why not? It is for me, because api/sixtyfps-rs/Cargo.toml has this:

```toml
backend-gl = ["sixtyfps-rendering-backend-default/sixtyfps-rendering-backend-gl"]
backend-qt = ["sixtyfps-rendering-backend-default/sixtyfps-rendering-backend-qt"]
default = ["backend-gl", "backend-qt"]
```

tronical: Yes, that is true, all "frontends". Another option is to keep it here but use `default-features = false` in the C++, Rust API and interpreter crate. Would that be okay?

ogoffart: right,  and the interpreter do the same so it should be ok

ogoffart (approved):

PR #496, opened 2021-09-13
Add support for querying globals in the Rust/C++ interpreter API

### LINKED ISSUE #495: Add API to list the names of exported global singletons and their properties
API exists to list the properties (name, type) of components, but what is missing is API to list the names of exported global singletons and their properties.

This requires API in
* [ ] C++
* [ ] Rust

tronical:
Fixes #495

### DIFF
--- a/sixtyfps_compiler/object_tree.rs
+++ b/sixtyfps_compiler/object_tree.rs
@@ -284,7 +284,7 @@ impl Component {
         self.exported_global_names
             .borrow()
             .iter()
-            .filter(|name| &***name != &self.root_element.borrow().id)
+            .filter(|name| name.as_str() != &self.root_element.borrow().id)
             .map(|name| name.original_name())
             .collect()
     }
--- a/sixtyfps_compiler/object_tree.rs
+++ b/sixtyfps_compiler/object_tree.rs
@@ -161,7 +161,7 @@ impl Document {
         }
     }
 
-    pub fn exports(&self) -> &Vec<(String, Type)> {
+    pub fn exports(&self) -> &Vec<(ExportedName, Type)> {
         &self.exports.0
     }
 }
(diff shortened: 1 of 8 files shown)


sixtyfps_runtime/interpreter/dynamic_component.rs
```diff
@@ -300,6 +300,26 @@ pub struct ComponentDescription<'id> {
 
     /// compiled globals
     compiled_globals: Vec<crate::global_component::CompiledGlobal>,
+    /// Map of all exported global singletons and their public properties/callbacks
+    exported_globals: BTreeMap<String, Rc<BTreeMap<String, PropertyDeclaration>>>,
```
ogoffart, line 304: This information should be accessible from the compiled globals. Is that not working?

ogoffart (commented):
Looks good

sixtyfps_compiler/object_tree.rs
```diff
@@ -278,12 +278,14 @@ impl Component {
...
 
+    /// Returns the names of aliases to global singletons, exactly as
+    /// specified in the .60 markup (not normalized).
     pub fn global_aliases(&self) -> Vec<String> {
         self.exported_global_names
             .borrow()
             .iter()
-            .filter(|name| *name != &self.root_element.borrow().id)
-            .map(|name| name.to_string())
+            .filter(|name| &***name != &self.root_element.borrow().id)
```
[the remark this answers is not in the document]
tronical: Good idea! Will try that

sixtyfps_runtime/interpreter/dynamic_component.rs
tronical: Hmm, the issue is that accessing the original requires one of those guards, which breaks the impl Iterator + ‘_. We did the same with the public properties. But I could probably change this to a map to a vec index and store the public props in the compiled globals entries. I’ll try that, might look better.

tronical: I've done that now in a follow-up commit. It adds a bit more code overall, but it improves the locality (in terms of readability of the code).

ogoffart (approved):

PR #497, opened 2021-09-13
Fix panic when using height-for-width element in natively styled TabW…

### LINKED ISSUE #412: NativeTabWidget: runtime panic of binding loop when it contains a widget that has height-for-width
The following cpanics when using the native style because of a binding recursion

```
import { TabWidget } from "sixtyfps_widgets.60";
App := Window {
    main_layout := VerticalLayout {
        tab_widget := TabWidget {
            the_tab := Tab {
                title: "The Tab";
                some_layout := VerticalLayout {
                    alignment: start;
                    the_text := Text { text: "lorem ipsum dolor sit amet"; wrap: word-wrap; }
                }
            }
        }
    }
}
```

The problem is that the NativeTabWidget::layouting_info has unconditional dependencies to the content's width and height. Same for the code in NativeTabWidget::init that computes the padding.

tronical:
…idget

The metrics properties and the implementation of layout_info needs to
strictly reading cross-axis properties to avoid circular dependencies.

This patch applies a INT_MAX / 2 on the cross-axis to QStyle when
querying. Mostly the styles appear to just add/subtract margins, so this
value should not produce overflows while still providing an
approximation of "infinite" :)

Fixes #412

### DIFF
--- a/sixtyfps_runtime/rendering_backends/qt/qt_widgets/tabwidget.rs
+++ b/sixtyfps_runtime/rendering_backends/qt/qt_widgets/tabwidget.rs
@@ -39,11 +39,13 @@ impl Item for NativeTabWidget {
     fn init(self: Pin<&Self>, _window: &WindowRc) {
         #[derive(Default, Clone)]
         #[repr(C)]
-        struct TabWidgetRects {
-            content: qttypes::QRectF,
-            tabbar: qttypes::QRectF,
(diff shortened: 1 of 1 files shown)


ogoffart:
:+1:  but there could be a test

ogoffart:
Ah yes, I forgot the Qt backend is not used by our tests.

ogoffart (approved):

PR #501, opened 2021-09-14
Fix listview not updating its geometry in Rust when the entire model …

### LINKED ISSUE #500: Replacing model does not update ListView
The following reduced test-case replaces a model that's used in a `ListView` while it is visible. The new data only becomes visible after "resizing" the window.

```60
import { ListView } from "sixtyfps_widgets.60";
export Demo := Window {
    width:  300px;
    height: 300px;
    property<[string]> model: ["first", "second"];
    ListView {
        for name in model: Text {
            text: name;
        }
    }
    TouchArea {
        clicked => {
            root.model = ["third", "fourth"];
        }
    }
}
```

In cargo-ui this is reproducible when showing the dependencies pane and changing the package.

tronical:
…is changed

The geometry tracker did not track the model property itself, so it did not become dirty.

Fixes #500

### DIFF
--- a/sixtyfps_runtime/corelib/model.rs
+++ b/sixtyfps_runtime/corelib/model.rs
@@ -619,7 +619,8 @@ impl<C: RepeatedComponent + 'static> Repeater<C> {
         let geometry_changed = listview_geometry_tracker
             .evaluate_if_dirty(|| {
                 // Fetch the model again to make sure that it is a dependency of this geometry tracker.
-                let model = self.model().0.expect("invariant: model existence was checked earlier");
+                // invariant: model existence was checked earlier, so unwrap() should be safe.
+                let model = self.model().0.unwrap();
 
                 let listview_height = listview_height.get();
                 // Compute the element height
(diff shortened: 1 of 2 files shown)


sixtyfps_runtime/corelib/model.rs
```diff
@@ -618,6 +618,9 @@ impl<C: RepeatedComponent + 'static> Repeater<C> {
 
         let geometry_changed = listview_geometry_tracker
             .evaluate_if_dirty(|| {
+                // Fetch the model again to make sure that it is a dependency of this geometry tracker.
+                let model = self.model().0.expect("invariant: model existence was checked earlier");
```
tronical, line 622: An alternate way would be to mark the geometry tracker explicitly dirty in self.model() when we notice that the model is dirty. I find this solution a bit cleaner though - but happy to change it if desired.

ogoffart: I wonder if the invariant should be written in the expect block or in a.comment

ogoffart (approved):

sixtyfps_runtime/corelib/model.rs
tronical: I'll change it back to a comment block for now, but I like the idea of something more description coming up in the event that the panic happens as we forgot to adjust the code. But we can discuss that separately :)

PR #510, opened 2021-09-16
Rewrite build system part of C++ header file generation

tronical:
This first bumps the cmake version and then moves the cbindgen xtask into sixtyfps-cpp's build.rs.

The CMake bump could - in theory - be omitted, but short of the ability to forward environment variables to the corrosion build, we would have to generate header files in the source directory, which is also a kind of boo-boo :)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -4,4 +4,3 @@ node_modules
 tools/figma_import/target
 tools/figma_import/figma_output
 *.code-workspace
-api/sixtyfps-cpp/generated_include
--- a/.gitignore
+++ b/.gitignore
@@ -4,3 +4,4 @@ node_modules
 tools/figma_import/target
 tools/figma_import/figma_output
 *.code-workspace
+api/sixtyfps-cpp/generated_include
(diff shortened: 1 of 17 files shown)


tronical:
Also note: The test driver uses the path in api/sixtyfps-cpp/generated_include, not a temporary directory. This is implicit through the build-dependency on the sixtyfps-cpp crate - I don't think there's a way for the test-driver-cpp crate to provide an environment variable to the sixtyfps-cpp crate it depends on.

ogoffart:
Looks good, but I think it would be better to keep the generated header in the target directory (from the OUT variable instead of MANIFEST PATH.)

tronical:
> Looks good, but I think it would be better to keep the generated header in the target directory (from the OUT variable instead of MANIFEST PATH.)

Good point, I agree. I've added a patch on top that changes it to this mechanism - the directory defaults to a sub-directory from OUT_DIR and that path is conveyed to the test driver via cargo links meta-data.

ogoffart (approved):
Thanks, this is better

Issue #238, opened 2021-05-21
Viewer should support saving and loading data


matklad:
My app makes heavy use of data, which is set by the rust code. For this reason, `viewer` isn't very helpful: without data, there's not much to view! Filling data by hand is tedious (and I failed to do this for an 2d array), it would be cool if I can snapshot the data from my live app, load it into the viewer, and then tweak.

ogoffart:
Yeah, we could have an option to load properties from a json file for example.

ogoffart:
I added the good first issue tag because it should be not too difficult to set these property in viewer/main.rs (https://github.com/sixtyfpsui/sixtyfps/blob/master/tools/viewer/main.rs#L55)
We can simply call [ComponentInstance::set_property](https://docs.rs/sixtyfps-interpreter/0.0.6/sixtyfps_interpreter/struct.ComponentInstance.html#method.set_property) for each property.
One just has to write some code to create the `sixtyfps_interpreter::Value` from the json.

tronical:
This could be used to implement something like https://sanana.kiev.ua/index.php/yad / https://news.ycombinator.com/item?id=28443736

PR #518, opened 2021-09-23
Viewer: add ability to load and save property values from/to json

ogoffart:
issue #238

### DIFF
--- a/tools/viewer/main.rs
+++ b/tools/viewer/main.rs
@@ -54,7 +54,7 @@ thread_local! {static CURRENT_INSTANCE: std::cell::RefCell<Option<ComponentInsta
 fn main() -> Result<()> {
     let args = Cli::from_args();
 
-    if args.auto_reload && (args.save_data.is_some()) {
+    if args.auto_reload && args.save_data.is_some() {
         eprintln!("Cannot pass both --auto-reload and --save-data");
         std::process::exit(-1);
     }
--- a/tools/viewer/main.rs
+++ b/tools/viewer/main.rs
@@ -8,7 +8,7 @@
     Please contact info@sixtyfps.io for more information.
 LICENSE END */
 
-use sixtyfps_interpreter::ComponentInstance;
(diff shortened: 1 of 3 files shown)


tools/viewer/main.rs
```diff
@@ -39,13 +39,26 @@ struct Cli {
...
+    save_data: Option<std::path::PathBuf>,
 }
 
 thread_local! {static CURRENT_INSTANCE: std::cell::RefCell<Option<ComponentInstance>> = Default::default();}
 
 fn main() -> Result<()> {
     let args = Cli::from_args();
 
+    if args.auto_reload && (args.save_data.is_some()) {
+        eprintln!("Cannot pass both --auto-reload and --save-data");
```
tronical, line 58: Why not? :)

tools/viewer/main.rs
```diff
@@ -39,13 +39,26 @@ struct Cli {
...
+    #[structopt(long, name = "save data file")]
+    save_data: Option<std::path::PathBuf>,
 }
 
 thread_local! {static CURRENT_INSTANCE: std::cell::RefCell<Option<ComponentInstance>> = Default::default();}
 
 fn main() -> Result<()> {
     let args = Cli::from_args();
 
+    if args.auto_reload && (args.save_data.is_some()) {
```
tronical, line 57:
```suggestion
    if args.auto_reload && args.save_data.is_some() {
```

tronical (approved):

tools/viewer/main.rs
ogoffart: When do we save the properties?

tronical: Before reloading, and then maybe load again with the new ui loaded?

Kind of a way to persist properties across ui edit cycles.

tronical: (Btw, just a side comment - don’t let that hold you off from merging :)

PR #520, opened 2021-09-27
Add API to sixtyfps::Image in Rust and C++ to access the optional path

tronical:


sixtyfps_runtime/corelib/graphics/image.rs
```diff
@@ -304,6 +304,14 @@ impl Image {
             ImageInner::EmbeddedImage(buffer) => [buffer.width() as _, buffer.height() as _].into(),
         }
     }
+
+    /// Returns the path of the image on disk, if it was constructed via [`Self::load_from_path`].
+    pub fn path(&self) -> Option<&std::path::Path> {
```
ogoffart, line 309: (this function is not tested, doc tests would be nice, if possible)

ogoffart (approved):

sixtyfps_runtime/corelib/graphics/image.rs
tronical: Ah yes, it's only implicit through the C++ test. I'd prefer an explicit API test though, as otherwise I'm not sure what to use as value of the path. It feels strange if the documentation example uses `Path::new(env!("CARGO_MANIFEST_DIR")).join("../../tests/cases/blah.png")`. Unless that's okay perhaps?

tronical: Ah, doesn't look too bad after adding it :-)

PR #521, opened 2021-09-28
Add sixtyfps::Window::show_fullscreen() and show_normal()

tronical:
This adds the API that was missing to programmatically create
full-screen apps.

The API surface for Rust, C++ and NodeJS are show_normal() and
show_fullscreen() as new functions in the Window type.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,9 @@ All notable changes to this project will be documented in this file.
  - sixtyfps-viewer gained ability to read or save the property values to a json file with `--save-data` and `--load-data`
  - `sixtyfps::Image` has now a `path()` accessor function in Rust and C++ to access the optional path
    of the file on disk that's backing the image.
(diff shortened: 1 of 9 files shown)


ogoffart:
Just wondering if this shouldn't be a property on the window

tronical:
> Just wondering if this shouldn't be a property on the window

Yes, that's an alternate API, or perhaps complementary. So perhaps an enum

```rust
enum WindowVisibility {
    Hidden,
    Normal,
    Fullscreen,
    Minimized,
    Maximized,
}
```

and then a `Window::visibility` property.

That said, while I do think it makes sense to offer this property in `.60`, for example in order to react to it and toggle visibility of certain elements, I think the primary use of visibility control is in native code.

ogoffart:
Note that a window can be both minimized and maximized (whether it get maximized when it is un-minimized) and maybe full screen can be a flag as well .

tronical:
We have this API now, so the old attempt can go away :)

PR #523, opened 2021-09-29
Dialog

ogoffart:
Add the Dialog element 

https://github.com/sixtyfpsui/sixtyfps/issues/490

### DIFF
--- a/docs/builtin_elements.md
+++ b/docs/builtin_elements.md
@@ -654,10 +654,10 @@ Example := Window {
 Dialog is like a window, but it has buttons that are automatically laid out.
 
 A Dialog should have one main element for the content, that is not a button.
-And the window can have any number of `StandardButton` widget.
+And the window can have any number of `StandardButton` widgets.
 The button will be layed out in an order that depends on the platform.
 
-The `kind` property of the `StandardButton`s needs to be set to a specific value. it cannot be a complex expression.
+The `kind` property of the `StandardButton`s needs to be set to a specific value. It cannot be a complex expression.
 
 ### Properties
 
--- a/docs/builtin_elements.md
+++ b/docs/builtin_elements.md
@@ -649,6 +649,34 @@ Example := Window {
 }
 ```
 
+## `Dialog`
+
+Dialog is like a window, but it has buttons that are automatically laid out.
+
+A Dialog should have one main element for the content, that is not a button.
+And the window can have any number of `StandardButton` widget.
+The button will be layed out in an order that depends on the platform.
+
+The `kind` property of the `StandardButton`s needs to be set to a specific value. it cannot be a complex expression.
+
+### Properties
+
+* **`title`** (*string*): The window title that is shown in the title bar.
+* **`icon`** (*image*): The window icon shown in the title bar or the task bar on window managers supporting it.
+
+### Example
+
+```60
+import { StandardButton } from "sixtyfps_widgets.60";
+Example := Dialog {
+    Text {
+      text: "This is a dialog box";
+    }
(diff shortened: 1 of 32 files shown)


docs/builtin_elements.md
```diff
@@ -649,6 +649,34 @@ Example := Window {
 }
 ```
 
+## `Dialog`
+
+Dialog is like a window, but it has buttons that are automatically laid out.
+
+A Dialog should have one main element for the content, that is not a button.
+And the window can have any number of `StandardButton` widget.
```
tronical, line 657:
```suggestion
And the window can have any number of `StandardButton` widgets.
```

docs/builtin_elements.md
```diff
@@ -649,6 +649,34 @@ Example := Window {
...
+The button will be layed out in an order that depends on the platform.
+
+The `kind` property of the `StandardButton`s needs to be set to a specific value. it cannot be a complex expression.
```
tronical, line 660:
```suggestion
The `kind` property of the `StandardButton`s needs to be set to a specific value. It cannot be a complex expression.
```

sixtyfps_compiler/generator/rust.rs
```diff
@@ -1598,12 +1598,26 @@ fn compile_expression(expr: &Expression, component: &Rc<Component>) -> TokenStre
...
+                        cells: Slice::from_slice(&cells),
+                    })
+                })
+            } else {
+                quote!(solve_grid_layout(&GridLayoutData{
+                    size: #size,
+                    spacing: #spacing,
+                    padding: #padding,
+                    cells: Slice::from_slice(&#cells),
+                }))
```
tronical, line 1619: Would be nice to get rid of the else since it's the same solve_grid_layout call, but perhaps that would make the code harder to read again as it requires concatenating the token stream. Just a thought - feel free to ignore :)

sixtyfps_runtime/rendering_backends/qt/qt_widgets/button.rs
```diff
@@ -10,6 +10,98 @@ LICENSE END */
 
 use super::*;
 
+#[allow(nonstandard_style)]
+#[allow(unused)]
+mod standard_button {
+    // Generated with
+    // bindgen /usr/include/qt/QtWidgets/qstyle.h  --whitelist-type QStyle -- -I /usr/include/qt -xc++ | grep _StandardPixmap_ -A1
+    pub const QStyle_StandardPixmap_SP_TitleBarMenuButton: QStyle_StandardPixmap = 0;
```
tronical, line 18: I hope these are compatible in Qt 6 :)

sixtyfps_runtime/corelib/layout.rs
```diff
@@ -719,6 +719,78 @@ pub fn solve_path_layout(data: &PathLayoutData, repeater_indexes: Slice<u32>) ->
...
+    {
+        // KDE variant
+        add_buttons(cells, roles, &mut idx, DialogButtonRole::Help);
+        add_buttons(cells, roles, &mut idx, DialogButtonRole::Reset);
+        idx += 1;
+        add_buttons(cells, roles, &mut idx, DialogButtonRole::Accept);
+        add_buttons(cells, roles, &mut idx, DialogButtonRole::Apply);
+        add_buttons(cells, roles, &mut idx, DialogButtonRole::Reject);
+    } else {
+        // GNOME variant
```
tronical, line 784:
```suggestion
        // GNOME variant and fallback for WASM build
```

tronical (approved):
Nice!

PR #528, opened 2021-09-30
Fix support for OpenGL ES 2.0

tronical:
We default to querying GlLatest from glutin, which might run into a code
path in glutin's `bind_and_get_api` in `egl/mod.rs` that will return
None as version and pass that to `choose_fbconfig`. That function will
panic with unimplemented() if the bound API is GLES, because we need to
know which version of GLES. The bound API in turn might be GLES if
desktop GL is not supported.

To accomodate for this situation, this patch changes the GL context
creation logic to first try explicitly GLES 2.0 and then fall back to
the GlLatest variants.

In addition this pulls in a newer version of femtovg to fix FBO
rendering with stencil with GLES 2.0.

### DIFF
--- a/sixtyfps_runtime/rendering_backends/gl/lib.rs
+++ b/sixtyfps_runtime/rendering_backends/gl/lib.rs
@@ -160,24 +160,68 @@ impl OpenGLContext {
     ) -> (Self, femtovg::renderer::OpenGl) {
         #[cfg(not(target_arch = "wasm32"))]
         {
+            use crate::event_loop::EventLoopInterface;
             use glutin::ContextBuilder;
             let windowed_context = crate::event_loop::with_window_target(|event_loop| {
-                let builder = ContextBuilder::new().with_vsync(true);
-                // With latest Windows 10 and VmWare glutin's default for srgb produces surfaces that are always rendered black :(
-                #[cfg(target_os = "windows")]
-                let builder = builder.with_srgb(false);
+                // Try different strategies for creating an GL context. First request our "favorite", OpenGL ES 2.0,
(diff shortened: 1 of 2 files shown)


sixtyfps_runtime/rendering_backends/gl/lib.rs
```diff
@@ -160,24 +160,71 @@ impl OpenGLContext {
...
+                    },
+                ];
+
+                let mut context = None;
+                let mut last_err = None;
+                for factory_fn in context_factory_fns {
+                    match factory_fn(window_builder.clone(), event_loop) {
+                        Ok(new_context) => {
+                            context = Some(new_context);
+                            break;
```
ogoffart, line 215: this could be a return, could it not?  Then you don't need a mutable option

ogoffart (commented):
There should be a comment with a link to the glutin issue so we know why we do this hack, and maybe we revert it when glutin is fixed

sixtyfps_runtime/rendering_backends/gl/lib.rs
tronical: Oh, that’s true! I didn’t think that we’re in a closure. Thanks, will fix.

tronical:
> There should be a comment with a link to the glutin issue so we know why we do this hack, and maybe we revert it when glutin is fixed

Okay, that makes sense - will do. I have a strong feeling that if glutin is changed, it’ll be in favor of returning an error.

tronical:
Force pushed to fix the commit message text - otherwise no changes

ogoffart (approved):
It would be nice to squash the commit so it is easier to revert

tronical:
Ok, sounds good - I'll do that before merging.

PR #529, opened 2021-09-30
Tweak dialog error messages

tronical:
Replace "should" with "must"/"can only" to emphasize that this is an error.


ogoffart (approved):
(no need to make PR to fix typos)

tronical:
> (no need to make PR to fix typos)

(just wanted to make sure I didn't miss anything after splitting up the message into two :)

PR #530, opened 2021-09-30
Try to optimize CI cache

ogoffart:
let's see if it is faster


tronical (approved):
And even if it's the same speed, it's great to delegate the logic for removing workspace artefacts to an existing action. Fingers crossed it works :)

PR #533, opened 2021-10-01
Add callback to TouchArea and use it to fix the Slider and Scroll Bar

### LINKED ISSUE #494: Slider and Scrollbar should update as the mouse moved
Currently, the value of the slider and some other widgets is only updated when the mouse is released. (Fluent and ugly style). Also in 0.1.2 it was changed to only be released when over the widget which makes it difficult.

It should in fact be updated as the mouse moved.

ogoffart:
Closes #494

### DIFF
--- a/sixtyfps_compiler/widgets/fluent/sixtyfps_widgets_impl.60
+++ b/sixtyfps_compiler/widgets/fluent/sixtyfps_widgets_impl.60
@@ -120,33 +120,35 @@ ScrollBar := Rectangle {
    // border-color: Palette.button-background;
     border-width: 1px;
     property <bool> horizontal;
-    property<length> max;
+    property<length> maximum;
     property<length> page-size;
-    // this is always negative and migger than  -max
+    // this is always negative and bigger than  -maximum
     property<length> value;
 
     handle := Rectangle {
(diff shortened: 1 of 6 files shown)


sixtyfps_compiler/widgets/fluent/sixtyfps_widgets_impl.60
```diff
@@ -120,33 +120,35 @@ ScrollBar := Rectangle {
    // border-color: Palette.button-background;
     border-width: 1px;
     property <bool> horizontal;
-    property<length> max;
+    property<length> maximum;
```
tronical, line 123: Makes sense for consistency with the other widgets (and ScrollBar seems internal, so no problem), but we have an inconsistency here in the sense that the widgets use minimum/maximum and the layout uses min/max.

tronical (approved):
Yay :)

sixtyfps_compiler/widgets/fluent/sixtyfps_widgets_impl.60
ogoffart: This is an internal api only.

The problem is that i can't use a property called `max` or i can't call the function `max(..., ...)`
See also https://github.com/sixtyfpsui/sixtyfps/issues/273 currently, functions resolves last.

PR #534, opened 2021-10-01
Fix abrupt process abort when showing a window with constraints in We…

### LINKED ISSUE #532: Fullscreen rendering with Wayland/Weston aborts when minimum size exceeds the screen display
When showing a window on a screen managed by Weston the Wayland compositor and setting `SIXTYFPS_FULLSCREEN=1` *and* either the minimum width or height of the Window exceeds the corresponding dimension of the screen, the application aborts with an error message (dimensions changed of course):

```
xdg_wm_base@15: error 4: xdg_surface buffer (772 x 504) is larger than the configured fullscreen state (800 x 425)
```

It could be that we are applying constraints somewhere and trying to make the window surface bigger than the size that we received. It might also be something that's happening inside winit or glutin.

A workaround is to not set `SIXTYFPS_FULLSCREEN=1` and instead run Weston with the kiosk shell (for example `--shell=kiosk-shell.so`), as that supports clipping the window.

tronical:
…ston/Wayland

Don't apply constraints to the winit window if we're in fullscreen mode,
as that might result in us allocating a surface that's bigger than the
screen, which in turn may result in the process exiting right away.

Fixes #532

### DIFF
--- a/sixtyfps_runtime/rendering_backends/gl/graphics_window.rs
+++ b/sixtyfps_runtime/rendering_backends/gl/graphics_window.rs
@@ -433,7 +433,16 @@ impl PlatformWindow for GraphicsWindow {
         if h > 0. {
             size.height = h as _;
         }
-        self.borrow_mapped_window().unwrap().opengl_context.window().set_inner_size(size);
+        {
+            let mapped_window = self.borrow_mapped_window().unwrap();
+            let winit_window = mapped_window.opengl_context.window();
+            // If we're in fullscreen state, don't try to resize the window but maintain the surface
+            // size we've been assigned to from the windowing system. Weston/Wayland don't like it
+            // when we create a surface that's bigger than the screen due to constraints (#532).
+            if winit_window.fullscreen().is_none() {
+                winit_window.set_inner_size(size);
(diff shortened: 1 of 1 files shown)


sixtyfps_runtime/rendering_backends/gl/graphics_window.rs
```diff
@@ -433,7 +433,16 @@ impl PlatformWindow for GraphicsWindow {
...
             size.height = h as _;
         }
-        self.borrow_mapped_window().unwrap().opengl_context.window().set_inner_size(size);
+        {
+            let mapped_window = self.borrow_mapped_window().unwrap();
+            let winit_window = mapped_window.opengl_context.window();
+            // If we're in fullscreen state, don't try to resize the window but maintain the surface
+            // size we've been assigned to from the windowing system. Weston/Wayland don't like it
+            // when we create a surface that's bigger than the screen due to constraints (#532).
+            if !winit_window.fullscreen().is_some() {
```
tronical, line 442: Uh, I should write that as `is_none()`

ogoffart (approved):

PR #540, opened 2021-10-04
Fix granularity of run-time borrows in GL backend window handling

### LINKED ISSUE #539: Hiding window in callback panics with BorrowMutError in GL backend
### Discussed in https://github.com/sixtyfpsui/sixtyfps/discussions/536

<div type='discussions-op-text'>

<sup>Originally posted by **russoue** October  3, 2021</sup>
Sorry for this noob question.  I am trying to play with sixtyfps and had the following simple dialog in my code:

```rust
sixtyfps::sixtyfps! {
    import { StandardButton } from "sixtyfps_widgets.60";

    ConfigWindow := Window {
        title: "Hello world";
        width: 320px;
        height: 64px;
        callback user_answered(bool);
        VerticalLayout {
            Text {
                text: "Do you want to create a default configuration?";
                color: black;
            }
            HorizontalLayout {
                StandardButton {
                    kind: yes;
                    clicked => { self.text = "Yes!!!"; root.user_answered(true); }
                }
                StandardButton {
                    kind: no;
                    clicked => { self.text = "No!!!"; root.user_answered(false); }
                }
            }
        }
    }
}
```

I have the corresponding Rust code:

```rust
    let user_answer = Arc::new(Mutex::new(false));
    let config_window = ConfigWindow::new();
    let user_answer_clone = Arc::clone(&user_answer);
    config_window.on_user_answered(move |answer| {
        *user_answer_clone.lock().unwrap() = answer;
    });
    config_window.run();
```

This works but I want to close the window after the user clicks any of those buttons.  So far I could not find a way that works.  The closest I got is calling `hide()` in the callback:

```rust
    ...
    let config_window_weak = config_window.as_weak();
    config_window.on_user_answered(move |answer| {
        *user_answer_clone.lock().unwrap() = answer;
        let config_window = config_window_weak.unwrap();
        config_window.hide();
    });
    ...
```

But the above encounters a runtime borrow checker error: `thread 'main' panicked at 'already borrowed: BorrowMutError'`.

What's the best way to close a window from a callback?  Also, is there a dialog widget already which does what I am trying to do: ask a question and collect a yes/no or ok/cancel answer from the user?</div>

tronical:
When processing window specific events, or generally when accessing
the ALL_WINDOWS refcell, keep the time for borrow to an absolute
miniumum. That way any subsequent method calls may further acquire the
refcell.

This patch re-organizes the entire deeply nested window event handling
in a function with slightly less indentation.

Fixes a panic when calling hide on a Window from within a callback
handler.

Fixes #539

### DIFF
--- a/sixtyfps_runtime/rendering_backends/gl/event_loop.rs
+++ b/sixtyfps_runtime/rendering_backends/gl/event_loop.rs
@@ -193,7 +193,6 @@ fn process_window_event(
     cursor_pos: &mut Point,
     pressed: &mut bool,
 ) {
-    corelib::animations::update_animations();
     let runtime_window = window.self_weak.upgrade().unwrap();
     match event {
         WindowEvent::Resized(size) => {
@@ -213,6 +212,7 @@ fn process_window_event(
             }
         }
         WindowEvent::ReceivedCharacter(ch) => {
+            corelib::animations::update_animations();
             // On Windows, X11 and Wayland sequences like Ctrl+C will send a ReceivedCharacter after the pressed keyboard input event,
             // with a control character. We choose not to forward those but try to use the current key code instead.
             let text: Option<SharedString> = if ch.is_control() {
@@ -240,6 +240,7 @@ fn process_window_event(
             runtime_window.set_focus(have_focus);
         }
         WindowEvent::KeyboardInput { ref input, .. } => {
+            corelib::animations::update_animations();
             window.currently_pressed_key_code.set(match input.state {
                 winit::event::ElementState::Pressed => input.virtual_keycode.clone(),
                 _ => None,
@@ -282,6 +283,7 @@ fn process_window_event(
             window.set_current_keyboard_modifiers(modifiers);
         }
         WindowEvent::CursorMoved { position, .. } => {
+            corelib::animations::update_animations();
             let position = position.to_logical(window.scale_factor() as f64);
             *cursor_pos = euclid::point2(position.x, position.y);
             window.process_mouse_input(MouseEvent::MouseMoved { pos: *cursor_pos });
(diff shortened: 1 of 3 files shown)


tronical:
This carries a certain risk of regressions. I've looked carefully for wrong early returns etc. and I'll do some testing now, but I think generally this is an improvement also in terms of readability / nesting.

sixtyfps_runtime/rendering_backends/gl/event_loop.rs
```diff
@@ -170,6 +175,181 @@ impl std::fmt::Debug for CustomEvent {
...
+
+fn process_window_event(
+    window: Rc<crate::graphics_window::GraphicsWindow>,
+    event: WindowEvent,
+    quit_behavior: sixtyfps_corelib::backend::EventLoopQuitBehavior,
+    control_flow: &mut winit::event_loop::ControlFlow,
+    cursor_pos: &mut Point,
+    pressed: &mut bool,
+) {
+    corelib::animations::update_animations();
```
ogoffart, line 196: Do we really want to update animation for every single event?
What is the motivation?

tronical: Good point. That was a little over-eager, we previously only did that for input events. I'll push a change on top to restore that and then it's also a bit more visible where that was/is.

ogoffart:
The diff is hard to read since it mixes refactoring and the actual fix. I must admit i don't really see the changes that fixes the problem. But if it works, ship it :-)

ogoffart (approved):

tronical:
> The diff is hard to read since it mixes refactoring and the actual fix. I must admit i don't really see the changes that fixes the problem. But if it works, ship it :-)

Ah yeah, that's a bit hard to see :(. It's the `ALL_WINDOWS` access that has an `RefCell` inside, which was borrowed for way too long. So instead the access to that is encapsulated in that `window_by_id` function. As I started introducing it I noticed how it's repeatedly called for all window events, and I marched towards the window event handling function.

PR #544, opened 2021-10-05
Add support for embedding data in generated C++ code

tronical:
This allows compiling with SIXTYFPS_EMBED_RESOURCES=true and
images/fonts are embedded as inline variables.

Generated data is emitted into the header file as

   inline uint8_t sfps_embedded_resources_123[789] = {
       0x1, 0x2, 0x3,
   };

### DIFF
--- a/sixtyfps_compiler/generator/cpp.rs
+++ b/sixtyfps_compiler/generator/cpp.rs
@@ -604,11 +604,13 @@ pub fn generate(doc: &Document, diag: &mut BuildDiagnostics) -> Option<impl std:
 
             let mut init = "{ ".to_string();
 
+            use std::fmt::Write;
+
             for (index, byte) in data.iter().enumerate() {
                 if index > 0 {
                     init.push(',');
                 }
-                init.push_str(&format!("0x{:x}", byte));
+                write!(&mut init, "0x{:x}", byte).unwrap();
                 if index % 16 == 0 {
                     init.push('\n');
                 }
--- a/sixtyfps_compiler/generator/cpp.rs
+++ b/sixtyfps_compiler/generator/cpp.rs
@@ -181,6 +181,7 @@ mod cpp_ast {
     pub struct Var {
         pub ty: String,
         pub name: String,
+        pub array_size: Option<usize>,
         pub init: Option<String>,
     }
 
@@ -188,6 +189,9 @@ mod cpp_ast {
         fn fmt(&self, f: &mut Formatter<'_>) -> Result<(), Error> {
             indent(f)?;
             write!(f, "{} {}", self.ty, self.name)?;
+            if let Some(size) = self.array_size {
+                write!(f, "[{}]", size)?;
+            }
             if let Some(i) = &self.init {
                 write!(f, " = {}", i)?;
             }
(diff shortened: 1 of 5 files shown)


tronical:
There are some open questions on this:

1. Should there be an extra linux C++ build that builds with `SIXTYFPS_EMBED_RESOURCES=true` perhaps?
2. The slice casting is ugly, especially casting the constness away. Would it make sense to have a `ConstSlice`?
3. The symbol naming is the same as in Rust (except lower case), but it could be done different. Are tweaks there worth it or is it better to aim for generating .cpp files in the future? (not this patch set)

sixtyfps_compiler/generator/cpp.rs
```diff
@@ -839,6 +880,7 @@ fn generate_component(
                     ty: format!("sixtyfps::private_api::Property<{}>", cpp_model_data_type),
                     name: "model_data".into(),
                     init: None,
```
ogoffart, line 882: I guess all the `init:None` can also be left out. (But that's just extra)

sixtyfps_compiler/generator/cpp.rs
```diff
@@ -1777,7 +1837,12 @@ fn compile_expression(
             match resource_ref {
                 crate::expression_tree::ImageReference::None => r#"sixtyfps::Image()"#.to_string(),
                 crate::expression_tree::ImageReference::AbsolutePath(path) => format!(r#"sixtyfps::Image::load_from_path(sixtyfps::SharedString(u8"{}"))"#, escape_string(path.as_str())),
-                crate::expression_tree::ImageReference::EmbeddedData { .. } => unimplemented!("The C++ generator does not support resource embedding yet")
+                crate::expression_tree::ImageReference::EmbeddedData { resource_id, extension } => {
+                    let symbol = format!("sfps_embedded_resource_{}", resource_id);
+                    let data_slice = format!("sixtyfps::Slice<uint8_t>{{std::data({}), std::size({})}}", symbol, symbol);
+                    let extension_slice = format!("sixtyfps::Slice<uint8_t>{{const_cast<uint8_t *>(reinterpret_cast<const uint8_t *>(\"{}\")), {}}}", extension, extension.as_bytes().len());
+                    format!("sixtyfps::Image(sixtyfps::cbindgen_private::types::ImageInner::EmbeddedData({}, {}))", data_slice, extension_slice)
```
ogoffart, line 1844: minor: could be done with a single format!("") to avoid too many intermediate allocations

sixtyfps_compiler/generator/cpp.rs
```diff
@@ -592,6 +598,34 @@ pub fn generate(doc: &Document, diag: &mut BuildDiagnostics) -> Option<impl std:
...
+            let data =
+                std::fs::read(path).expect(&format!("unable to read file for embedding: {}", path));
+
+            let mut init = "{ ".to_string();
+
+            for (index, byte) in data.iter().enumerate() {
+                if index > 0 {
+                    init.push(',');
+                }
+                init.push_str(&format!("0x{:x}", byte));
```
ogoffart, line 612: could the format! here be avoided to avoid many allocations?
I believe you can use `write!` instead.

ogoffart:
There should be a test, now

sixtyfps_compiler/generator/cpp.rs
tronical: Ahh, `write!`. Thanks, that's what I was looking for but failed to find :)

ogoffart (approved):

PR #546, opened 2021-10-05
about logo

tronical:
This proposes a new about logo widget.


tronical:
There are at least two issues that could go into a follow-up:

1. Compress all embedded .svgs to .svgz. In this particular case this saves several KB.
2. React to clicking on the link by opening a web browser (xdc-open, etc.).

sixtyfps_compiler/generator/cpp.rs
```diff
@@ -599,8 +599,13 @@ pub fn generate(doc: &Document, diag: &mut BuildDiagnostics) -> Option<impl std:
 
     file.declarations.extend(doc.root_component.embedded_file_resources.borrow().iter().map(
         |(path, id)| {
-            let data =
-                std::fs::read(path).expect(&format!("unable to read file for embedding: {}", path));
+            let data = match path.strip_prefix("builtin:/") {
+                None => std::fs::read(path)
+                    .expect(&format!("unable to read file for embedding: {}", path)),
```
ogoffart, line 604: Pre existing: this should not be a panic, but a compilation error.
Or is it already handled before?
Well, in that case, it should be a string literal or a unwrap_or with a panic so there is no format in the non error case

sixtyfps_compiler/widgets/common/common.60
```diff
@@ -84,4 +84,27 @@ export TextEdit := ScrollView {
...
+            font-size: 24px;
+            font-weight: 700;
+            horizontal-alignment: center;
+        }
+        Image {
+            source: @image-url("sixtyfps_logo.svg");
+            preferred-width: 128px;
+        }
+        Text {
+            text: "Version 0.1.3\n© 2021 SixtyFPS GmbH\nhttps://sixtyfps.io/";
```
ogoffart, line 105: Not sure about the copyright there.  It might look like the all app has this copyright

sixtyfps_runtime/interpreter/Cargo.toml
```diff
@@ -30,6 +30,7 @@ derive_more = "0.99.5"
 generativity = "1"
 once_cell = "1.5"
 thiserror = "1"
+lazy_static = "1"
```
ogoffart, line 33: Lazy_static is the old pattern.  The once_cell is the new pattern that might go into the std lib (<https://doc.rust-lang.org/stable/std/lazy/index.html>)
So once_cell should be used instead if possible.

sixtyfps_runtime/interpreter/eval.rs
```diff
@@ -1079,3 +1097,16 @@ pub fn default_value_for_type(ty: &Type) -> Value {
         }
     }
 }
+
+lazy_static::lazy_static! {
+    static ref INTERNED_STRINGS: std::sync::Mutex<std::collections::HashSet<String>> = Default::default();
+}
+
+pub fn interned_str(s: &str) -> &'static str {
```
ogoffart, line 1105: Where is that used? I don't see the reference to it.

Is there maybe a crate we can use?

sixtyfps_runtime/interpreter/eval.rs
```diff
@@ -1079,3 +1097,16 @@ pub fn default_value_for_type(ty: &Type) -> Value {
...
+    let mut cache = INTERNED_STRINGS.lock().unwrap();
+    if !cache.contains(s) {
+        cache.insert(s.into());
+    }
+    let s: &str = cache.get(s).unwrap();
+    return unsafe { std::mem::transmute(s) };
```
ogoffart, line 1111: I don't think that's valid.
The String could be moved when the Hash is resized and the String might implement small string optimization in the future, who knows.

sixtyfps_runtime/interpreter/eval.rs
```diff
@@ -477,7 +477,25 @@ pub fn eval_expression(expression: &Expression, local_context: &mut EvalLocalCon
...
+
+                    let data = match path.strip_prefix("builtin:/") {
+                        None => unimplemented!(),
+                        Some(builtin) => {
+                         sixtyfps_compilerlib::library::load_file(std::path::Path::new(builtin))
+                                .expect("non-existent internal file referenced")
+                        }
+                    };
+                    Ok(corelib::graphics::Image::from(
+                        corelib::graphics::ImageInner::EmbeddedData{ data: corelib::slice::Slice::from_slice(data), format: corelib::slice::Slice::from_slice(interned_str(extension).as_bytes()) }
```
ogoffart, line 496: Ah, i see where interned_str is used now. Somehow the github interface was hiding it from me.

Do we really need to support embedding data in the interpreter for generic data? I believe this is enough to do it for the data coming from the library, for which the data is static.

ogoffart: Reading more, i see that the problem is the extension.
Could `load_file` also return the filename or extension as a &'static  (since the builtin library has the filenames built in)

tronical: You’re right, that’s an elegant solution.

sixtyfps_compiler/widgets/common/common.60
tronical: Okay, I'll remove it

sixtyfps_runtime/interpreter/Cargo.toml
tronical: Yep, lazy_static is gone again, thanks to your idea about the static ref :)

sixtyfps_compiler/tests/syntax/basic/image.60
```diff
@@ -17,6 +17,10 @@ SubElements := Rectangle {
         source: @image-url("foo.png");
```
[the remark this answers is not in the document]
tronical: It’s not a builtin:/ path, so we don’t try to embed it and thus don’t check at compile time if it exists.

sixtyfps_compiler/passes/embed_resources.rs
```diff
@@ -29,29 +39,41 @@ fn embed_resources_from_expression(
...
+                    *resource_ref = ImageReference::EmbeddedData {
+                        resource_id,
+                        extension: std::path::Path::new(path)
+                            .extension()
+                            .and_then(|e| e.to_str())
+                            .map(|x| x.to_string())
+                            .unwrap_or_default(),
+                    }
+                } else {
+                    diag.push_error("Cannot find image file".into(), source_location);
```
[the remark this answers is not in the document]
tronical: Ok, will do tomorrow:)

PR #630, opened 2021-11-03
[DOCS] Update cmake.md

nagayev:
Hello. Bulding any library from the source code is quit difficult. You should have compiler and time.
So I suppose to swap paragraphs `Building from Sources` and `Binary packages`


tronical (approved):
I think that's a good change.

This is also pointing out a missing piece here: The documentation (that is versioned) points to CI packages of a branch that may give different results than what's documented.

So for the next release we need to polish this to include the correct link to the GitHub release and mention which platforms are covered.

tronical:
Filed #631 - let's take this in. Thank you!

PR #633, opened 2021-11-03
Refactor the build_item_tree

ogoffart:
Use a trait instead of two functions

The idea is that it would allow to share more code with the other frontend, but in fact, i'm not sure if that's the case.

### DIFF
--- a/sixtyfps_compiler/generator.rs
+++ b/sixtyfps_compiler/generator.rs
@@ -143,42 +143,54 @@ pub fn build_array_helper(component: &Component, mut visit_item: impl FnMut(&Ele
     }
 }
 
+/// A reference to this trait is passed to the [`build_item_tree`] function.
+/// It can be used to build the array for the item tree.
+pub trait ItemTreeBuilder {
+    /// Some state that contains the code on how to access some particular component
+    type SubComponentState;
+
+    fn push_repeated_item(
+        &mut self,
+        item: &crate::object_tree::ElementRc,
+        repeater_count: u32,
+        parent_index: u32,
+        component_state: &Self::SubComponentState,
+    );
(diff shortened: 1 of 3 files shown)


sixtyfps_compiler/generator.rs
```diff
@@ -143,51 +143,63 @@ pub fn build_array_helper(component: &Component, mut visit_item: impl FnMut(&Ele
...
-    initial_state: ComponentState,
-    mut visit_item: impl FnMut(&ComponentState, &Rc<Component>, &ElementRc, u32, u32),
-    mut visit_sub_component: impl FnMut(
-        &ComponentState,
-        &Rc<Component>,
-        &ElementRc,
-        u32,
-    ) -> ComponentState,
+    initial_stack: &T::SubComponentStack,
+    helper: &mut T,
```
tronical, line 181: I suggest to call this parameter `builder` (since it must implement `ItemTreeBuilder`).

sixtyfps_compiler/generator.rs
```diff
@@ -143,51 +143,63 @@ pub fn build_array_helper(component: &Component, mut visit_item: impl FnMut(&Ele
     }
 }
 
+/// A reference to this trait is passed to the [`build_item_tree`] function.
+/// It can be used to build the array for the item tree.
+pub trait ItemTreeBuilder {
+    /// Some state that contains the code on how to access some particular component
+    type SubComponentStack;
```
tronical, line 150: In the two implementations this associated type is either a `String` or a new-typed `int` (the item index), neither of which I think are stacks or treated as such. I'd prefer "State" instead of "Stack".

tronical (approved):

PR #634, opened 2021-11-03
Fix the C++ printer demo with disabled inline

tronical:
For the following reduced test-case the order in how the dynamic nodes
in the item tree were generated (and dyn indices assigned) differed from
the way the visit_dynamic_children slots were generated:

```
Blah := Rectangle {
    for x in 1: Text {
        text: "Should be on the right";
    }
}

MainWindow := Window {
    width: 772px;
    height: 504px;
    Text {
        if (false): TouchArea {
        }
    }
    Blah {
        x: 200px;
    }
}
```

The item tree node was constructed using build_item_tree, which
basically assigned dyn index 0 to the "repater" for the touch area
and "1" to the one for the repeater inside the sub-component.

Afterwards we traversed the element tree - without descending into the
sub-components - to generate the fields and the dispatch in in the
dynamic visitor. Here a subtle order would result in a mismatch of
indices:

recurse_elem_level_order would end up visiting Text, Blah and then
Text's children, assigning the first dynamic index to Blah.

This is now fixed by merging the two iterations into one.

### DIFF
--- a/sixtyfps_compiler/generator/cpp.rs
+++ b/sixtyfps_compiler/generator/cpp.rs
@@ -12,6 +12,8 @@ LICENSE END */
 
 // cSpell:ignore cstdlib cmath constexpr nullptr decltype intptr uintptr
 
+use std::fmt::Write;
+
 /// This module contains some data structure that helps represent a C++ code.
 /// It is then rendered into an actual C++ text using the Display trait
 mod cpp_ast {
@@ -609,8 +611,6 @@ pub fn generate(doc: &Document, diag: &mut BuildDiagnostics) -> Option<impl std:
 
             let mut init = "{ ".to_string();
 
-            use std::fmt::Write;
-
             for (index, byte) in data.iter().enumerate() {
                 if index > 0 {
                     init.push(',');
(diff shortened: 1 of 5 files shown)


tronical:
Oh, hold on - this breaks some tests I think :(

tronical:
Fixed - was a -Werror problem :)

sixtyfps_compiler/generator/cpp.rs
```diff
@@ -1208,6 +1261,41 @@ fn generate_component(
...
+
+                self.init.push(format!("{}.init(self_weak.into_dyn());", member_name));
+
+                let sub_component_repeater_count = sub_component.repeater_count();
+                if sub_component_repeater_count > 0 {
+                    let mut case_code = String::new();
+                    for local_repeater_index in 0..sub_component_repeater_count {
+                        case_code.push_str("case ");
+                        case_code.push_str(&(repeater_count + local_repeater_index).to_string());
+                        case_code.push_str(": ");
```
ogoffart, line 1285:
```suggestion
                        write!(case_code, "case {}: ", repeater_count + local_repeater_index).unwrap();
```

ogoffart (approved):

ogoffart:
can there be a test?

PR #637, opened 2021-11-05
Document MSRV in the crate README

tronical:
Another option would be to spell out an entire policy, like this](https://github.com/BurntSushi/byteorder/commit/663358f9d29bddadc1a8e84290ec96925f2cb851). 

I'm fine with just the version, but happy to adjust this merge requests to include policy text like that if that's preferred.

### DIFF
--- a/api/sixtyfps-rs/README.md
+++ b/api/sixtyfps-rs/README.md
@@ -60,3 +60,7 @@ You can quickly try out the [examples](/examples) by cloning this repo and runni
 # Runs the "printerdemo" example
 cargo run --release --bin printerdemo
 ```
+
+### Minimum Supported Rust Version
+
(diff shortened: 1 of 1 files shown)


ogoffart (approved):
We need to remember all the places we need to update when changing that.
I guess a grep 1.54 will answer that question.

I'm expecting that this number might even increase before we do the next release

> Another option would be to spell out an entire policy, [like this](BurntSushi/byteorder@663358f).

I think we don't want to be so conservative at this point. (ie. "latest stable" would be good enough as far as i'm concerned.)

tronical:
> I think we don't want to be so conservative at this point. (ie. "latest stable" would be good enough as far as i'm concerned.)

Great, I agree.

PR #639, opened 2021-11-08
Allow handling Shift-Tab

phi-gamma:
Treating the Tab key as text disallows handling Shift-Tab because
the shift modifier will not be passed down. This patch makes the
handling of Tab analogous to Escape, Return and other non-text
keycodes.


ogoffart:
Thanks for the patch, that looks good!

I just had to handle the tab key also on the Qt backend.

PR #640, opened 2021-11-09
Make the inlining decision part of the CompilerConfiguration

ogoffart:


sixtyfps_compiler/lib.rs
```diff
@@ -87,11 +93,23 @@ impl CompilerConfiguration {
...
 
+        let inline_all_elements = match std::env::var("SIXTYFPS_DISABLE_INLINING") {
+            Ok(var) => {
+                !var.parse::<bool>().unwrap_or_else(|_|{
+                    panic!("SIXTYFPS_DISABLE_INLINING has incorrect value. Must be either unset, 'true' or 'false'")
+                })
+            }
+            // Currently, the interpreter needs the inlining to be on.
+            //Err(_) => output_format == crate::generator::OutputFormat::Interpreter,
+            Err(_) => false,
```
tronical, line 104:
```suggestion
            Err(_) => true,
```

I think `true` needs to be the default until we make this an opt-in (SIXTYFPS_FORCE_INLINING=true or so)

ogoffart: Yeah, that was wrong. 
I changed to enable it by default now, unless with the interpreter.

sixtyfps_compiler/lib.rs
```diff
@@ -87,11 +93,22 @@ impl CompilerConfiguration {
             }
         };
 
+        let inline_all_elements = match std::env::var("SIXTYFPS_DISABLE_INLINING") {
+            Ok(var) => {
+                !var.parse::<bool>().unwrap_or_else(|_|{
+                    panic!("SIXTYFPS_DISABLE_INLINING has incorrect value. Must be either unset, 'true' or 'false'")
+                })
+            }
```
tronical, line 101:
```suggestion
        let inline_all_elements = match std::env::var("SIXTYFPS_INLINING") {
            Ok(var) => {
                var.parse::<bool>().unwrap_or_else(|_|{
                    panic!("SIXTYFPS_INLINING has incorrect value. Must be either unset, 'true' or 'false'")
                })
            }
```

I think it's good that this merge request defaults to inlining disabled for Rust/C++. In that case though I think the environment variable that we can use to debug the other case should not require a double-negation (I have to set DISABLE_INLINING to false in order to *enable* inlining).

tronical (approved):

PR #641, opened 2021-11-10
Const Propagation

ogoffart:
Inline access to constant properties

Some refactoring was needed.

### DIFF
--- a/sixtyfps_compiler/passes/const_propagation.rs
+++ b/sixtyfps_compiler/passes/const_propagation.rs
@@ -35,6 +35,79 @@ fn simplify_expression(expr: &mut Expression) -> bool {
             }
             false
         }
+        Expression::BinaryExpression { lhs, op, rhs } => {
+            let mut can_inline = simplify_expression(lhs);
+            can_inline &= simplify_expression(rhs);
+
+            let new = match (*op, &mut **lhs, &mut **rhs) {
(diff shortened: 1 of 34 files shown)


sixtyfps_compiler/passes/materialize_fake_properties.rs
```diff
@@ -71,13 +85,14 @@ fn must_initialize(elem: &Element, prop: &str) -> bool {
     }
 }
 
-fn maybe_materialize(
-    property_declarations: &mut BTreeMap<String, PropertyDeclaration>,
+/// Returns a type if the property needs to be materialized, false otherwise
```
tronical, line 88: Nitpick: The docs say that the function returns a bool but it doesn’t do that anymore

sixtyfps_compiler/passes/const_propagation.rs
```diff
@@ -0,0 +1,85 @@
...
+        Expression::PropertyReference(nr) => {
+            if nr.is_constant() {
+                // Inline the constant value
+                if let Some(result) = extract_constant_property_reference(nr) {
+                    *expr = result;
+                } else {
+                    return false;
+                }
+            }
+            true
```
tronical, line 37: If nr is not constant, shouldn’t this function return false?

tronical: Nevermind :)

tronical (approved):

tronical (approved):

Issue #554, opened 2021-10-07
Windows 10 window resize and maximize wonky


L0uisc:
I don't know if it is related or similar, but I have the following issue with window resizing with the memory game tutorial:

![sixtyFPS_bug](https://user-images.githubusercontent.com/38226392/136283456-ec3d6157-0a9b-438b-886d-3bf60a883e94.gif)

I cannot resize, but the window does not become invisible when I try.

The window does not get maximized when clicking the "maximize" button. Instead, it is resized to the screen size, but not positioned correctly. Part of the window is thus off-screen. When I then click Maximize again, it does get maximized. Restoring down then also works normally.

OS: Windows 10
sixtyFPS: 0.1.3
rustc: 1.54

_Originally posted by @L0uisc in https://github.com/sixtyfpsui/sixtyfps/issues/553#issuecomment-937146128_

L0uisc:
Ok no, your tutorial only leaves a rather large margin around the right and bottom of the tiles. I made the main window dimensions 286px and the tiles fit snugly as expected.

ogoffart:
Fixed by https://github.com/slint-ui/slint/pull/642

PR #642, opened 2021-11-10
Fix the maximization button with a fixed inner size

Aursen:
This pull request fix the issue [#554](https://github.com/sixtyfpsui/sixtyfps/issues/554)

### DIFF
--- a/sixtyfps_runtime/rendering_backends/gl/graphics_window.rs
+++ b/sixtyfps_runtime/rendering_backends/gl/graphics_window.rs
@@ -93,26 +93,21 @@ impl GraphicsWindow {
         let component = ComponentRc::borrow_pin(&component);
         let root_item = component.as_ref().get_item_ref(0);
 
-        let (window_title, no_frame, w_height, w_width) = if let Some(window_item) =
+        let (window_title, no_frame, is_resizable) = if let Some(window_item) =
             ItemRef::downcast_pin::<corelib::items::WindowItem>(root_item)
         {
             (
                 window_item.title().to_string(),
                 window_item.no_frame(),
-                window_item.height(),
-                window_item.width(),
+                window_item.height() == 0. && window_item.width() == 0.,
             )
         } else {
-            ("SixtyFPS Window".to_string(), false, 0., 0.)
+            ("SixtyFPS Window".to_string(), false, true)
         };
 
-        let window_builder = winit::window::WindowBuilder::new().with_title(window_title);
-
-        let window_builder = if w_height > 0. && w_width > 0. {
-            window_builder.with_resizable(false)
-        } else {
-            window_builder
-        };
+        let window_builder = winit::window::WindowBuilder::new()
+            .with_title(window_title)
+            .with_resizable(is_resizable);
 
         let window_builder = if std::env::var("SIXTYFPS_FULLSCREEN").is_ok() {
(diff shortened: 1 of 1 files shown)


sixtyfps_runtime/rendering_backends/gl/graphics_window.rs
```diff
@@ -117,7 +117,7 @@ impl GraphicsWindow {
                 // Make sure that the window's inner size is in sync with the root window item's
                 // width/height.
                 self.set_geometry(s.width, s.height);
-                window_builder.with_inner_size(s)
+                window_builder.with_inner_size(s).with_resizable(false)
```
tronical, line 120: Thanks for the patch! I'm wondering, won't this make even windows that are meant to be resizable not resizable by the user?

Aursen: Indeed I had only tested with slide_puzzle and memory. If preferred-width or preferred-height is set, it also blocks the maximization. I'll fix it as soon as possible.

Aursen: I don't know if this is the best way.

Aursen (commented):
I don't know if this is the best way.

sixtyfps_runtime/rendering_backends/gl/graphics_window.rs
```diff
@@ -93,15 +93,27 @@ impl GraphicsWindow {
...
-            (window_item.title().to_string(), window_item.no_frame())
+            (
+                window_item.title().to_string(),
+                window_item.no_frame(),
+                window_item.height(),
+                window_item.width(),
+            )
         } else {
-            ("SixtyFPS Window".to_string(), false)
+            ("SixtyFPS Window".to_string(), false, 0., 0.)
```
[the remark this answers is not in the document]
Aursen: Yes it's true, i will correct

tronical:
Thanks! This is much more readable. I think similar logic is also needed in `apply_window_properties`. In the `if w <= 0. || h <= 0.` branch the window is resizable, otherwise it has a fixed size - this can be applied then to the `winit::Window` via `set_resizable(bool)`.

Aursen:
I It is not possible to do this condition in `apply_window_properties because` the size is already initialized with values.

> Thanks! This is much more readable. I think similar logic is also needed in `apply_window_properties`. In the `if w <= 0. || h <= 0.` branch the window is resizable, otherwise it has a fixed size - this can be applied then to the `winit::Window` via `set_resizable(bool)`.

tronical:
> It is not possible to do this condition in `apply_window_properties` because the size is already initialized with values.
> 
> > Thanks! This is much more readable. I think similar logic is also needed in `apply_window_properties`. In the `if w <= 0. || h <= 0.` branch the window is resizable, otherwise it has a fixed size - this can be applied then to the `winit::Window` via `set_resizable(bool)`.

Ah yes, that's true.

tronical (approved):

tronical:
I think that this might not work with the live-preview when re-using a window. For example when previewing a `.60` file that has a fixed width (window becomes not resizable) and then switching to something that uses a layout. But Olivier had the idea that it might be possible to detect this by looking at the min/max and see if they differ - and that's perhaps also how this could be done in apply_window_properties.

I'll check it out separately though, we can merge this first.

PR #643, opened 2021-11-11
Initial software rendering backend skeleton

tronical:
This change set applies a bunch of cleanups to the GL renderer to allow re-using parts of the winit code and adds a new mcu backend create:

  * It is not hooked up but needs to be selected explicitly by the application and selected with an `init()` call
  * The Rust API crate can be built without any backend.
  * There's no rendering yet (except some debug code to draw something into the frame buffer)

I think this is a reasonable starting point, but there are a few reasons to wait with merging:

 * I'm using a separate test app, but I could try to add a `examples/mcu/hello_world` example, which we don't have to link to from anywhere, but which can serve as in-repo testbed.
 * Don't merge this until the renderer traverses the item tree and some *something*.

I'm in favour of doing these in separate merge requests, but if that's not desired then I'd be happy to add to this one first.


sixtyfps_runtime/rendering_backends/default/lib.rs
```diff
@@ -27,57 +27,68 @@ In order for the crate to be available at runtime, they need to be added as feat
 #![doc(html_logo_url = "https://sixtyfps.io/resources/logo.drawio.svg")]
 
 cfg_if::cfg_if! {
-    if #[cfg(any(target_os="windows", target_os="macos", target_os = "linux", target_os = "dragonfly", target_os = "freebsd", target_os = "openbsd", target_os = "netbsd"))] {
+    if #[cfg(feature = "sixtyfps-rendering-backend-qt")] {
         use sixtyfps_rendering_backend_qt as default_backend;
```
ogoffart, line 31: I think there is no need for the `default_backend`  anymore.  This can just be removed.

ogoffart:
:+1:

ogoffart:
> Don't merge this until the renderer traverses the item tree and some something

Maybe it make sense to merge all the refactoring to avoid conflicts. For example, #642 touches the same code, does it not.

sixtyfps_runtime/rendering_backends/default/lib.rs
tronical: It's used to export what the api crate re-exports, and the compiler accesses as well:

```
pub use default_backend::{
            native_widgets, Backend, NativeGlobals, NativeWidgets, HAS_NATIVE_STYLE,
        };
```

tronical:
> > Don't merge this until the renderer traverses the item tree and some something
> 
> Maybe it make sense to merge all the refactoring to avoid conflicts. For example, #642 touches the same code, does it not.

Okay, I'll take those changes out of the merge request and push them directly.

sixtyfps_runtime/rendering_backends/default/lib.rs
ogoffart: ah yes.

ogoffart (approved):

PR #644, opened 2021-11-11
Move more of the winit related code from PlatformWindow impl into Win…

tronical:
…itWindow

This will allow code sharing with the simulator in the future.

Not 100% sure it's worth it, but OTOH I don't like the idea of duplicating it.

Also open to ideas to making it prettier :) - the window_handle access is kind of ugly.

### DIFF
--- a/sixtyfps_runtime/rendering_backends/gl/event_loop.rs
+++ b/sixtyfps_runtime/rendering_backends/gl/event_loop.rs
@@ -13,13 +13,15 @@ LICENSE END */
     [PlatformWindow] trait used by the generated code and the run-time to change
     aspects of windows on the screen.
 */
+use corelib::component::ComponentRc;
 use corelib::items::PointerEventButton;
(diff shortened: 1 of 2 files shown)


ogoffart (approved):
I have no objection to this, if you say this helps with code sharing.

Well, to me apply_window_properties does not really sounds like it belong in event_loop, but i must say i'm not really sure i understand the respective role of these two modules.

tronical:
> I have no objection to this, if you say this helps with code sharing.
> 
> Well, to me apply_window_properties does not really sounds like it belong in event_loop, but i must say i'm not really sure i understand the respective role of these two modules.

I wanted to rename `event_loop.rs` into `winit.rs` when done. Another option would be to split up `event_loop.rs` into `winit_loop.rs` and `winit_window.rs`.

`glwindow.rs` should have as little winit specific code as possible and as much gl rendering backend specific code as possible. Does that make sense?

PR #645, opened 2021-11-12
Fix transitioning from fixed size to resizable window with the GL bac…

tronical:
…kend

PR #642 fixed making the window not resizable when it has initially a
fixed size. This change applies a different logic with the same
resulting effect when the window item changes later, for example when
transiting in the live-preview from a fixed size to a resizable window -
by looking at the min/max constraints.

### DIFF
--- a/sixtyfps_runtime/rendering_backends/gl/event_loop.rs
+++ b/sixtyfps_runtime/rendering_backends/gl/event_loop.rs
@@ -134,6 +134,7 @@ pub trait WinitWindow: PlatformWindow {
         });
 
         let mut must_resize = false;
+        let mut resizable = None;
         let mut w = width;
         let mut h = height;
         if (size.width as f32 - w).abs() < 1. || (size.height as f32 - h).abs() < 1. {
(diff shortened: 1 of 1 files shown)


tronical:
cc @Aursen

sixtyfps_runtime/rendering_backends/gl/glwindow.rs
```diff
@@ -322,16 +323,17 @@ impl PlatformWindow for GLWindow {
...
                     must_resize = true;
                 }
+                let ver_info = component.as_ref().layout_info(Orientation::Vertical);
                 if h <= 0. {
-                    let info = component.as_ref().layout_info(Orientation::Vertical);
-                    h = info.preferred_bounded();
+                    h = ver_info.preferred_bounded();
                     must_resize = true;
                 }
+                resizable = Some(hor_info.min != hor_info.max && ver_info.min != ver_info.max);
```
ogoffart, line 336:
```suggestion
                resizable = Some(hor_info.min < hor_info.max && ver_info.min < ver_info.max);
```

I believe using `<` instead of `!=` will silence clippy warning about comparing float.
Alternatively one could also round it to integer.

ogoffart (approved):

sixtyfps_runtime/rendering_backends/gl/glwindow.rs
tronical: That's a good point and I find your suggestion slightly more readable :)

PR #648, opened 2021-11-15
preprocess the images at compile time

ogoffart:


sixtyfps_runtime/corelib/graphics/image.rs
```diff
@@ -179,6 +181,32 @@ impl PartialEq for SharedImageBuffer {
...
+    AlphaMap,
+}
+
+#[repr(C)]
+#[derive(Clone, PartialEq, Debug)]
+/// Some raw pixel data which is typically stored in the binary
+pub struct StaticTexture {
+    /// The position and size of the texture within the image
+    pub rect: IntRect,
+    /// FIXME: an enum
```
tronical, line 202: What does the FIXME here mean?

sixtyfps_compiler/passes/embed_images.rs
```diff
@@ -72,3 +59,216 @@ fn embed_images_from_expression(
...
+            i.pixels().flat_map(|(_, _, p)| IntoIterator::into_iter(p.0).take(3)).collect()
+        }
+        PixelFormat::Rgba => {
+            i.pixels().flat_map(|(_, _, p)| IntoIterator::into_iter(p.0)).collect()
+        }
+        PixelFormat::AlphaMap(_) => i.pixels().map(|(_, _, p)| p[3]).collect(),
+    }
+}
+
+fn load_image(file: crate::fileaccess::VirtualFile) -> image::ImageResult<image::RgbaImage> {
```
tronical, line 235: If the markup specifies a fixed width (or height) then it would probably be worthwhile to take that into account here. But I understand that this may also require knowledge about the device pixel ratio. Perhaps worth a TODO comment here?

tronical (approved):

PR #656, opened 2021-11-17
Add a simple fuzzer for the parser in the compilerlib

hunger:
Run with `cargo fuzz run parser`.

Fails very soon with this 3 ASCII char string: `/**`

### DIFF
--- a/sixtyfps_compiler/fuzz/.gitignore
+++ b/sixtyfps_compiler/fuzz/.gitignore
@@ -1,0 +1,3 @@
+target
+corpus
+artifacts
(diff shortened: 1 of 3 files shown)


hunger:
I wrote that file and I do not plan to claim you did, just to make the pipeline happy!

LoganDark:
> I wrote that file and I do not plan to claim you did, just to make the pipeline happy!

The license notice defines who owns the copyright to the project - not who wrote the files that are incorporated into the project. So don't worry, you're fine.

hunger:
@LoganDark: I still do not want to put some random person's name on my commit. That is not fair to that random person :-)

PR #657, opened 2021-11-17
Pr/fuzz fixes

hunger:
Some fixes for the first three panics in the compilerlib parser triggered by `cargo fuzz`.

### DIFF
--- a/sixtyfps_compiler/lexer.rs
+++ b/sixtyfps_compiler/lexer.rs
@@ -132,7 +132,7 @@ pub fn lex_string(text: &str, state: &mut LexState) -> usize {
                     state.template_string_stack.push(0);
                     return stop + 2;
                 }
-                end = stop + 1 + text[stop + 1 ..].chars().next().map_or(0, |c| c.len_utf8())
+                end = stop + 1 + text[stop + 1..].chars().next().map_or(0, |c| c.len_utf8())
             }
             _ => unreachable!(),
         }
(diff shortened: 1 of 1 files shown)


sixtyfps_compiler/lexer.rs
```diff
@@ -123,11 +124,21 @@ pub fn lex_string(text: &str, state: &mut LexState) -> usize {
                 return stop + 1;
             }
             b'\\' => {
+                if text_len <= stop + 1 {
+                    return text_len;
```
ogoffart, line 128: for consistency with the code just a few lines before

```suggestion
                    // FIXME: report an error for unterminated string
                    return 0;
```

sixtyfps_compiler/lexer.rs
```diff
@@ -123,11 +124,21 @@ pub fn lex_string(text: &str, state: &mut LexState) -> usize {
...
                 }
-                end = stop + 2;
+                end = stop + 1;
+                while end < text_len && text.as_bytes()[end] > 127u8 {
+                    // Iterate till end of UTF8 sequence started at stop + 1
+                    end += 1;
+                }
+                if end < text_len {
+                    end = end + 1 // begin of next UTF8 sequence
+                }
```
ogoffart, line 141: I think this skips one character too many. because something like "\€" for example, it will not see the end character.
This is, i believe, an error reported later anyway (invalid escape sequence) but we still need to lex that right.

```suggestion
                end = stop + 1 + text[stop + 1 ..].chars().next().map_or(0, |c| c.len_utf8())
```
(untested)

ogoffart:
You mention crash in the first commit message, but this is a panic, i believe.

ogoffart:
thanks!

PR #660, opened 2021-11-19
Make it easier to run built application binaries on Windows when usin…

tronical:
…g FetchContent

For the standalone build, place the dll into bin/ as that's also where we place the examples. When
we're used via FetchContent, place an extra copy of the dll into the binary dir,
as that's where the application binary will likely end up by default (as in the sixtyfps-cpp-template)

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -17,6 +17,9 @@ option(BUILD_TESTING "Build tests" OFF)
 add_feature_info(BUILD_TESTING BUILD_TESTING "configure whether to build the test suite")
 include(CTest)
 
+# We're building SixtyFPS as a standalone package, not embedded into another project via FetchContent
+set(SIXTYFPS_STANDALONE_BUILD ON)
+
 add_subdirectory(api/sixtyfps-cpp/)
 
 option(SIXTYFPS_BUILD_EXAMPLES "Build SixtyFPS Examples" OFF)
--- a/api/sixtyfps-cpp/CMakeLists.txt
+++ b/api/sixtyfps-cpp/CMakeLists.txt
@@ -166,8 +166,14 @@ include(GNUInstallDirs)
 install(FILES $<TARGET_FILE:sixtyfps-cpp-shared> TYPE LIB)
 if(WIN32)
     install(FILES $<TARGET_LINKER_FILE:sixtyfps-cpp-shared> TYPE LIB)
-    # Copy the dll to the top-level bin directory, where the examples will
-    # will also be located, so that they can find the dll.
+    # For our "own" build, copy the dll into the bin directory, where we collect
+    # all the example binaries. Otherwise copy to the top-level build directory, as
+    # then simple projects that just use `add_executable()` will end up with executables
+    # that run out of the box.
+    set(dll_destdir "${CMAKE_BINARY_DIR}")
+    if(SIXTYFPS_STANDALONE_BUILD)
+        set(dll_destdir "${dll_destdir}/bin")
+    endif()
     get_property(GENERATOR_IS_MULTI_CONFIG GLOBAL PROPERTY GENERATOR_IS_MULTI_CONFIG)
     if(GENERATOR_IS_MULTI_CONFIG)
         set(config_subdir_genex "$<CONFIG>/")
@@ -175,7 +181,7 @@ if(WIN32)
     add_custom_target(SixtyFPS_dll_convenience ALL DEPENDS sixtyfps-cpp-shared
         COMMAND ${CMAKE_COMMAND} -E copy_if_different
             $<TARGET_FILE:sixtyfps-cpp-shared>
-	    ${CMAKE_BINARY_DIR}/bin/${config_subdir_genex}$<TARGET_FILE_NAME:sixtyfps-cpp-shared>)
+	    ${dll_destdir}/${config_subdir_genex}$<TARGET_FILE_NAME:sixtyfps-cpp-shared>)
 endif()
 
 install(PROGRAMS $<TARGET_FILE:sixtyfps-compiler> TYPE BIN)


CMakeLists.txt
```diff
@@ -17,6 +17,9 @@ option(BUILD_TESTING "Build tests" OFF)
 add_feature_info(BUILD_TESTING BUILD_TESTING "configure whether to build the test suite")
 include(CTest)
 
+# We're building SixtyFPS as a standalone package, not embedded into another project via FetchContent
+set(STANDALONE_BUILD ON)
```
ogoffart, line 21: i guess it should be called SIXTYFPS_STANDALONE_BUILD

api/sixtyfps-cpp/CMakeLists.txt
```diff
@@ -168,14 +168,18 @@ if(WIN32)
     install(FILES $<TARGET_LINKER_FILE:sixtyfps-cpp-shared> TYPE LIB)
     # Copy the dll to the top-level bin directory, where the examples will
     # will also be located, so that they can find the dll.
```
ogoffart, line 170: this comment need to be updated

ogoffart (commented):
I'm not sure if we should do the copy to the top level...  Most project don't put their binary there, do they? Is it idiomatic to do that?
We could also do the copy within the sixtyfps-cpp-template CMakeLists.txt

tronical:
> I'm not sure if we should do the copy to the top level... Most project don't put their binary there, do they? Is it idiomatic to do that?

Well, where else should be put it?

> We could also do the copy within the sixtyfps-cpp-template CMakeLists.txt

I don't think our template should deviate from typical/minimal CMake code. By default `add_executable()` puts into into `${CMAKE_BINARY_DIR}`.

CMakeLists.txt
tronical: It's just a variable, it won't end up in the cache (won't leak into user CMakeCache.txt). But I can rename it, sure.

api/sixtyfps-cpp/CMakeLists.txt
tronical: That's true

CMakeLists.txt
ogoffart: ah, good point, i thought it was used in api/sixtyfps-cpp/CMakeLists.txt by anyone using sixtyfps

tronical:
Another option is to not do this change at all, of course. I'd also be okay with this. But I think this is a simple compromise that improves the changes of things working out of the box when people try it the first time.

Yes, once the re-organize the directory structure for their application binary they'll need to change their PATH for launching, on Windows.

ogoffart:
> Well, where else should be put it?

Not moving it at all. Or use CMAKE_RUNTIME_OUTPUT_DIRECTORY
You also need to account for the Debug/ and Release/ folder

> I don't think our template should deviate from typical/minimal CMake code. 

> By default add_executable() puts into into ${CMAKE_BINARY_DIR}.

It is `${CMAKE_CURRENT_BINARY_DIR}`  and maybe within a Release or Debug folder.
So it only works for very small project that have only one executable in the root.

ogoffart:
> Yes, once the re-organize the directory structure for their application binary they'll need to change their PATH for launching, on Windows.

But they will still have the sixtyfps.dll at the top level of their build directory for their monorepo and no way to change that.

> I don't think our template should deviate from typical/minimal CMake code.

I understand, but is that not a typical thing to do?
How does it work with other project, do people just set the PATH?

tronical:
> > Well, where else should be put it?
> 
> Not moving it at all. 

Yes, that's option (1) then (don't do anything unless we're building sixtyfps standalone)

> Or use CMAKE_RUNTIME_OUTPUT_DIRECTORY 

Yes, that's option (2) then: *If* the user sets CMAKE_RUNTIME_OUTPUT_DIRECTORY *before* including SixtyFPS then we can place it there. Otherwise we won't do anything. Advantage: User can control in once place where to place it. Disadvantage: The user needs to make that decision before the FetchContent bits; our template would have to set it.

> You also need to account for the Debug/ and Release/ folder

I believe that is accounted for. Where do you see it missing?

> > I don't think our template should deviate from typical/minimal CMake code.
> 
> > By default add_executable() puts into into ${CMAKE_BINARY_DIR}.
> 
> It is `${CMAKE_CURRENT_BINARY_DIR}` and maybe within a Release or Debug folder. So it only works for very small project that have only one executable in the root.

Ok, what do you suggest then? One of the above options?

ogoffart:
> Ok, what do you suggest then? One of the above options?

Option 2:  if CMAKE_RUNTIME_OUTPUT_DIRECTORY is set, but it there, otherwise don't do anything.
In fact, shouldn't that be done by corrosion?

ogoffart:
Shouldn't https://github.com/AndrewGaspar/corrosion/blob/master/cmake/Corrosion.cmake#L249-L251 account for CMAKE_RUNTIME_OUTPUT_DIRECTORY ?

tronical:
> > Ok, what do you suggest then? One of the above options?
> 
> Option 2: if CMAKE_RUNTIME_OUTPUT_DIRECTORY is set, but it there, otherwise don't do anything. 

Ok!

> In fact, shouldn't that be done by corrosion?

Hmm. Yes, you're right. That would be the best indeed. Then corrosion kind of behaves as if the developer wrote `add_library(myrustlib SHARED lib.rs some_module.rs)`, which sounds exactly right.

tronical:
I thought about it again and you're totally right - this is best solved inside Corrosion. This is basically AndrewGaspar/corrosion#65 . We just need to set `CMAKE_RUNTIME_OUTPUT_DIRECTORY` early on on our end and that's it. I'll close this MR in favour of that ticket in corrosion - something I wouldn't mind working on at some point. I looked into it a bit and I've come to the same conclusion as the ticket (which I discovered afterwards): `BYPRODUCTS` for the internal `_add_cargo_build` needs to be split into different parameters, one for each type of artefact (archive, library, runtime, pdb, etc.).

PR #661, opened 2021-11-19
Reorganize GitHub action workflows

tronical:
This separates the workflows for creating artefacts, such as docs, wasm demos, C++ packages or the lsp/viewer into a separate snapshot step. The ci.yaml will not produces published artefacts anymore and the CMake build is back to doing a debug build and run tests in debug.

The snapshot step is NOT set up yet to run automatically. I can integrate that into this change if desired, but otherwise I propose a follow-up change and we try this by running it once a day manually first.


tronical:
There is a chance that this ci.yaml might no pass the pull request CI because it references changed workflows with @master which are only in this changeset.

Referencing another workflow can only be done via tag/branch or sha1, for security reasons.

If this is not desired, then I can also duplicate that again between ci.yaml and the snapshot (although I prefer the proposed way)

tronical:
Ok, it doesn't seem to fail but the docs job etc. are no-ops :-)

ogoffart (approved):

PR #662, opened 2021-11-20
compilerlib: Fix an assert triggered by linear_gradient

hunger:
Just remove the variant with the unserscore: Using it triggers a debug_assert down
the line, so it does not seem to be used in the wild:-)

### DIFF
--- a/sixtyfps_compiler/parser/expressions.rs
+++ b/sixtyfps_compiler/parser/expressions.rs
@@ -20,6 +20,7 @@ use super::prelude::*;
 /// #aabbcc
 /// (something)
 /// @image-url("something")
+/// @image_url("something")
 /// some_id.some_property
 /// function_call()
 /// function_call(hello, world)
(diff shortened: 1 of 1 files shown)


hunger:
If you want to be more careful and keep @linear_gradient supported, then I can also change the debug_asser_eq!(...).

tronical:
> If you want to be more careful and keep @linear_gradient supported, then I can also change the debug_asser_eq!(...).

That’s the syntax we permit for the image url, changing the debug assert seems consistent to me.

Perhaps along with test coverage in the syntax tests.

Just my two cents, I defer to @ogoffart for the final call :-)

ogoffart (changes requested):
We claim in the documentation that foo_bar is the same as foo-bar, so we should accept `@linear_gradient`

hunger:
OK, I'll update the patch and remove or update the debug_assert instead.

hunger:
This time with both spellings intact and asserted for + a test for the alternate spelling of @image-url and @linear-gradient.

ogoffart (approved):

Issue #170, opened 2021-02-11
Add support for additional non-printable keys


tronical:
In particular for this ticket, we want

  * F1, F2, etc.
  * Page Up/Down
  * Cursor Up / Down
  * Insert / Delete
 
These need to be mapped and documented.

ogoffart:
This comment mention some code we could use: https://github.com/sixtyfpsui/sixtyfps/issues/638#issuecomment-962135961

jturcotte:
If you need to encode non-printable character keys as obscure codes encoded as Unicode in strings, then please add human-readable aliases so that event handlers don't have to look like this: https://stackoverflow.com/a/9310900/56315

ogoffart:
@jturcotte we will expose some constant string values in the `Keys` namespace, so one will be able to write

```
if (event.text == Keys.arrow-left) {
}
```

Now the questions is how do we define the names of the keys.
I think using the same unicode mapping as apple <https://www.unicode.org/Public/MAPPINGS/VENDORS/APPLE/CORPCHAR.TXT> is a good idea.
For the names, we could re-use these from the W3C spec: https://www.w3.org/TR/uievents-key/#named-key-attribute-values but snake cased.

ogoffart:
PR at https://github.com/sixtyfpsui/sixtyfps/pull/731

ogoffart:
Note that the PR uses camel case for keys, like `Keys.ArrowLeft`
We are unsure if we shouldn't use snake case instead.

PR #663, opened 2021-11-20
Fix 'Cannot parse string literal' for \t

jturcotte:
This is needed when wanting to handle the Tab key,
without this fix, the message above is reported in diagnostics
and Show preview isn't working.

### DIFF
--- a/sixtyfps_compiler/literals.rs
+++ b/sixtyfps_compiler/literals.rs
@@ -92,6 +92,7 @@ pub fn unescape_string(string: &str) -> Option<String> {
             b'"' => result += "\"",
             b'\\' => result += "\\",
             b'n' => result += "\n",
+            b't' => result += "\t",
             b'u' => {
                 if string.as_bytes().get(pos)? != &b'{' {
                     return None;
@@ -115,6 +116,7 @@ fn test_unescape_string() {
     assert_eq!(unescape_string(r#""fo\xa""#), None);
     assert_eq!(unescape_string(r#""fooo\""#), None);
     assert_eq!(unescape_string(r#""f\n\n\nf""#), Some("f\n\n\nf".into()));
+    assert_eq!(unescape_string(r#""f\t\t\tf""#), Some("f\t\t\tf".into()));
     assert_eq!(unescape_string(r#""music\♪xx""#), None);
     assert_eq!(unescape_string(r#""music\"♪\"🎝""#), Some("music\"♪\"🎝".into()));
     assert_eq!(unescape_string(r#""foo_bar"#), None);


sixtyfps_compiler/literals.rs
```diff
@@ -64,8 +64,8 @@ fn test_parse_color_literal() {
 }
 
 pub fn unescape_string(string: &str) -> Option<String> {
-    if string.contains('\n') {
-        // FIXME: new line in string literal not yet supported
+    if string.contains('\n') || string.contains('\t') {
```
jturcotte, line 67: I think I shouldn't be adding this, tab characters themselves should be possible to add in string literals.

But since I also don't understand why \n wouldn't be supported, I wanted to ask first.

jturcotte:
Oh, no this isn't specific to LSP, I thought my code compiled without this fix but I guess it didn't since now it doesn't work.

sixtyfps_compiler/literals.rs
ogoffart: the reason why we don't support line return in string literal is that we haven't made our mind of what the semantic would be. is a \n considered as a space or a \n, what about the identation.
I guess \t can stay \t, so i guess this change should indeed not be added.

ogoffart (approved):
Thanks, looks fine.

What's your use case for using `\t`?

ogoffart:
> What's your use case for using \t?

Ah, i see you already explain that in the message: 
> handle the Tab key,

In the mean time, you can do `\u{9}`

sixtyfps_compiler/literals.rs
jturcotte: Removed

ogoffart:
After a small discussion i had with Simon, we decided we don't want to support `\t` in our string literal. For the same reason rust don't support `\v` for vertical tab (<https://graydon2.dreamwidth.org/218040.html>): We don't think it it is something that should be used in our strings that are used to be shown on screen.

Now there is the valid use case of the key handling, but we want to get a generic solution for that (issue #170) and in the mean time, these keys can be used with `\u{...}`  syntax with the (to be documented) corresponding code.

Anyway, thanks again for sending a pull request, this is appreciated.

jturcotte:
The reason I've been brought this way is that doing `debug(event)` in the event handler after pressing the tab key printed me:

`"{ modifiers: { alt: false, control: false, meta: false, shift: false }, text: \t }"`

That was on Linux, but I just tried on macOS and there the text is even empty:

`"{ modifiers: { alt: false, control: false, meta: false, shift: false }, text:  }"`

So for discoverability reasons you might want to align that output with the input necessary to handle those keys.
In general though I'm not personally thrilled about using strings to handle keys not meant for direct text input. It means that I have to annotate the event handler with a comment explaining what key a piece of code is meant to handle while an enum would be self-explanatory. Hopefully you'll find a convenient way to express this either way.

PR #669, opened 2021-11-23
Fix the description of where @image-url looks for files

### LINKED ISSUE #665: image-url doesn't use the relative path of the folder containing cargo.toml on macOS
I'm following along with the memory game example but the paths for the images in the icons folder are not found. When I add in the full system path though then the images are found.

jrmoulton:
Fixes #665 

Update the rust memory-tile tutorial to reflect where @image-url looks for files to be relative to the same folder as the .60 file


tronical:
Thank you!

PR #672, opened 2021-11-23
Rename the sub component id to avoid error in the generated code with duplicated names

ogoffart:
This is a regression because of non-inlining discovered by the new "crater" test


tronical (approved):
Great :)

PR #681, opened 2021-11-24
WIP:  make corelib no_std

ogoffart:
It doesn't compile with no_std yet, we need to find a solution for thread_local

Also a bunch of dependencies still use the std, such as lyon, weak_table, maybe more

### DIFF
--- a/sixtyfps_runtime/corelib/Cargo.toml
+++ b/sixtyfps_runtime/corelib/Cargo.toml
@@ -15,7 +15,13 @@ path = "lib.rs"
 [features]
 # Allow the viewer to query at runtime information about item types
 rtti = []
+# Expose C ABI
 ffi = []
+# Use the standard library
+std = []
+
+default = ["std"]
+
 
 [dependencies]
 const-field-offset = { version = "0.1", path = "../../helper_crates/const-field-offset" }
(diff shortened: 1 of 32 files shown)


sixtyfps_runtime/corelib/Cargo.toml
```diff
@@ -15,7 +15,13 @@ path = "lib.rs"
 [features]
 # Allow the viewer to query at runtime information about item types
 rtti = []
+# Expose C ABI
 ffi = []
+# Use the standard library
+std = []
```
tronical, line 21: Is it intentional not to enable the std features of euclid, etc. here? (Just curious)

sixtyfps_runtime/corelib/lib.rs
```diff
@@ -21,6 +21,8 @@ only be used with `version = "=x.y.z"` in Cargo.toml.
 */
 #![doc(html_logo_url = "https://sixtyfps.io/resources/logo.drawio.svg")]
 #![deny(unsafe_code)]
+//TODO #![cfg_attr(not(feature = "std"), no_std)]
+extern crate alloc;
```
tronical, line 25: What does the TODO mean?

ogoffart: That we need to do this:

```suggestion
#![cfg_attr(not(feature = "std"), no_std)]
extern crate alloc;
```

When the crate can compile without the "std" feature :-)

sixtyfps_runtime/corelib/Cargo.toml
ogoffart: yes, i don't believe we need them, at least it compiles without it

tronical: That makes sense - great :)

sixtyfps_runtime/corelib/lib.rs
tronical: Ah yes, of course - sorry :)

PR #688, opened 2021-11-26
Native style: respond to the StyleChange event

### LINKED ISSUE #687: Issue with switching away from dark mode on MacOS native interface
When switching away from dark mode, things generally seem to break. 

This is on MacOS 12.0.1.

https://user-images.githubusercontent.com/25753364/143527325-1be202ac-b2a8-4b02-9f4c-3fea91342520.mov

ogoffart:
So that color scheme is updated when switching to the dark style, for example

Fixes #687


ogoffart:
I did not test on mac, but it works on linux.

tronical:
It does something on macOS, but not all colors are updated. For example the window background color remains. I guess that's a bug somewhere else.

tronical (approved):

tronical:
I can try to debug the background color issue, unless somebody beats me to it :) (maybe later today)

ogoffart:
> It does something on macOS, but not all colors are updated. For example the window background color remains. I guess that's a bug somewhere else.

You mean it does something compare to not having it? Because without this patch, the widget themselves should already adapt.
What this patch does is try to adapt to the other colors that are in the StyleMetrics, like the window background color.
So if the window background color don't change, it means this patch does nothing.

cmot17:
This is how it behaves with the latest commits:


https://user-images.githubusercontent.com/25753364/143608889-0cd16f27-c1dc-404f-80bb-19355e524254.mov

sixtyfps_runtime/rendering_backends/qt/qt_window.rs
```diff
@@ -1277,6 +1277,15 @@ impl PlatformWindow for QtWindow {
...
             widget_ptr->setWindowTitle(title);
             auto pal = widget_ptr->palette();
+
+            // If the background color is the same as what NativeStyleMetrics supplied from QGuiApplication::palette().color(QPalette::Window),
+            // then the setColor (implicitly setBrush) call will not detach the palette. However it will set the resolveMask, which due to the
+            // lack of a detach changes QGuiApplicationPrivate::app_pal's resolve mask and thus breaks future theme based palette changes.
+            // Therefore we force a detach
+            {
+                pal.setResolveMask(~pal.resolveMask());
+                pal.setResolveMask(~pal.resolveMask());
```
ogoffart, line 1287: why do you call it twice?
maybe you want to be explicit and call pal.detach() ?

tronical: detach() is a private function, so that won't compile with MSVC I think :(

PR #691, opened 2021-11-27
Cut text using Ctrl+x

RustemB:


sixtyfps_runtime/corelib/items/text.rs
```diff
@@ -396,6 +396,10 @@ impl Item for TextInput {
                     } else if event.text == "v" {
                         self.paste(window);
                         return KeyEventResult::EventAccepted;
+                    } else if event.text == "x" {
+                        self.copy();
+                        self.insert("", window);
```
tronical, line 401:
```suggestion
                        self.delete_selection(window);
```

tronical (commented):
Looks generally good - thanks! I made one small suggestion. Would you have time to also add a test case? (Like in tests/cases/text) otherwise I can also do that on Monday

RustemB:
Ok, i'll do it

RustemB:
Am I allowed to put my name to the header of a test file? :D

tronical:
> Am I allowed to put my name to the header of a test file? :D

Sorry, our script is overzealous :) - I’ll look into fixing that.

tronical (approved):

ogoffart:
Thanks for the changes!

Should there be a changelog entry?

RustemB:
I think should, because this is important feature, as in most toolkits

tronical:
I can add it after landing this, unless you'd like to add a commit on top of your branch with an entry :)

PR #694, opened 2021-11-29
Refactor Repeater model tracking

ogoffart:
Don't use weak-table

### DIFF
--- a/sixtyfps_runtime/corelib/model.rs
+++ b/sixtyfps_runtime/corelib/model.rs
@@ -10,24 +10,34 @@ LICENSE END */
 
 //! Model and Repeater
 
-use alloc::boxed::Box;
-use alloc::rc::{Rc, Weak};
-use alloc::vec::Vec;
-use core::cell::{Cell, RefCell};
-use core::pin::Pin;
+// Safety: we use pointer to Repeater in the DependencyList, bue the Drop of the Repeater
+// will remove them from the list so it will not be accessed after it is dropped
+#![allow(unsafe_code)]
 
 use crate::item_tree::TraversalOrder;
 use crate::items::ItemRef;
 use crate::layout::Orientation;
+use crate::properties::dependency_tracker::DependencyNode;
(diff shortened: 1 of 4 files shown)


ogoffart:
Note: the reason behind the change in the interpreter is that we only support inserting a ModelPeer in a single ModelNotify now.

sixtyfps_runtime/corelib/model.rs
```diff
@@ -92,15 +122,13 @@ impl ModelNotify {
 impl ModelTracker for ModelNotify {
     /// Attach one peer. The peer will be notified when the model changes
     fn attach_peer(&self, peer: ModelPeer) {
-        peer.inner.upgrade().map(|rc| self.inner.borrow_mut().insert(rc));
+        if let Some(peer) = peer.inner.upgrade() {
+            self.inner().project_ref().repeaters.append(peer.as_ref())
```
tronical, line 126: The terminology of ModelPeer and their attachment to a Model make me think that perhaps `repeaters` should be called `peers`.

sixtyfps_runtime/corelib/model.rs
```diff
@@ -47,37 +57,57 @@ impl ModelTracker for () {
     fn track_row_count_changes(&self) {}
 }
 
+#[pin_project]
+#[derive(Default)]
+struct ModelNotifyInner {
+    #[pin]
```
tronical, line 63: Nice to see pin_project in use :)

sixtyfps_runtime/corelib/model.rs
tronical: I wonder if it's worth to debug_assert here that the peer is not in the node list already?

tronical (approved):

sixtyfps_runtime/corelib/model.rs
ogoffart: it might be in the list when we change model.

true, peers might be a better name

PR #696, opened 2021-11-29
Add support for no_std to VRcMapped and VWeakMapped and use atomic-polyfill

tronical:
Port away from owning_ref, which doesn't support no_std, use a raw pointer instead.

### DIFF
--- a/helper_crates/vtable/CHANGELOG.md
+++ b/helper_crates/vtable/CHANGELOG.md
@@ -4,6 +4,8 @@ All notable changes to this crate will be documented in this file.
 ## [0.1.5]
 
  - Make it `#[no_std]`
+ - Use `atomic-polyfill` to support compiling to architectures where a polyfill
+   using critical sections is needed.
 
 ## [0.1.4] - 2021-11-24
 
--- a/helper_crates/vtable/Cargo.toml
+++ b/helper_crates/vtable/Cargo.toml
@@ -13,4 +13,4 @@ homepage = "https://sixtyfps.io"
 [dependencies]
 vtable-macro = { version = "=0.1.5", path = "./macro" }
 const-field-offset = { version = "0.1", path = "../const-field-offset" }
-owning_ref = { version = "0.4.1" }
+stable_deref_trait = { version = "1.2.0", default-features = false }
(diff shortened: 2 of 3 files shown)


helper_crates/vtable/Cargo.toml
```diff
@@ -13,3 +13,4 @@ homepage = "https://sixtyfps.io"
 [dependencies]
 vtable-macro = { version = "=0.1.6", path = "./macro" }
 const-field-offset = { version = "0.1", path = "../const-field-offset" }
+atomic-polyfill = "0.1.5"
```
ogoffart, line 16: i wonder if it should be a opetional dependency

helper_crates/vtable/CHANGELOG.md
```diff
@@ -1,6 +1,10 @@
 # Changelog
 All notable changes to this crate will be documented in this file.
 
+## [0.1.6]
```
ogoffart, line 4: 0.1.5 was not yet released, so no need to bump the version at this point.

helper_crates/vtable/src/vrc.rs
```diff
@@ -340,32 +331,19 @@ impl<VTable: VTableMetaDropInPlace + 'static, X: HasStaticVTable<VTable> + 'stat
     }
 }
 
-/// Safety: The data VRc manages is held by `VRcInner`, which maintains its address when the VRc
-/// is moved.
-unsafe impl<VTable: VTableMetaDropInPlace + 'static, X> owning_ref::StableAddress
```
ogoffart, line 345: I wonder if we still need to depends on the stable address trait crate and keep implemented these traits
otherwise it is a semver breaking change.

helper_crates/vtable/CHANGELOG.md
tronical: Ahh, thanks - will revert then.

helper_crates/vtable/src/vrc.rs
tronical: You're right - I'll fix that :)

helper_crates/vtable/Cargo.toml
tronical: No strong feeling here - although my guts feeling is that less "#ifdefs" are easier to maintain. Happy to change it to a #cfg if you prefer - let me know :)

ogoffart: you're right, it is fine to keep it because on other platform it is a no-op anyway

PR #699, opened 2021-11-30
Make floating point operations compile with no_std

tronical:
Use num_traits and libm instead.

### DIFF
--- a/sixtyfps_runtime/corelib/Cargo.toml
+++ b/sixtyfps_runtime/corelib/Cargo.toml
@@ -27,6 +27,8 @@ default = ["std"]
 # from a single core, and not in a interrupt or signal handler.
 unsafe_single_core = []
 
+libm = ["num-traits/libm", "euclid/libm"]
+
 [dependencies]
 const-field-offset = { version = "0.1", path = "../../helper_crates/const-field-offset" }
 vtable = { version="0.1.1", path = "../../helper_crates/vtable" }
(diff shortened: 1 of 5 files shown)


sixtyfps_runtime/corelib/Cargo.toml
```diff
@@ -51,6 +51,7 @@ unicode-segmentation = "1.8.0"
 # Note: the rgb version is extracted in ci.yaml for rustdoc builds
 rgb = "0.8.27"
 pin-project = "1"
+num-traits = { version = "0.2", default-features = false, features = ["libm"] }
```
ogoffart, line 54: do we really want to enable the libm feature unconditionally?  maybe this dependency should be optional

ogoffart (commented):
cool.  Just wondering if we shouldn't make this optional so we don't depend on libm form normal builds

sixtyfps_runtime/corelib/graphics/brush.rs
```diff
@@ -15,6 +15,9 @@ use super::{Color, Point};
 use crate::properties::InterpolatedPropertyValue;
 use crate::SharedVector;
 
+#[cfg(not(feature = "std"))]
```
tronical, line 18: This could also be `feature = "libm"` instead, but then... it's no_std that disables the float point ops it seems. I'm not sure how this should work identically on SOCs that do have a hardware floating point unit and us combining that with no_std. Maybe the ops *are* available then?

ogoffart (approved):

PR #700, opened 2021-11-30
Make it possible to compile sharedvector and sharedstring with atomic…

tronical:
…-polyfill

### DIFF
--- a/sixtyfps_runtime/corelib/Cargo.toml
+++ b/sixtyfps_runtime/corelib/Cargo.toml
@@ -51,6 +51,7 @@ unicode-segmentation = "1.8.0"
 # Note: the rgb version is extracted in ci.yaml for rustdoc builds
 rgb = "0.8.27"
 pin-project = "1"
+atomic-polyfill = { version = "0.1.5" }
 
 [target.'cfg(target_arch = "wasm32")'.dependencies]
 instant = { version = "0.1", features = [ "wasm-bindgen", "now" ] }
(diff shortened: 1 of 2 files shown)


sixtyfps_runtime/corelib/Cargo.toml
```diff
@@ -51,6 +53,7 @@ unicode-segmentation = "1.8.0"
 # Note: the rgb version is extracted in ci.yaml for rustdoc builds
 rgb = "0.8.27"
 pin-project = "1"
+atomic-polyfill = { version = "0.1.5", optional = true }
```
ogoffart, line 56: since this is not an optional dependency of vtable, we don't need to have it as an optional dependency

sixtyfps_runtime/corelib/Cargo.toml
```diff
@@ -27,6 +27,8 @@ default = ["std"]
 # from a single core, and not in a interrupt or signal handler.
 unsafe_single_core = []
 
+atomic_polyfill = ["triomphe/atomic-polyfill", "atomic-polyfill"]
```
ogoffart, line 30:
```suggestion
atomic-polyfill = ["triomphe/atomic-polyfill"]
```
i think that should be enough.  having an underscore is confusing.

But i think this doesn't need to be an optional dependency.  But can't be merged if the triomphe patch is not released.

(also, maybe we could use SharedVector within SharedString, instead of triomphe)

ogoffart (approved):

PR #702, opened 2021-11-30
Make it possible to delegate the timer handling to the backend

tronical:
When building with no_std, the backend can seed the Instant wrapper and
provide updates to it.

### DIFF
--- a/sixtyfps_runtime/corelib/animations.rs
+++ b/sixtyfps_runtime/corelib/animations.rs
@@ -75,6 +75,16 @@ impl Instant {
     pub fn duration_since(self, earlier: Instant) -> core::time::Duration {
         self - earlier
     }
+
+    /// Wrapper around [`std::time::Instant::now()`] that delegates to the backend
+    /// and allows working in no_std environments.
+    pub fn now() -> Self {
+        Self(Self::duration_since_start().as_millis() as u64)
+    }
+
+    fn duration_since_start() -> core::time::Duration {
+        crate::backend::instance().map(|backend| backend.duration_since_start()).unwrap_or_default()
+    }
 }
 
 /// The AnimationDriver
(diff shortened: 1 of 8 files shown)


sixtyfps_runtime/corelib/animations.rs
```diff
@@ -68,30 +68,54 @@ impl core::ops::SubAssign<core::time::Duration> for Instant {
...
+    }
+
+    fn duration_since_start() -> core::time::Duration {
+        #[cfg(feature = "std")]
+        {
+            let the_beginning = *INITIAL_INSTANT.get_or_init(|| instant::Instant::now());
+            return instant::Instant::now() - the_beginning;
+        }
+        #[cfg(not(feature = "std"))]
+        crate::backend::instance().unwrap().duration_since_start()
```
ogoffart, line 96: why not always delegate to the backend, but the default implementation would use instand::Instant

ogoffart (commented):
the adventage of having the backend always tell the time is that the testing backend can implement its own time passing time so that we can have test for the timer code (which currently is not tested,  only the animation are tested)

In a way, it would be nice if the timer and the animation used the same time reference.

sixtyfps_runtime/corelib/animations.rs
tronical: The main issue with that is that for the desktop builds we may not always have a backend yet. The API crate initializes it but we use timers early on in main.

ogoffart: and then we don't even need to store initial_instant in this crate.

tronical: Ok :)

ogoffart (approved):

PR #703, opened 2021-11-30
More on no_std

ogoffart:


tronical (approved):
Looks good to me. Just some build issue with the guards I guess

PR #706, opened 2021-12-01
Start making the generated code work with no_std

ogoffart:
There are still a couple of issues, like `register_font_from_memory`


tronical (approved):
Great that you found a solution for the item tree :)

PR #707, opened 2021-12-01
Use SharedVector internally in SharedString

ogoffart:
Get rid of the triomphe dependency


tronical:
This simplifies the implementation a lot - nice! I have just one question: it looks like a default constructed SharedString will - on the C++ side - will have data() return a dangling pointer. Differently put: it looks like the null termination is inconsistent.

Do we need it?

ogoffart:
Good point. I guess I need to special case data() to return a "\0" instead when the vector is empty.

tronical (approved):

Issue #577, opened 2021-10-14
Ability to change the Mouse cursor


ogoffart:
Discussed in https://github.com/sixtyfpsui/sixtyfps/discussions/576

We should add the ability to change the mouse cursor in TouchArea (or possibly on all element?)

```qml
   TouchArea {
      mouse-cursor: pointer;
   }
```

I'm  the `mouse-cursor:` name here, instead of `cursor:` like in CSS in case we extend the property to be used on all elements.

The value should probably be an enum consisting of the keywords from the [CSS spec](https://developer.mozilla.org/en-US/docs/Web/CSS/cursor).
Support for custom cursor could be added later.

lukas-jung:
Hey. I'd like to work on this issue if that's ok.

ogoffart:
@lukas-jung that'd be great, just let me know if you have any questions

lukas-jung:
I've got a first version running with both backends. there are some problems however.
not all css cursors are available in qt:
context-menu
cell
vertical-text
no-drop
all-scroll
zoom-in
zoom-out
and all unidirectional resizes.

Currently i'm replacing no-drop with not-allowed and the unidirectional resizes with their bidirectional counterpart.
All the other cursors can perhaps be added in a later Issue/PR. One could probably "steal" the bitmaps from webbrowsers or something like that.

However there is a second problem with the move cursor: since 'move' is a rust keyword I cannot use it as an enum variant!
Is it possible to map the .60 move to a rust enum variant that has another name?

Or perhaps the rust side of the enums should always use idiomatic camel case variant names and then the snake case .60 enum variants must be properly mapped to those.

PR #708, opened 2021-12-02
Feat: Add cursor_style property to TouchArea in order to change the cursor style 

jrmoulton:
Issue #577 
This PR is far from being done but I think I've added several of the components necessary to get this working. This is my first time really digging into the project and I'm hitting my limits of understanding of the interop between the rendering backends with rust and the overall architecture of the project. 

What would be the type of the property in ```TouchArea```? I have it as an enum but I don't think that's quite right.
How does the RC window receive the ```FIELD_OFFSETS``` values?
I can't find the right types/functions to work with qt and I'm not sure where I should be setting the cursor in gl. (some pointers on the high level architecture of how an RCwindow operates with a rendering window would be awesome)

Any direction is appreciated :)


sixtyfps_runtime/corelib/items.rs
```diff
@@ -391,6 +393,7 @@ impl Item for TouchArea {
             Self::FIELD_OFFSETS.mouse_y.apply_pin(self).set(pos.y);
         }
         Self::FIELD_OFFSETS.has_hover.apply_pin(self).set(!matches!(event, MouseEvent::MouseExit));
+        Self::FIELD_OFFSETS.cursor_style.apply_pin(self).set(CursorStyle::grab);
```
ogoffart, line 396: Here you shouldn't set the cursor_style property.
Instead, you should read the property with get(),  and pass that property to `_window`
You want to add interfaces to the `sixtyfps_corelib::window::PlatformWindow` like a set_cursor function there.
then re-implement that in the Qt and GL backend.

ogoffart (commented):
Thanks for the patch, this is a great start!

I added some comment inline. But the gist of it is that you need to add a function to `sixtyfps_corelib::window::PlatformWindow::set_cursor` and re-implement them in the backends. 

> the trait bound `CursorStyle: From<Value>` is not satisfied

It should be added in `declare_ValueType` in rtti.rs

ogoffart:
Closing in favour of https://github.com/sixtyfpsui/sixtyfps/pull/721

Thanks anyway again for submitting this, it is appreciated.

PR #715, opened 2021-12-04
Mark constant properties as not needing dependency nodes when accessed

ogoffart:
The first commit adds some debug facilities to the properties to aid debugging and benchmarking

This show that this patch reduces a lot the number of dependency nodes by not having them for property we know are constant.


ogoffart:
According to my early stats, it seems this reduce the amount of node by 1/2 to 1/3

tronical (approved):

PR #716, opened 2021-12-05
Update Flickable and ScrollView docs

jrmoulton:
Include documentation about needing to manually set the viewport-width and viewport-height when using a for loop.
I also added some more general documentation to `Flickable` but it likely needs to be edited as my explanation was a bit wordy. 
Are there other points in the documentation that need this same explanation or is this sufficient?

### DIFF
--- a/docs/builtin_elements.md
+++ b/docs/builtin_elements.md
@@ -578,6 +578,18 @@ FIXME: write docs
 
 FIXME: write docs
 
+The `Flickable` is a lower-level item that is the base for scrollable
+elements, such as the ScrollView widget. When the viewport-width or the
+viewport-height is greater than the parent width or parent height
+respectively the element becomes scrollable although the `Flickable`
+does not create a scrollbar. When using a for loop to populate the
+elements, the viewport-width and viewport-height are not automatically
+calculated and scrolling will not work properly until the viewport-width
+or viewport-height are set manually for the horizontal and vertical
+scrolling respectively. The ability to automatically calculate the
+viewport-width and viewport-height when using for loops may be added in
+the future and is tracked in issue #407.
+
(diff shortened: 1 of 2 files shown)


docs/builtin_elements.md
```diff
@@ -578,6 +578,18 @@ FIXME: write docs
 
 FIXME: write docs
```
ogoffart, line 579: I guess this line can go away 🙂

docs/builtin_elements.md
```diff
@@ -578,6 +578,18 @@ FIXME: write docs
...
 
+The `Flickable` is a lower-level item that is the base for scrollable
+elements, such as the ScrollView widget. When the viewport-width or the
+viewport-height is greater than the parent width or parent height
+respectively the element becomes scrollable although the `Flickable`
+does not create a scrollbar. When using a for loop to populate the
```
ogoffart, line 585: But one should say that it does automatically computes it based on its contents. With the exception of the bug if it is a for loop

ogoffart (approved):
Great, thanks.

It would be nice to have a "Properties" and "Example" section, like for the other items.
The properties are listed there: https://github.com/sixtyfpsui/sixtyfps/blob/cd47ef11005512bfd7799ce16a98dd4c71186a41/sixtyfps_compiler/builtins.60#L150-L161

But I actually can do that myself next week after meeting

ogoffart:
I don't think the CI failure is related to this change.

jrmoulton:
I added some of those changes. I didn't add all of the properties to Flickable because it looked like similar properties were omitted in the docs on the other elements as well. But if needed I can add those back in. I also didn't add descriptions one some properties because I'm not exactly sure what their functions are.

ogoffart:
Good enough, thank you!
I'll just do some adjustments after merging

PR #718, opened 2021-12-06
Fix issues related to alias to globals

ogoffart:
two problems
 - For the interpreter, properly handle the case of an alias to a global
 - For the rust generator, we can't return a reference to a property held
   in a global, so we must 'inline' the get_xx funciton on sub-components


tronical:
Ah, bummer that we can't return the reference anymore - but makes sense for aliases to globals. I vaguely recall that there was some reason for it, but perhaps it's outdated. Tests will show :)

tronical (approved):

PR #719, opened 2021-12-06
C++: add version macro

### LINKED ISSUE #705: Add version macros to the C++ API
In order to make it easier for a C++ application to support different versions of potentially slightly incompatible APIs, it would be useful to identify the current version via pre-processor macros.

I suggest we add

```cpp
#define SIXTYFPS_VERSION_MAJOR XX
#define SIXTYFPS_VERSION_MINOR YY
#define SIXTYFPS_VERSION_PATCH ZZ
#define SIXTYFPS_VERSION ((SIXTYFPS_VERSION_MAJOR << 16) | (SIXTYFPS_VERSION_MINOR << 8) | SIXTYFPS_VERSION_PATCH)
```

to `sixtyfps.h`.

ogoffart:
Fixes #705


ogoffart:
I wonder if we should document them, and how?

tronical:
I think the macros should be documented. Something like this in `sixtyfps.h` maybe?

```cppp
#if defined(DOXYGEN)
/// This macro expands to the to the numeric value of the major version of SixtyFPS you're developing against.
/// For example if you're using version 1.5.2, this macro will expand to 1.
#define SIXTYFPS_VERSION_MAJOR /* value computed */
/// This macro expands to the to the numeric value of the minor version of SixtyFPS you're developing against.
/// For example if you're using version 1.5.2, this macro will expand to 5.
#define SIXTYFPS_VERSION_MINOR  /* value computed */
/// This macro expands to the to the numeric value of the patch version of SixtyFPS you're developing against.
/// For example if you're using version 1.5.2, this macro will expand to 2.
#define SIXTYFPS_VERSION_PATCH  /* value computed */
#endif
```

tronical (approved):

api/sixtyfps-cpp/include/sixtyfps.h
```diff
@@ -42,6 +42,19 @@ struct ItemVTable;
 /// \endrst
 namespace sixtyfps {
 
+// these macro are defined in the generated sixtyfps_internal.h
+#if defined(DOXYGEN)
+/// This macro expands to the to the numeric value of the major version of SixtyFPS you're developing against.
+/// For example if you're using version 1.5.2, this macro will expand to 1.
+#define SIXTYFPS_VERSION_MAJOR SIXTYFPS_VERSION_MAJOR
```
tronical, line 49: Clever choice of the documented value :)

tronical (approved):

PR #721, opened 2021-12-06
Add Ability to change the Mouse cursor

### LINKED ISSUE #577: Ability to change the Mouse cursor

lukas-jung:
This fixes #577.

Some css cursors are not implemented, since they do not have a counterpart in the qt library.
Also the help cursor does not work on macos with the qt backend because of this 10(!) year old bug: https://bugreports.qt.io/browse/QTBUG-10711


sixtyfps_runtime/rendering_backends/qt/key_generated.rs
```diff
@@ -7,17 +7,17 @@
     This file is also available under commercial licensing terms.
     Please contact info@sixtyfps.io for more information.
 LICENSE END */
-/*! Generated with
+/*! Generated with Qt5 and
```
ogoffart, line 10: Yes, that's right, it's with Qt5.

Out of curiosity, did you see a difference with Qt6?
Because we kind of rely of the values being the same in Qt5 and Qt6, and we did not really verify that.

ogoffart (approved):
That's great, thank you.

The CI is failling because of the MCU backend, you can just add a dummy implementation of the function there.  And also you will need to run rustfmt.

Otherwise it's just missing documentation in `docs/builtin_elements.md` and a changelog entry.

Would be nice if you could just use that property in `tests/cases/elements/toucharea.60`
so this is tested

ogoffart:
Looks like the r#move also breaks cbindgen: https://github.com/eqrion/cbindgen/issues/410

ogoffart:
I'd say we can leave "move" out for now and try to fix cbingen later

ogoffart:
Thanks a lot for the patch!

Vollbrecht:
Is there a missing test for the now builtin property mouse-cursor and the MouseCursor enum?
I get an "Unknown property mouse-cursor in TouchArea" and "Unknown unqualified identifier 'crosshair'" in the current build, but it is compiling and working as intended for me.

minimal example where it happens is : 
```
 export AppWindow := Window {   
    VerticalBox { 
       Rectangle {
            border-color: #000000;
            border-width: 3px;
            min-width: 640px;
            min-height: 480px;                         
            TouchArea {                    
                mouse-cursor: crosshair;
            }              
    }    
}

``

tronical:
Where do you see those? In vs code / the lsp?

Could it be that this feature is not in the latest release and thus requires nightly?

Vollbrecht:
i am using the latest git master branch and i get it in vscode

tronical:
Maybe you need to uninstall the vs code extension and install the nightly https://marketplace.visualstudio.com/items?itemName=SixtyFPS.sixtyfps-vscode-nightly 

(best not to install both)

Vollbrecht:
thank you! using the nightly extension fix it for me. so there is no issue on my site anymore.

Issue #725, opened 2021-12-07
The opacity property should not be applied to children of an element but apply evenly to the element and the children


tronical:
The following test case is currently rendered incorrectly (different than the way for example opacity works in CSS):

```
export Demo := Window {
    preferred-width: 800px;
    preferred-height: 600px;

    Rectangle {
        opacity: 0.5;
        width: 100px;
        height: 100px;

        background: red;

        Rectangle {
            background: blue;
            width: 50px;
            height: 50px;
            x: 25px;
            y: 25px;
        }
    }
}
```

Current (wrong) output (with the GL backend):

<img width="90" alt="wrong" src="https://user-images.githubusercontent.com/1486/145045825-d0a12905-c9bc-4f46-b7f0-a1a847a8cb0e.png">

However the red and the blue rectangle should be rendered without opacity and then blended against the background with the specified opacity:

<img width="90" alt="correct" src="https://user-images.githubusercontent.com/1486/145045917-c04a324e-26d5-49b2-ad81-d6c799393449.png">

Issue #714, opened 2021-12-03
The Rectangle's opacity is not applied to the drop shadow


tronical:
Small test-case included:

```
export Demo := Window {
    width:  300px;
    height: 300px;
    background: red;
    Rectangle {
        width: 100px;
        height: 100px;
//        background: green;
        drop-shadow-color: blue;
        drop-shadow-blur: 10px;
        drop-shadow-offset-x: 10px;
        drop-shadow-offset-y: 10px;
        opacity: 0.3;
    }
}
```

Happens at the least with the GL renderer.

ogoffart:
Fixed in recent slint

PR #726, opened 2021-12-07
Fix opacity property not affecting the shadow color

tronical:
The opacity on a rectangle should be applied before the shadow, so that it has
a visible effect.

This is a partial fix for #714 but needs also #725 to be entirely visually correct.

### DIFF
--- a/sixtyfps_compiler/passes.rs
+++ b/sixtyfps_compiler/passes.rs
@@ -105,15 +105,18 @@ pub async fn run_passes(
         lower_popups::lower_popups(component, &doc.local_registry, diag);
         lower_layout::lower_layouts(component, &mut type_loader, diag).await;
         z_order::reorder_by_z_order(component, diag);
-        lower_shadows::lower_shadow_properties(component, &doc.local_registry, diag);
-        clip::handle_clip(component, &global_type_registry.borrow(), diag);
         transform_and_opacity::handle_transform_and_opacity(
             component,
             &global_type_registry.borrow(),
             diag,
         );
-        default_geometry::default_geometry(component, diag);
+        lower_shadows::lower_shadow_properties(component, &doc.local_registry, diag);
+        clip::handle_clip(component, &global_type_registry.borrow(), diag);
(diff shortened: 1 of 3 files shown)


sixtyfps_compiler/passes.rs
```diff
@@ -105,14 +105,14 @@ pub async fn run_passes(
...
         lower_layout::lower_layouts(component, &mut type_loader, diag).await;
         z_order::reorder_by_z_order(component, diag);
-        lower_shadows::lower_shadow_properties(component, &doc.local_registry, diag);
-        clip::handle_clip(component, &global_type_registry.borrow(), diag);
         transform_and_opacity::handle_transform_and_opacity(
             component,
             &global_type_registry.borrow(),
             diag,
         );
         default_geometry::default_geometry(component, diag);
```
ogoffart, line 113: I hope that the default_geometry is not needed to apply the geometry to these elements.

ogoffart (approved):

sixtyfps_compiler/passes.rs
tronical: Good question. I checked this now:

 * The clip pass explicitly installs bindings for the `width` and `height` properties of the newly created `Clip` element (that sits between the `Rectangle` and its children)
 * The shadows pass installs bindings for `x`/`y`/`width` and `height` on the newly created `BoxShadow` sibling that refer to the original element.
 * The situation where the `BoxShadow` becomes the new root of a repeated element moves the bindings to the geometry properties from the "old root` to the newly created `BoxShadow`. However the previous root does not have bindings for `width` and `height` anymore and relies on the default geometry pass. There's also a comment that underlines this:

>//  Finally we default geometry pass following this shadow lowering will apply a binding to
>//  the width and height of the inner to follow the size of the parent (the shadow).

This means that this patch would break the default geometry for Rectangles in repeaters that use a box-shadow and this patch needs fixing (code and comment).

ogoffart (approved):

PR #730, opened 2021-12-08
GL backend: allow to override scale factor with an env variable

ogoffart:
SIXTYFPS_SCALE_FACTOR override the scale factor at runtime


tronical (approved):
Should `WindowEvent::ScaleFactorChanged` be ignored if an override is set? (i.e. when moving windows across screens)

ogoffart:
> Should WindowEvent::ScaleFactorChanged be ignored if an override is set? (i.e. when moving windows across screens)

I was wondering the same.
Maybe it should indeed, what do you think?

tronical:
> > Should WindowEvent::ScaleFactorChanged be ignored if an override is set? (i.e. when moving windows across screens)
> 
> I was wondering the same. Maybe it should indeed, what do you think?

I'm in favour of the override resulting in the `WindowEvent::ScaleFactorChanged` being ignored

PR #731, opened 2021-12-09
Add support for additional non-printable keys

### LINKED ISSUE #170: Add support for additional non-printable keys

ogoffart:
Closes #170

 - centralize the list of keys in one file
 - add the ability to name the keys in the .60 files via `Keys`
 - As a test case, add support for up and down keys in the StandardListView

Note: 
 * I opted for a CamelCase key name in the Keys namespace because that's the same casing as what's used there: https://www.w3.org/TR/uievents-key/#named-key-attribute-values
Should we instead use the snake case: `Keys.left-arrow`  instead of `Keys.LeftArrow` ?

 * Thinking about it, the use of key code might be a problem if someone wants to print these text. Some replacement character will be used. We don't even have a is_printable API, or a way to say that a given code should not be used as text. (The code for the line edit hardcode the private range as an exclusion)

### DIFF
--- a/docs/builtin_elements.md
+++ b/docs/builtin_elements.md
@@ -458,6 +458,10 @@ Example := Window {
 
 The FocusScope exposes callback to intercept the pressed key when it has focus.
 
+The KeyEvent has a text property which is a character of the key entered.
+When a non-printable key is pressed, the character will be either a control character,
+or a private unicode character. The special characters are available in the `Keys` namespace
+
 ### Properties
 
 * **`has-focus`** (*bool*): Set to `true` when item is focused and receives keyboard events.
@@ -481,6 +485,9 @@ Example := Window {
             debug(event.text);
             if (event.modifiers.control) {
                 debug("control was pressed during this event");
+            }
+            if (event.text == Keys.Escape) {
+                debug("Esc key was pressed")
             }
             accept
         }
--- a/docs/builtin_elements.md
+++ b/docs/builtin_elements.md
@@ -460,7 +460,7 @@ The FocusScope exposes callback to intercept the pressed key when it has focus.
 
 The KeyEvent has a text property which is a character of the key entered.
 When a non-printable key is pressed, the character will be either a control character,
-or a private unicode character. The special characters are available in the `Keys` namespace
+or it will be mapped to a private unicode character. The mapping of these non-printable, special characters is available in the `Keys` namespace
 
 ### Properties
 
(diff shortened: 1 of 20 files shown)


tronical:
> * I opted for a CamelCase key name in the Keys namespace because that's the same casing as what's used there: https://www.w3.org/TR/uievents-key/#named-key-attribute-values
>   Should we instead use the snake case: `Keys.left-arrow`  instead of `Keys.LeftArrow` ?

That's a good question. I favour snake case for constants like that, but it's just a weird bias of mine :-)

> * Thinking about it, the use of key code might be a problem if someone wants to print these text. Some replacement character will be used. We don't even have a is_printable API, or a way to say that a given code should not be used as text. (The code for the line edit hardcode the private range as an exclusion)

Good point. I guess for that we would need API for Rust/C++/JS. Like

```
pub enum Key {
   F1, ...
}

pub fn key_from_string(it: &mut dyn Iterator<Item = char>) -> Option<Key> {
}
```

(The name is not a serious suggestion, of course - neither is exactly the signature)

I think that's out of scope of this patch though.

docs/builtin_elements.md
```diff
@@ -458,6 +458,10 @@ Example := Window {
 
 The FocusScope exposes callback to intercept the pressed key when it has focus.
 
+The KeyEvent has a text property which is a character of the key entered.
+When a non-printable key is pressed, the character will be either a control character,
+or a private unicode character. The special characters are available in the `Keys` namespace
```
tronical, line 463:
```suggestion
The KeyEvent has a text property which is a character of the key entered.
When a non-printable key is pressed, the character will be either a control character,
or it will be mapped to a private unicode character. The mapping of these non-printable, special characters is available in the `Keys` namespace
```

sixtyfps_runtime/corelib/items/text.rs
```diff
@@ -371,21 +370,23 @@ impl Item for TextInput {
...
                         Self::FIELD_OFFSETS.accepted.apply_pin(self).call(&());
                         return KeyEventResult::EventAccepted;
                     }
                 }
 
                 // Only insert/interpreter non-control character strings
                 if event.text.is_empty()
-                    || event.text.as_str().chars().any(|ch| ch.is_control() && ch != '\n')
+                    || event.text.as_str().chars().any(|ch| {
+                        (ch >= '\u{f700}' && ch <= '\u{f7ff}') || (ch.is_control() && ch != '\n')
```
tronical, line 388:
```suggestion
                        // exclude the private use area as we encode special keys into it
                        (ch >= '\u{f700}' && ch <= '\u{f7ff}') || (ch.is_control() && ch != '\n')
```

tronical (approved):

tronical (approved):

PR #740, opened 2021-12-11
Improve timer docs

tronical:
Use the term interval instead of duration to emphasize that it's not a single shot.

(The single-shot docs still use duration)


Be-ing (approved):

Be-ing:
thanks

ogoffart (approved):

PR #741, opened 2021-12-11
Timer: allow FnMut callbacks

Be-ing:


ogoffart (approved):

ogoffart:
Would be nice to have a test, but I realize the Timer doesn't seem to have tests currently 😕.

It was not really possible to test the timer before because it would have required to actually wait, but now that we mock the time, it should be possible to have proper tests using `sixtyfps_mock_elapsed_time`.
But anyway, this change is already good. Thank you!

Be-ing:
Thanks for the quick merge!

PR #742, opened 2021-12-12
wasm32: Fix reentrancy issues with invoke_from_event_loop

jturcotte:
If it's called on top of the an event processed on the event loop,
it might be that GLOBAL_PROXY is already borrowed further down
the stack.


tronical:
Cool! What still remains then is that the queued events aren't delivered, right?

PR #746, opened 2021-12-13
Add a simple frames per second counter to the GL and Qt backend

tronical:
It can be activated using `SIXTYFPS_DEBUG_PERFORMANCE=1`

### DIFF
--- a/sixtyfps_runtime/corelib/graphics/fps_counter.rs
+++ b/sixtyfps_runtime/corelib/graphics/fps_counter.rs
@@ -1,0 +1,137 @@
+/* LICENSE BEGIN
+    This file is part of the SixtyFPS Project -- https://sixtyfps.io
+    Copyright (c) 2021 Olivier Goffart <olivier.goffart@sixtyfps.io>
+    Copyright (c) 2021 Simon Hausmann <simon.hausmann@sixtyfps.io>
+
+    SPDX-License-Identifier: GPL-3.0-only
+    This file is also available under commercial licensing terms.
+    Please contact info@sixtyfps.io for more information.
+LICENSE END */
+/*!
+This module contains a simple helper type to measure the average number of frames rendered per second.
+*/
+
+use crate::timers::*;
+use std::cell::RefCell;
+use std::convert::TryFrom;
+use std::rc::Rc;
+
+enum RefreshMode {
+    Lazy,      // render only when necessary (default)
+    FullSpeed, // continuously render to the screen
+}
+
+impl<'a> TryFrom<&Vec<&'a str>> for RefreshMode {
+    type Error = ();
+
+    fn try_from(options: &Vec<&'a str>) -> Result<Self, Self::Error> {
+        if options.contains(&"refresh_lazy") {
+            Ok(Self::Lazy)
+        } else if options.contains(&"refresh_full_speed") {
+            Ok(Self::FullSpeed)
+        } else {
+            Err(())
+        }
+    }
+}
+
+/// Helper class that rendering backends can use to provide an FPS counter
+pub struct FPSCounter {
+    frame_times: RefCell<Vec<instant::Instant>>,
+    update_timer: Timer,
+    refresh_mode: RefreshMode,
+    output_console: bool,
+    output_overlay: bool,
+}
+
+impl FPSCounter {
+    /// Returns a new instance of the counter if requested by the user (via `SIXTYFPS_DEBUG_PERFORMANCE` environment variable).
+    /// The environment variable holds a comma separated list of options:
+    ///     * `refresh_lazy`: selects the lazy refresh mode, where measurements are only taken when a frame is rendered (due to user input or animations)
+    ///     * `refresh_full_speed`: frames are continuously rendered
+    ///     * `console`: the measurement is printed to the console
+    ///     * `overlay`: the measurement is drawn as overlay on top of the scene
+    pub fn new() -> Option<Rc<Self>> {
+        let options = match std::env::var("SIXTYFPS_DEBUG_PERFORMANCE") {
+            Ok(var) => var,
+            _ => return None,
+        };
+        let options: Vec<&str> = options.split(",").collect();
+
+        let refresh_mode = match RefreshMode::try_from(&options) {
+            Ok(mode) => mode,
+            Err(_) => {
+                eprintln!("Missing refresh mode in SIXTYFPS_DEBUG_PERFORMANCE. Please specify either refresh_full_speed or refresh_lazy");
+                return None;
+            }
+        };
+
+        let output_console = options.contains(&"console");
+        let output_overlay = options.contains(&"overlay");
+
+        if !output_console && !output_overlay {
+            eprintln!("Missing output mode in SIXTYFPS_DEBUG_PERFORMANCE. Please specify either console or overlay (or both)");
+            return None;
+        }
+
+        Some(Rc::new(Self {
+            frame_times: Default::default(),
+            update_timer: Default::default(),
+            refresh_mode,
+            output_console,
+            output_overlay,
+        }))
+    }
+
+    /// Call this function if you want to start measurements. This will also print out some system information such as whether
+    /// this is a debug or release build, as well as the provided winsys_info string.
+    pub fn start(self: &Rc<Self>, winsys_info: &str) {
+        #[cfg(debug_assertions)]
+        let build_config = "debug";
+        #[cfg(not(debug_assertions))]
+        let build_config = "release";
+
+        eprintln!("SixtyFPS: Build config: {}; Backend: {}", build_config, winsys_info);
+
+        let this = self.clone();
+        self.update_timer.stop();
+        self.update_timer.start(TimerMode::Repeated, std::time::Duration::from_secs(1), move || {
+            this.trim_frame_times();
+            if this.output_console {
+                eprintln!("average frames per second: {}", this.frame_times.borrow().len());
+            }
+        })
+    }
+
+    fn trim_frame_times(self: &Rc<Self>) {
+        let mut i = 0;
+        let mut frame_times = self.frame_times.borrow_mut();
+        while i < frame_times.len() {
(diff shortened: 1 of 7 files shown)


sixtyfps_runtime/rendering_backends/gl/glwindow.rs
```diff
@@ -42,6 +42,8 @@ pub struct GLWindow {
     // This cache only contains textures. The cache for decoded CPU side images is in crate::IMAGE_CACHE.
     pub(crate) texture_cache: RefCell<TextureCache>,
 
+    fps_counter: Option<Rc<FPSCounter>>,
```
ogoffart, line 45: does it really belong in the window?  or should it be in the backend?

sixtyfps_runtime/corelib/graphics/fps_counter.rs
```diff
@@ -0,0 +1,70 @@
...
+        let mut frame_times = self.frame_times.borrow_mut();
+        while i < frame_times.len() {
+            if frame_times[i].elapsed() > std::time::Duration::from_secs(1) {
+                frame_times.remove(i);
+            } else {
+                i += 1
+            }
+        }
+
+        eprintln!("average frames per second: {}", frame_times.len());
```
tronical, line 68: I'd like to use `debug_log!` here, but I can't seem to figure out how.

Just using `debug_log!` I get this:

```
error: cannot find macro `debug_log` in this scope
  --> sixtyfps_runtime/corelib/graphics/fps_counter.rs:68:9
   |
68 |         debug_log!("average frames per second: {}", frame_times.len());
   |         ^^^^^^^^^
   |
   = note: consider importing this macro:
           crate::debug_log
   = help: have you added the `#[macro_use]` on the module/import?
```

If I follow the "advice" and write `use crate::debug_log;` in the line before, I get this:

```
error: cannot determine resolution for the import
  --> sixtyfps_runtime/corelib/graphics/fps_counter.rs:68:13
   |
68 |         use crate::debug_log;
   |             ^^^^^^^^^^^^^^^^
   |
   = note: import resolution is stuck, try simplifying other imports

error: cannot determine resolution for the macro `debug_log`
  --> sixtyfps_runtime/corelib/graphics/fps_counter.rs:69:9
   |
69 |         debug_log!("average frames per second: {}", frame_times.len());
   |         ^^^^^^^^^
   |
   = note: import resolution is stuck, try simplifying macro imports

error: macro-expanded `macro_export` macros from the current crate cannot be referred to by absolute paths
   --> sixtyfps_runtime/corelib/graphics/fps_counter.rs:68:13
    |
68  |         use crate::debug_log;
    |             ^^^^^^^^^^^^^^^^
    |
    = note: `#[deny(macro_expanded_macro_exports_accessed_by_absolute_paths)]` on by default
    = warning: this was previously accepted by the compiler but is being phased out; it will become a hard error in a future release!
    = note: for more information, see issue #52234 <https://github.com/rust-lang/rust/issues/52234>
note: the macro is defined here
   --> sixtyfps_runtime/corelib/tests.rs:111:9
    |
111 | /         macro_rules! debug_log {
112 | |             ($($t:tt)*) => (eprintln!($($t)*))
113 | |         }
    | |_________^

error: could not compile `sixtyfps-corelib` due to 3 previous errors
```

So instead of `crate` if I use `use super::super::tests::debug_log;` I get:

```
error[E0432]: unresolved import `super::super::tests::debug_log`
  --> sixtyfps_runtime/corelib/graphics/fps_counter.rs:68:13
   |
68 |         use super::super::tests::debug_log;
   |             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ no `debug_log` in `tests`
   |
   = note: this could be because a macro annotated with `#[macro_export]` will be exported at the root of the crate instead of the module where it is defined
help: a macro with this name exists at the root of the crate
   |
68 |         use ::debug_log;
   |             ~~~~~~~~~~~

error: cannot determine resolution for the macro `debug_log`
  --> sixtyfps_runtime/corelib/graphics/fps_counter.rs:69:9
   |
69 |         debug_log!("average frames per second: {}", frame_times.len());
   |         ^^^^^^^^^
   |
   = note: import resolution is stuck, try simplifying macro imports
```

And if I then use `use ::debug_log;` I get this:

```
error[E0432]: unresolved import `debug_log`
  --> sixtyfps_runtime/corelib/graphics/fps_counter.rs:68:13
   |
68 |         use ::debug_log;
   |             ^^^^^^^^^^^ no external crate `debug_log`

error: cannot determine resolution for the macro `debug_log`
  --> sixtyfps_runtime/corelib/graphics/fps_counter.rs:69:9
   |
69 |         debug_log!("average frames per second: {}", frame_times.len());
   |         ^^^^^^^^^
   |
   = note: import resolution is stuck, try simplifying macro imports
```

Finally `use super::super::debug_log;` also doesn't work.

Do you have any idea what I may be missing here?

tronical:
Does this approach look reasonable to you? What performance numbers do you get with this? I'm wondering if there's any error in the logic.

sixtyfps_runtime/corelib/graphics/fps_counter.rs
```diff
@@ -0,0 +1,70 @@
...
+    /// Call this function if you want to start measurements. This will also print out some system information such as whether
+    /// this is a debug or release build, as well as the provided winsys_info string.
+    pub fn start(self: &Rc<Self>, winsys_info: &str) {
+        #[cfg(debug_assertions)]
+        let build_config = "debug";
+        #[cfg(not(debug_assertions))]
+        let build_config = "release";
+
+        eprintln!(
+            "sixtyfps: starting fps measurements. build config: {}; backend/windowing system: {}",
```
ogoffart, line 41:
```suggestion
            "SixtyFPS: Build config: {}; Backend: {}",
```

I like it shorter and to the point

ogoffart (commented):
So when there is no animation running, this will not print anything, or only in mouse input. 
And then the frame per second will be really low, because we don't draw frames.

Say you have a blinking cursor that blinks every second, then you've got one frame per second only, which people would think it is slow

sixtyfps_runtime/corelib/graphics/fps_counter.rs
ogoffart: Macro needs to be declared before their used.  So make  sure that you declare the module that has this macro before the one that uses it.

tronical: Ahh, thank you! that works

sixtyfps_runtime/corelib/graphics/fps_counter.rs
tronical: 👍

sixtyfps_runtime/corelib/graphics/fps_counter.rs
tronical: Ah no, that also doesn't help. I thought it did, but my build was a "red herring" :)

sixtyfps_runtime/rendering_backends/gl/glwindow.rs
tronical: In principle the data is per-window - different windows might render at different frame rates.

Eventually I'd also rather print this as overlay text instead of going via stdout.

Without this patch, try this on your Linux system:

  * set the `GALLIUM_HUD=fps` environment variable
  * Run any SixtyFPS app with `SIXTYFPS_BACKEND=GL`
  * Watch the fps counter and graph :-)

I think this would be nice to have (even without the graph, just text) and then it's per-window. That could be for example `SIXTYFPS_DEBUG_PERFORMANCE=overlay`.

tronical:
> So when there is no animation running, this will not print anything, or only in mouse input. And then the frame per second will be really low, because we don't draw frames.
> 
> Say you have a blinking cursor that blinks every second, then you've got one frame per second only, which people would think it is slow

The timer fires every second - regardless of whether an animation is running or there is user input - and prints the FPS.

It's true that for a "blinking cursor" shows a low frame rate, but at the same time I think that data is also useful: It shows that we don't render unnecessarily (and waste cpu cycles). I'd be concerned if a blinking cursor renders at 60 fps :)

The gallium hud works the same way.

Another option would be to say that if the fps counter is enabled, we never put the event loop to sleep and ask it to continuously produce frames. That however skews cpu measurements.

Maybe that should be basically a "parameter" to `SIXTYFPS_DEBUG_PERFORMANCE` ?

tronical:
So the latest patch set implements the two different measurement modes and an overlay mode. If this is okay, then I can also document this. I wonder where the best place would be....

sixtyfps_runtime/corelib/graphics/fps_counter.rs
```diff
@@ -0,0 +1,147 @@
...
+                height: crate::properties::Property::new(100.),
+                text: crate::properties::Property::new(
+                    format!("FPS: {}", self.frame_times.borrow().len()).into(),
+                ),
+                color: crate::properties::Property::new(
+                    crate::Color::from_rgb_u8(0, 128, 128).into(),
+                ),
+                ..Default::default()
+            };
+            let label = unsafe { core::pin::Pin::new_unchecked(&label) };
```
tronical, line 143: TODO: use stack pin macro

tronical:
This version is now without `unsafe` and instead adds a helper method to the `ItemRenderer` to draw the text for the overlay.

ogoffart:
Should it be documented somwehre? 
(we should perhaps document all the env variable like the SIXTYFPS_BACKEND or SIXTYFPS_STYLE and others)

tronical:
Yes, that was another thing I was wondering about, too. We could start a "Debugging Techniques" section in markdown, try to keep it language independent (with notes about specifics though) and link it from both. I can start that in this MR, if that sounds good to you?

ogoffart (approved):

tronical:
Landing this, will create a separate MR for the docs. (Changed `Fixes` into `cc` in the commit, to avoid closing the ticket)

PR #760, opened 2021-12-21
reuse_xtask: Use xtask to make license headers more SPDX compliant

hunger:
Long term I would like to use xtask to verify that we have the right licenses in places and the reuse tool for SPDX compliance.

This PR just does a bulk-update of the existing license headers, assuming that those are OK:-) It does get the easy bulk of the work done and updates the xtask to accept the new copyright headers.

Relevant changes are to make the copyright header shorter and to properly document the license as GPL or commercial in a proper SDPX license expression.

Xtask will no longer check enforce the copyright holders to be Olivier and Simon, and is OK with any: Our CLA says contributors keep their copyright, so they need to be listed there.

Note that the `--fix-it` option of `check_license_header` *WILL* change the copyright holder at this time. I am not sure we need to change this?


tronical:
> Note that the --fix-it option of check_license_header WILL change the copyright holder at this time. I am not sure we need to change this?

Indeed, I don't think that will be needed.

tronical:
+1

ogoffart:
+1

tronical:
The following two files slipped through the xtask and still have the old copyright and SPDX tags:

```
Cross.toml
tools/lsp/sixtyfps.ksyntaxhighlighter.xml
tools/lsp/sublime/SixtyFPS.tmPreferences
```

tronical:
> The following two files slipped through the xtask and still have the old copyright and SPDX tags:
> 
> ```
> Cross.toml
> tools/lsp/sixtyfps.ksyntaxhighlighter.xml
> tools/lsp/sublime/SixtyFPS.tmPreferences
> ```

To be done in a follow-up change

tronical (approved):

PR #763, opened 2021-12-22
Work towards REUSE compatibility

hunger:
Do some simple mass license header changes, clean up the remaining occurances of the old license header, clean up some potential license issue and make the xtask license checker ignore files we are going to add later.


tronical:
I think Github also would like to have a `LICENSE.md` in the root directory.

ogoffart:
Having license headers on .toml files is over the top is it not? Did you not say we can have a separate file describing the license of most files?

ogoffart:
another thing which we currently do wrong is that each crates should have a license in the crate package.

hunger:
@ogoffart Well, you do have copyright headers in the cmake files, conceptually that is not different from the Cargo.toml files. I also expect some of these to have a copyright other than the "default".

Basically you have the following options to specify copyright information:

 * in the file itself.
 * in a separate file with the same name as the file the declaration applies to with ".license" appended.
 * in .reuse/dep5 file at the directory you run the reuse tool in. Dep5 files anywhere else are going to be ignored.

Reuse also requires license information for *all* files.

Given all that, I greatly prefer having the copyright information inside the file itself if possible: the .license-files litter a lot and the dep5 file is nice, but if we want to support running the reuse tool at a crate level as well as from the top-level, then we need to duplicate the copyright information in the crate level and the top level dep5 file.

This patchset only has changes that work both for a top level reuse run as well as a from a crate level. So basically no crate level LICENSES folders and no dep5 files:-)

I did use .license files mostly in places where that file replaces some other file with license information. So no extra litter:-)

On top of this PR I do have crate-level LICENSE folders and .reuse/dep5 files. I am currently checking how I can use that informtion to enable top-level runs of the reuse tool. Maybe a script that creates the top-level dep5 files from its children? Or maybe there is a way to tell reuse about "child projects" or something? I wanted to check in on the FSFe peopke at RC3 this week. Maybe they have an idea?

I am open to use other tools as well, but the reuse one seemed to be the one with the best infrastructure support (e.g. GH actions;-), so I went

PR #771, opened 2021-12-24
docs: fix Kate LSP JSON, add Windows support

Be-ing:
Copying & pasting the JSON should just work now without the user
needing to edit the hardcoded path.

### DIFF
--- a/tools/lsp/README.md
+++ b/tools/lsp/README.md
@@ -70,14 +70,16 @@ cargo install sixtyfps-lsp
 
 Then go to *Settings > Configure Kate*. In the *Plugins* section, enable the *LSP-Client* plugin.
 This will add a *LSP Client* section in the settings dialog. In that *LSP Client* section,
-go to the *User Server Settings*, and  enter the following in the text area: (**Replace YOUR_USER by your username**)
+go to the *User Server Settings*, and  enter the following in the text area:
 
 ```json
 {
   "servers": {
-	"SixtyFPS": {
-	  "command": ["/home/YOUR_USER/.cargo/bin/sixtyfps-lsp"],
-	  "highlightingModeRegex": "SixtyFPS"	}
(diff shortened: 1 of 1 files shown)


tools/lsp/README.md
```diff
@@ -75,9 +75,11 @@ go to the *User Server Settings*, and  enter the following in the text area: (**
 ```json
 {
   "servers": {
-	"SixtyFPS": {
-	  "command": ["/home/YOUR_USER/.cargo/bin/sixtyfps-lsp"],
-	  "highlightingModeRegex": "SixtyFPS"	}
+    "SixtyFPS": {
+      "path": ["${ENV:HOME}/.cargo/bin", "%{ENV:USERPROFILE}/.cargo/bin"],
```
tronical, line 79: I can see %{ENV:..} in the kate docs but not ${ENV:..}. Are you sure it’s correct to use both?

Be-ing: I think so. I took this from [Kate's configuration for rust-analyzer](https://invent.kde.org/utilities/kate/-/blob/33a748a44d90299210703f0f0cbce88e14cb8a5c/addons/lspclient/settings.json#L104).

Be-ing: Oh, that uses `%` for both... hmm...

Be-ing: I asked in the Kate Matrix room and `%` is correct.

tronical:
Thanks! Looks good. Can you also remove the `(**Replace YOUR_USER by your username**)` part in the paragraph before this block?

Alternatively, I can also do it afterwards.

Be-ing:
> Can you also remove the (**Replace YOUR_USER by your username**) part in the paragraph before this block?

Good catch, fixed.

tronical (approved):

PR #785, opened 2022-01-02
Some polish to the Rust crate docs

tronical:
* Improve headings and linkage
* Replace "you can use XXX to YYY" with "Use XXX to YYY"
* In the limitation about exported components, link to the tracking
issue.


Vollbrecht (approved):
Thank you, four your continuous work on a better documentation. On my first contact with sixtyfps in the memory game i was a bit confused because it uses the inline version. while in every other place the external files would be used. This should make it clearer for other newcomers.

ogoffart (approved):

PR #792, opened 2022-01-03
Fix wrong count of sub components in the item tree

### LINKED ISSUE #781: Weird behavior for `StandardButton` with the OpenGL backend
Trying to compile this example seems to show two instances of the cancel button instead of two separate buttons:

```javascript
import { StandardButton } from "sixtyfps_widgets.60";

App := Window {
    HorizontalLayout {
        StandardButton { 
            kind: cancel;
        }
        StandardButton { 
            kind: ok;
        }
    }
}
```

image

- Hovering or clicking the right button will make the left button react.
- This is not reproducible with the Qt backend
- This is not reproducible with sixtyfps-viewer
- This is not reproducible when using normal `Button`s instead of `StandardButton`

I can't make much sense of the generated code, but it seems to me that there is only one StandardButton instantiated in the InnerApp there instead of two hen using Button.

ogoffart:
Fixes #781

### DIFF
--- a/sixtyfps_compiler/generator.rs
+++ b/sixtyfps_compiler/generator.rs
@@ -171,12 +171,10 @@ pub fn build_item_tree<T: ItemTreeBuilder>(
     // sub-component children, needed to allocate the correct amount of
     // index spaces for sub-components.
     fn item_sub_tree_size(e: &ElementRc) -> usize {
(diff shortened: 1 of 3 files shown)


tronical (approved):
Ouch!

PR #794, opened 2022-01-03
Merge `Rc::ptr_eq` on traits use into a SharedModel

hunger:
It is now safe to replace the manual implementation with the derived
implementation, so do that.

This is a continuation of #311, which I accidentally closed when I deleted the branch.

### DIFF
--- a/api/sixtyfps-rs/migration.md
+++ b/api/sixtyfps-rs/migration.md
@@ -62,3 +62,6 @@ fn model_tracker(&self) -> &dyn ModelTracker {
 }
 ```
 
+#### Minor changes to `Model`
+
+`ModelRc` replaces `ModelHandle`.
--- a/api/sixtyfps-rs/migration.md
+++ b/api/sixtyfps-rs/migration.md
@@ -64,4 +64,4 @@ fn model_tracker(&self) -> &dyn ModelTracker {
 
 #### Minor changes to `Model`
 
-`ModelRc` replaces `ModelHandle`.
+`ModelHandle` was renamed [`ModelRc`].
--- a/sixtyfps_compiler/generator/rust.rs
+++ b/sixtyfps_compiler/generator/rust.rs
@@ -78,7 +78,7 @@ fn rust_type(ty: &Type) -> Option<proc_macro2::TokenStream> {
         Type::Struct { name: Some(name), .. } => Some(struct_name_to_tokens(name)),
         Type::Array(o) => {
             let inner = rust_type(o)?;
-            Some(quote!(sixtyfps::re_exports::ModelHandle<#inner>))
+            Some(quote!(sixtyfps::re_exports::ModelRc<#inner>))
         }
         Type::Enumeration(e) => {
             let e = ident(&e.name);
@@ -536,7 +536,7 @@ fn generate_sub_component(
 
         let mut model = compile_expression(&repeated.model, &ctx);
         if repeated.model.ty(&ctx) == Type::Bool {
-            model = quote!(sixtyfps::re_exports::ModelHandle::new(sixtyfps::re_exports::Rc::<bool>::new(#model)))
+            model = quote!(sixtyfps::re_exports::ModelRc::new(sixtyfps::re_exports::Rc::<bool>::new(#model)))
         }
 
         init.push(quote! {
@@ -1273,7 +1273,7 @@ fn compile_expression(expr: &Expression, ctx: &EvaluationContext) -> TokenStream
                     ))
                 }
                 (Type::Float32, Type::Model) | (Type::Int32, Type::Model) => {
-                    quote!(sixtyfps::re_exports::ModelHandle::new(sixtyfps::re_exports::Rc::<usize>::new(#f as usize)))
+                    quote!(sixtyfps::re_exports::ModelRc::new(sixtyfps::re_exports::Rc::<usize>::new(#f as usize)))
                 }
                 (Type::Float32, Type::Color) => {
                     quote!(sixtyfps::re_exports::Color::from_argb_encoded(#f as u32))
@@ -1446,7 +1446,7 @@ fn compile_expression(expr: &Expression, ctx: &EvaluationContext) -> TokenStream
             let base_e = compile_expression(array, ctx);
             let index_e = compile_expression(index, ctx);
             let value_e = compile_expression(value, ctx);
-            quote!((#base_e).set_row_data(#index_e as usize, #value_e as _);)
+            quote!((#base_e).set_row_data(#index_e as usize, #value_e as _))
         }
         Expression::BinaryExpression { lhs, rhs, op } => {
             let (conv1, conv2) = match crate::expression_tree::operator_class(*op) {
@@ -1536,7 +1536,7 @@ fn compile_expression(expr: &Expression, ctx: &EvaluationContext) -> TokenStream
             let val = values.iter().map(|e| compile_expression(e, ctx));
             if *as_model {
                 let rust_element_ty = rust_type(element_ty).unwrap();
-                quote!(sixtyfps::re_exports::ModelHandle::new(
+                quote!(sixtyfps::re_exports::ModelRc::new(
                     sixtyfps::re_exports::Rc::new(sixtyfps::re_exports::VecModel::<#rust_element_ty>::from(
                         sixtyfps::re_exports::vec![#(#val as _),*]
                     ))
--- a/sixtyfps_runtime/corelib/model.rs
+++ b/sixtyfps_runtime/corelib/model.rs
@@ -464,7 +464,7 @@ impl Model for bool {
 
 /// Properties of type Array in the .60 language are represented as
 /// a `ModelRc` that implements the `Model` trait.
-#[derive(derive_more::Deref, derive_more::DerefMut, derive_more::From, derive_more::Into)]
+#[derive(derive_more::Deref, derive_more::DerefMut)]
 pub struct ModelRc<T>(Option<Rc<dyn Model<Data = T>>>);
 
 impl<T> core::fmt::Debug for ModelRc<T> {
--- a/sixtyfps_runtime/corelib/model.rs
+++ b/sixtyfps_runtime/corelib/model.rs
@@ -463,7 +463,7 @@ impl Model for bool {
 }
 
 /// Properties of type Array in the .60 language are represented as
-/// a `ModelRc` that implements the `Model` trait.
+/// a `ModelRc` that implements the [`Model`] trait.
 #[derive(derive_more::Deref, derive_more::DerefMut)]
 pub struct ModelRc<T>(Option<Rc<dyn Model<Data = T>>>);
 
--- a/sixtyfps_runtime/corelib/model.rs
+++ b/sixtyfps_runtime/corelib/model.rs
@@ -239,14 +239,14 @@ pub trait Model {
 
     /// Return something that can be downcast'ed (typically self)
     ///
-    /// This is useful to get back to the actual model from a ModelHandle stored
+    /// This is useful to get back to the actual model from a `ModelRc` stored
     /// in a component.
     ///
     /// ```
     /// # use sixtyfps_corelib::model::*;
     /// # use std::rc::Rc;
     /// let vec_model = Rc::new(VecModel::from(vec![1i32, 2, 3]));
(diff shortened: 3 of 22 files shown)


ogoffart:
Is the title of this PR still accurate?

So the main thing this does, if i read correctly, is that it change `ModelHandle<T>` to be a `Option<SharedModel<T>>`  and that's a new class.
I don't remember the rational for changing the API there, but i don't understand why we need both ModelHandle and SharedModel.  SharedModel itself could wrap an `Option<Rc<dyn Model>>`  I believe we said we don't want to have the internal pub anyway. And so there should only be one of these class in the public API

hunger:
The original reason to have this was to deduplicate the `Rc::ptr_eq` that we used to have and that clippy is not too happy abouti with trait object in the picture.

This puts all the code into one place that does `Rc::ptr_eq` on traits.

ogoffart:
So what we said in the call is that we don't want to have `Option<SharedModel>`, but instead, we want `SharedModel` that internally contains an `Option<Rc<dyn Model>` but without being public.

I think we want conversion functions like so:
Currently, ModelHandle has a conversion `From<Option<Rc<...>>>` , but i think it should just be `From<Rc<...>>`.
We can also have a `TryInto<Rc<...>>`

Currently, `ModelHandle::new` Takes a Rc,  but i think it would be better if  it was `fn new(model: impl Model<T>)`
or maybe it could be `fn new(model: impl Into<Rc<dyn Model<Data = T>>)` 
The rational is that i want to be able to write code like this:
```rust
ModelHandle::new(VecModel::from_slice(&[...]))
```

Maybe we also want `impl <T, F: Model<Data= T>> From<F>  for SharedModel<T>`

tronical:
> Maybe we also want impl <T, F: Model<Data= T>> From<F> for SharedModel<T>

I'd like that, too. On the one hand it's good to make the `Rc` allocation/wrapping visible to the user, on the other hand `SharedModel` implies the wrapping in a way and I've always found the `as Rc<dyn Model<Data = ...>>` part annoying to write. So 👍 for the above if that adds the convenience. I think it'll show pretty quickly in our own examples and tests whether it helps or not.

ogoffart:
Regarding the `Shared` Terminology:
When we have, say:
```rust
let a = SharedString::new("hello");
let mut b = a.clone();
b.push_str("world");
// a didn't change
assert_eq!(a, "hello");
```

But that's not the case with the models:
```
let a = SharedModel::from(...);
let b = a.clone();
b.set_row_data(1, "foobar".into());
// a **is** changed
assert_eq!(a.row_data(1).unwrap(), "foobar".into());
```

Which is why the `Shared` terminology maybe isn't optimal.

Maybe `ModelRc<T>` would actually be a better name?

But if you think it's not a problem, I don't oppose `ModelShared<T>`

api/sixtyfps-rs/migration.md
```diff
@@ -56,6 +56,22 @@ fn model_tracker(&self) -> &dyn ModelTracker {
 }
 ```
 
+#### `Option<SharedModel>` replaces `ModelHandle`
+
+`Option<SharedModel>` replaces `ModelHandle`.
+
+Old code:
+
+```rust
```
ogoffart, line 65: I don't think you need to always put code sample for such trivial things like that.
Anyway, the ```rust must compile as it is going to be a doctest, so you need to add  `ignore` 
But i realize that this file is not included in the rust doc yet.

api/sixtyfps-rs/migration.md
```diff
@@ -56,6 +56,22 @@ fn model_tracker(&self) -> &dyn ModelTracker {
 }
 ```
 
+#### `Option<SharedModel>` replaces `ModelHandle`
```
tronical, line 59: This doesn't seem quite right, because in < 0.2.0 there's no `Option<SharedModel>`. I think perhaps you meant "ModelHandle was renamed to SharedModel"?

In that case it's a simple search and replace, if we don't change the API in a way that breaks existing code, just add new convenience (like the `From` from the model without Rc).

ogoffart: > if we don't change the API in a way that breaks existing code

We will if we remove the Deref to the Option<Rc> and the pub access to .0, which we should.

sixtyfps_compiler/generator/rust.rs
```diff
@@ -1398,11 +1398,12 @@ fn compile_expression(expr: &Expression, ctx: &EvaluationContext) -> TokenStream
             debug_assert!(matches!(array.ty(ctx), Type::Array(_)));
             let base_e = compile_expression(array, ctx);
             let index_e = compile_expression(index, ctx);
-            quote!(match &#base_e { x => {
+            quote!({
+                let x = &(#base_e);
```
ogoffart, line 1402: that's not the same.  We want to make sure that the temporaries in #base_e are not destroyed.

sixtyfps_compiler/generator/rust.rs
```diff
@@ -1763,10 +1764,11 @@ fn compile_builtin_function_call(
         }
         BuiltinFunction::ImageSize => quote!( #(#a)*.size()),
         BuiltinFunction::ArrayLength => {
-            quote!(match &#(#a)* { x => {
+            quote!({
+                let x = &#(#a)*;
```
ogoffart, line 1768: ditto

sixtyfps_runtime/corelib/model.rs
```diff
@@ -694,45 +688,40 @@ impl<C: RepeatedComponent> PinnedDrop for Repeater<C> {
...
-    fn model(self: Pin<&Self>) -> ModelHandle<C::Data> {
-        // Safety: Repeater does not implement drop and never let access model as mutable
+    fn model(self: Pin<&Self>) -> SharedModel<C::Data> {
+        // Safety: Repeater does not implement drop and never allows access to model as mutable
         let model = self.project_ref().model;
 
         if model.is_dirty() {
             *self.inner.borrow_mut() = RepeaterInner::default();
             self.is_dirty.set(true);
-            if let ModelHandle(Some(m)) = model.get() {
```
ogoffart, line 704: Too bad that we loose this optimisation that we won't allocate stuff if the model is empty.
I guess that's fine...
Otherwise you could use pub(crate)  to allow us to access the inner one.
Or we make TryInto Rc<> work

sixtyfps_runtime/corelib/model.rs
```diff
@@ -769,11 +758,7 @@ impl<C: RepeatedComponent + 'static> Repeater<C> {
...
 
-        let model = if let ModelHandle(Some(model)) = self.model() {
-            model
-        } else {
-            return empty_model();
-        };
+        let model = self.model();
         let row_count = model.row_count();
         if row_count == 0 {
             return empty_model();
```
ogoffart, line 764: empty_model can now be inlined there, if i'm not mistaken.

sixtyfps_runtime/corelib/model.rs
```diff
@@ -462,56 +462,50 @@ impl Model for bool {
     }
 }
 
-/// Properties of type array in the .60 language are represented as
-/// an [`Option`] of an [`Rc`] of something implemented the [`Model`] trait
+/// Properties of type Array in the .60 language are represented as
+/// a `SharedModel` that implements the `Model` trait.
+#[repr(transparent)]
 #[derive(derive_more::Deref, derive_more::DerefMut, derive_more::From, derive_more::Into)]
```
ogoffart, line 468: This needs to go away as well, but can be done in another commit

sixtyfps_runtime/corelib/model.rs
```diff
@@ -462,56 +462,50 @@ impl Model for bool {
...
-pub struct ModelHandle<T>(pub Option<Rc<dyn Model<Data = T>>>);
+pub struct SharedModel<T>(Option<Rc<dyn Model<Data = T>>>);
 
-impl<T> core::fmt::Debug for ModelHandle<T> {
-    fn fmt(&self, f: &mut core::fmt::Formatter<'_>) -> core::fmt::Result {
-        write!(f, "ModelHandle({:?})", self.0.as_ref().map(|_| "dyn Model"))
+impl<T> std::fmt::Debug for SharedModel<T> {
```
ogoffart, line 471:
```suggestion
impl<T> core::fmt::Debug for SharedModel<T> {
```

We want to compile without std on MCU, remember

ogoffart (changes requested):

ogoffart:
Maybe we can have a ` #[deprectated] type ModelHandle<T> = SharedModel<T>;`  to ease porting?

sixtyfps_compiler/generator/rust.rs
hunger: I do not get what you are going at, sorry.

hunger:
Does the deprecation make sense in the transition from 0.1 to 0.2?

sixtyfps_compiler/generator/rust.rs
ogoffart:
```rust
let x = &String::new("42").trim();   // doesn't work (take address of temporary)
println!("{}", x); 
```
```rust
match (&String::new("42").trim()) {
  // works
  x => { println!("{}", x); }
```

So please undo that.

Why are you changing it?

ogoffart: I guess something like this would fail to compile in rust:

```
property <string> world: (["hello", "world"])[1];
```

hunger: Wow, tricky, but obvious now that you pointed it out!

sixtyfps_runtime/corelib/model.rs
hunger: Will an empty model ever get dirty?

ogoffart: it can grow, yes.

ogoffart: I mean, what's an empty model.  a model with row_count = 0  but an actual model can grow.
But a model backed by None cannot change unless the property itself is changed. So no, a model with None cannot become dirty

hunger: So we do not loose the optimization: `None` will never get dirty, so it will never end up here in the first place.

api/sixtyfps-rs/migration.md
```diff
@@ -62,3 +62,6 @@ fn model_tracker(&self) -> &dyn ModelTracker {
 }
 ```
 
+#### Minor changes to `Model`
+
+`ModelRc` replaces `ModelHandle`.
```
ogoffart, line 67:
```suggestion
`ModelHandle` was renamed [`ModelRc`].
```

sixtyfps_runtime/corelib/model.rs
```diff
@@ -462,56 +462,49 @@ impl Model for bool {
     }
 }
 
-/// Properties of type array in the .60 language are represented as
-/// an [`Option`] of an [`Rc`] of something implemented the [`Model`] trait
-#[derive(derive_more::Deref, derive_more::DerefMut, derive_more::From, derive_more::Into)]
-pub struct ModelHandle<T>(pub Option<Rc<dyn Model<Data = T>>>);
+/// Properties of type Array in the .60 language are represented as
+/// a `ModelRc` that implements the `Model` trait.
```
ogoffart, line 466:
```suggestion
/// a `ModelRc` that implements the [`Model`] trait.
```

sixtyfps_runtime/corelib/model.rs
ogoffart: `model` in this case is a `Property<ModelRc>`  It's the property that gets dirty here, not the model.
But i guess we can live with that for now.

ogoffart (approved):

api/sixtyfps-rs/migration.md
tronical: renamed *to* :-)

tronical (approved):
I really like the name btw. This is a nice improvement of the API.

sixtyfps_runtime/corelib/model.rs
```diff
@@ -462,56 +462,49 @@ impl Model for bool {
...
+#[derive(derive_more::Deref, derive_more::DerefMut)]
```
ogoffart, line 467: We still need to get rid of Deref and DerefMut, but we can do that in another PR

Issue #754, opened 2021-12-20
iterating with `for` does not compile within `Path`


Be-ing:
Minimal example:
```
Waveform := Path {
  height: 100px;
  width: 100px;

  MoveTo {
    x: 0;
    y: 0;
  }
  for sample[i] in [ 0, 1, 2, 3 ] : LineTo {
    x: i;
    y: sample;
  }
  Close {}
}
```

The error occurs accessing either the key (`i`) or value (`sample`). Replacing both the key and the value with numeric literals successfully compiles.

Compile error:
```
moire on  main [$!?] is 📦 v0.1.0 via 🦀 v1.57.0 
❯ RUST_BACKTRACE=full mold -run cargo run -- ~/music/Emancipator\ -\ 2006\ -\ Soon\ It\ Will\ Be\ Cold\ Enough/02\ Emancipator\ -\ Soon\ It\ Will\ Be\ Cold\ Enough\ To\ Build\ Fires.flac
   Compiling moire v0.1.0 (/home/be/sw/moire)
error: failed to run custom build command for `moire v0.1.0 (/home/be/sw/moire)`

Caused by:
  process didn't exit successfully: `/home/be/sw/moire/target/debug/build/moire-add10bb873b8dc2f/build-script-build` (exit status: 101)
  --- stderr
  thread 'main' panicked at 'called `Option::unwrap()` on a `None` value', /home/be/.cargo/git/checkouts/sixtyfps-1a243e0ff9772996/75136e1/sixtyfps_compiler/object_tree.rs:1572:69
  stack backtrace:
     0:     0x5561b81425fc - <std::sys_common::backtrace::_print::DisplayBacktrace as core::fmt::Display>::fmt::h3264ad1edb4a78ca
     1:     0x5561b816536c - core::fmt::write::h85b247e856e081ad
     2:     0x5561b813e435 - std::io::Write::write_fmt::he3b80a66fc2ba5d3
     3:     0x5561b81441d0 - std::panicking::default_hook::{{closure}}::hf2bcdea8e51c0c13
     4:     0x5561b8143d85 - std::panicking::default_hook::h8864cfaaed08e53a
     5:     0x5561b81449f4 - std::panicking::rust_panic_with_hook::hb272975096b9f3ce
     6:     0x5561b8144472 - std::panicking::begin_panic_handler::{{closure}}::hbbd6c4e11858ba52
[...]

PR #799, opened 2022-01-04
Make it possible to use a string binding for Path commands

tronical:
This was requested a few days ago in the chat and might also present an alternative for #754


ogoffart:
This adds more dependency to the runtime library, which we would like to avoid, i think.
But if we really need this feature, the implementation looks good

tronical:
It would be possible to replace the direct dependency to lyon_svg with a dependency to svgtypes and call the parser directly. The former is a thin wrapper around the latter. The GL backend depends on it through usvg.

PR #800, opened 2022-01-04
More clippy fixes

hunger:
Based on earlier experience: These patches might not be uncontroversial, so I am putting them here for review:-)

1. The first one annotates some more functions with the #[must_use] directive. That is visual noise, but I find it rather worthwhile anyway.
2. The second one allows `clippy::transmute_num_to_bytes` in places where the `cpp` macro is used. That silences some warnings, but again at the price of some clutter.
3. The third removes some `return`s that `clippy` claims are redundant. I like that it removes the warning messages, but in some places I think the return has at least some aesthetic value.

Any opinions?

### DIFF
--- a/sixtyfps_runtime/interpreter/dynamic_component.rs
+++ b/sixtyfps_runtime/interpreter/dynamic_component.rs
@@ -836,7 +836,7 @@ pub(crate) fn generate_component<'id>(
             _children_offset: u32,
             _component_state: &Self::SubComponentState,
         ) -> Self::SubComponentState {
-            ()
+            /* nothing to do */
         }
 
         fn enter_component_children(
--- a/api/sixtyfps-rs/lib.rs
+++ b/api/sixtyfps-rs/lib.rs
@@ -530,6 +530,7 @@ pub trait ComponentHandle {
         Self: Sized;
 
     /// Returns a clone of this handle that's a strong reference.
+    #[must_use]
     fn clone_strong(&self) -> Self;
 
     /// Internal function used when upgrading a weak reference to a strong one.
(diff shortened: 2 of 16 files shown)


tronical:
No strong feeling about them, so +1 from me :-)

ogoffart:
> The second one allows clippy::transmute_num_to_bytes in places where the cpp macro is used. That silences some warnings, but again at the price of some clutter.

Maybe this should be added to the cpp macro instead.

ogoffart:
> The first one annotates some more functions with the #[must_use] directive. That is visual noise, but I find it rather worthwhile anyway.

This is a good change.  And since ti touches some public API, it should be mentioned in the ChangeLog

sixtyfps_runtime/interpreter/dynamic_component.rs
```diff
@@ -836,7 +836,7 @@ pub(crate) fn generate_component<'id>(
             _children_offset: u32,
             _component_state: &Self::SubComponentState,
         ) -> Self::SubComponentState {
-            ()
+            /* nothing to do */
```
ogoffart, line 839: I believe the `()`  is because `Self::SubComponentState` is a `()`

ogoffart (commented):
Yeah, this is good. But it would be better if the problem in the cpp crates were upstreamed in the cpp crate.

hunger:
Upstreaming the cpp changes is indeed a good idea!

I need to get into the mindset of "hey I can fix that upstream" again:-)

hunger:
I only left the first two patches in place that we agreed to have.

ogoffart:
I think we agreed to have 1.  And 3.

ogoffart:
The one about transmute should be fixed in the cop create

hunger:
That's why I pushed this for review: I am sure you said 1 and 2 on Friday, but I remembered that we had talked about doing 2 upstream, so I wanted to have this opportunity for clarification:-)

PR #801, opened 2022-01-05
Complete the C++ Timer API

tronical:
With this patch it matches the Rust API, with start(), restart(), running() and
a default constructor.

### DIFF
--- a/api/sixtyfps-cpp/include/sixtyfps.h
+++ b/api/sixtyfps-cpp/include/sixtyfps.h
@@ -337,23 +337,66 @@ private:
     private_api::WindowRc inner;
 };
 
+#if !defined(DOXYGEN)
+using cbindgen_private::TimerMode;
+#else
+/// The TimerMode specifies what should happen after the timer fired.
+///
+/// Used by the sixtyfps::Timer::start() function.
+enum class TimerMode {
+    /// A SingleShot timer is fired only once.
+    SingleShot,
+    /// A Repeated timer is fired repeatedly until it is stopped or dropped.
+    Repeated,
+};
+#endif
+
 /// A Timer that can call a callback at repeated interval
 ///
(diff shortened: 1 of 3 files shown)


sixtyfps_runtime/corelib/timers.rs
```diff
@@ -360,14 +361,19 @@ pub(crate) mod ffi {
     /// The timer MUST be stopped with sixtyfps_timer_stop
     #[no_mangle]
     pub extern "C" fn sixtyfps_timer_start(
+        id: i64,
```
ogoffart, line 364: can that be documented that -1 means create?

api/sixtyfps-cpp/include/sixtyfps.h
```diff
@@ -337,24 +337,67 @@ class Window
     private_api::WindowRc inner;
 };
 
+#if !defined(DOXYGEN)
```
ogoffart, line 340: This is quite unfortunate. I see that we have the same problem for other enums.
Too bad there is no equivalent to #[doc(inline)] for doxygen.
Can we think of another way?

ogoffart (approved):

sixtyfps_runtime/corelib/timers.rs
tronical: Sure, can do

api/sixtyfps-cpp/include/sixtyfps.h
tronical: Yeah, it's a bit of a shame. We're almost there since cbindgen extracts the comments as well and doxygen could process them (if generated in time).

Perhaps this is fixable by instructing cbindgen to emit the enum into the public namespace. I can give it a try (separate from this merge request, as this affects other enums as well).

PR #802, opened 2022-01-05
Improve robustness of unmapping windows with the GL backend

tronical:
Make sure to make the GL context is current before destroying the femtovg canvas.

The drop implementation of the canvas will call functions to delete GL resources such as textures, FBOs, etc.
and for that to work reliably the context must be current.


ogoffart (approved):
it's sad that we have to use an Option despite we know it's never None but i don't see another way.

tronical:
I agree. I was contemplating using the drop_move crate but that’s yet another dependency.

PR #803, opened 2022-01-05
Fix the Rust/C++ Timer API to be more convenient to use

tronical:
Allow calling restart() on a repeated timer if it has been previously stopped.


ogoffart (approved):
Looks alright.

A problem which was already before is that the tests uses real time.
We should mock the time like what we do in the property_animation.60 test and some other animation test

tronical:
> Looks alright.
> 
> A problem which was already before is that the tests uses real time. We should mock the time like what we do in the property_animation.60 test and some other animation test

Yes, good point. Now that the backend defines the starting point of time this might be easier to apply to timers.

PR #804, opened 2022-01-05
Implement Default for VWeakMapped

ogoffart:


tronical (approved):

ogoffart:
(The reason i did that is because i saw that almost all field in the generated component implements default, so we can just do #[derive(Default)]  instead of writing it manually.  Well, the VWeakMapped field did not implement default until now.

PR #805, opened 2022-01-05
Allow for bigger indices in VisitChildrenResult

hunger:
The indexes stored in `VisitChildrenResult` are unsigned. We have 64
bits to store two values and we need to have one special value as a flag.

So accept any index `< u32::MAX` instead of `< i32::MAX`, which should
allow for more data to be visited;-)

I added the breaking change label, as this does change the interface we expose in a subtle way.

### DIFF
--- a/api/sixtyfps-cpp/include/sixtyfps.h
+++ b/api/sixtyfps-cpp/include/sixtyfps.h
@@ -695,16 +695,16 @@ public:
         viewport_height->set(h);
     }
 
-    intptr_t visit(TraversalOrder order, private_api::ItemVisitorRefMut visitor) const
+    uintptr_t visit(TraversalOrder order, private_api::ItemVisitorRefMut visitor) const
     {
         for (std::size_t i = 0; i < inner->data.size(); ++i) {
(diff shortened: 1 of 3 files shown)


tronical:
> I added the breaking change label, as this does change the interface we expose in a subtle way.

Fortunately that is not part of the public API of the API crate or the generated code, so I think this isn't quite breaking.

ogoffart (approved):
Yes, that's not a breaking change

Even if VisitChildrenResult was exposed in the public API,  the .0 is not `pub`  so it's internal only

ogoffart:
But this broke the C++ build

hunger:
I added the `breaking change` label since the entire thing is treated as one 64bit integer in the FFI layer and that type has changed. So that is visible more widely than the private definition in rust implies.

ogoffart:
> I added the breaking change label since the entire thing is treated as one 64bit integer in the FFI layer and that type has changed. So that is visible more widely than the private definition in rust implies.

Yeah, the meaning of this label is change that breaks the downstream users.  But the FFI layer is also internal to us.  

However we need to adapt the C++ layer of course, and it is still throwing some warning apparently (CI failure)

hunger:
I saw it and am onto those.

PR #808, opened 2022-01-06
Don't panic in the compiler when encountering a Path element with for-in

tronical:
Instead, produce a diagnostic message.

cc #754

### DIFF
--- a/sixtyfps_compiler/passes/compile_paths.rs
+++ b/sixtyfps_compiler/passes/compile_paths.rs
@@ -98,8 +98,16 @@ pub fn compile_paths(
                             element_name
                         ),
                     };
-                    let bindings = std::mem::take(&mut child.borrow_mut().bindings);
-                    path_data.push(PathElement { element_type, bindings });
+
+                    if child.borrow().repeated.is_some() {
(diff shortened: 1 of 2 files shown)


sixtyfps_compiler/passes/compile_paths.rs
```diff
@@ -98,8 +98,16 @@ pub fn compile_paths(
                             element_name
                         ),
                     };
-                    let bindings = std::mem::take(&mut child.borrow_mut().bindings);
-                    path_data.push(PathElement { element_type, bindings });
+
+                    if child.borrow().repeated.is_some() {
+                        diag.push_error(
+                            "Path elements are not supported with `for`-`in` syntax".into(),
```
ogoffart, line 104: should we say "yet" and link to the issue?
That way people can be invovled

ogoffart (approved):

sixtyfps_compiler/passes/compile_paths.rs
tronical: Sure, I was a bit hesitant but if we're certain that we're going to support this eventually then it makes sense.

PR #809, opened 2022-01-06
Add key_code for `BackTab`

hunger:


sixtyfps_runtime/rendering_backends/gl/event_loop.rs
```diff
@@ -354,6 +354,21 @@ fn process_window_event(
...
+    fn key_event(
+        event_type: KeyEventType,
+        text: SharedString,
+        modifiers: KeyboardModifiers,
+    ) -> KeyEvent {
+        let mut event = KeyEvent { event_type, text, modifiers };
+
+        // map Shift-Tab into (Shift) Backtab to have a similar behavior as Qt backend
+        if event.text == "\t" && modifiers.shift {
+            event.text = SharedString::from("\u{0019}");
```
ogoffart, line 366: can you use `corelib::input::key_codes::Backtab` instead of hard-coding?

```suggestion
            let mut buffer = [0; 6];
            event.text = SharedString::from(corelib::input::key_codes::Backtab.encode_utf8(&mut buffer) as &str);
```

hunger:
@ogoffart: Is your approach involving a the utf8-encoding into a buffer faster than just converting to a string like I did in the newest version?

ogoffart (approved):

ogoffart:
> Is your approach involving a the utf8-encoding into a buffer faster than just converting to a string like I did in the newest version?

Yes, you are doing two allocations: the String, then the SharedString.

But that's good enough.

PR #812, opened 2022-01-06
allow FnMut as argument for `on_callback`

Be-ing:


ogoffart:
Thanks for the patch.
Could you change the test in callbacks/handler.60 so this does not regress?

ogoffart:
Thanks

PR #813, opened 2022-01-07
Feat: Add math funcs Log10, Log2, Log and Exp

### LINKED ISSUE #755: exponent and logarithm math functions
It would be useful to use exponential and logarithm functions on numeric values in `.60` files.

jrmoulton:
Fixes #755
I'm not sure if the Log10 and Log2 functions need to stay but I added them because both c++ and rust have them as a standard.


ogoffart:
Great change, thanks!
It just missed tests (in math.60 for example) and documentation (in langref.md)

I wonder if we really need log2 and log10 if we can use log(foo, 2) and log(bar, 10). Also I think exp should take a parameter with the exponent.

ogoffart:
I realize that C++ and rust and JavaScript calls the exp function with an arbitrary base "pow". So we should have this function as well.

jrmoulton:
`exp` shouldn't take a power it's just e^x. But I could also add a pow function for a more generic exponential function which I guess might be what the issue was referring to. 

I'd lean towards removing `log10` and `log2` and just using `log`. By the same token should I remove `exp` and just implement a `pow` function?

ogoffart:
Looks good, thank you.
We'll merge once the CI passes.

ogoffart:
Ah, it is still missing the documentation in langref.md
And it also deserve an entry to the changelog.

jrmoulton:
There we go. That should be all of it (hopefully the changelog entry looks fine)

ogoffart:
Perfect, thanks!

PR #815, opened 2022-01-07
Speed up C++ builds

tronical:
Replace the use of the external fmtlib with C++20 std::format. This avoids extra cloning.

### DIFF
--- a/examples/iot-dashboard/CMakeLists.txt
+++ b/examples/iot-dashboard/CMakeLists.txt
@@ -4,18 +4,20 @@
 cmake_minimum_required(VERSION 3.14)
 project(sixtyfps_cpp_iot_dashboard LANGUAGES CXX)
 
+if (NOT "cxx_std_20" IN_LIST CMAKE_CXX_COMPILE_FEATURES)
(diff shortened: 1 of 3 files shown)


examples/iot-dashboard/CMakeLists.txt
```diff
@@ -4,18 +4,20 @@
 cmake_minimum_required(VERSION 3.14)
 project(sixtyfps_cpp_iot_dashboard LANGUAGES CXX)
 
+if (NOT "cxx_std_20" IN_LIST CMAKE_CXX_COMPILE_FEATURES)
+    message(STATUS "The C++ iot dashboard demo requires C++ >= 20. The compiler you're using does not support that, so the build of this example is skipped.")
+    return()
```
ogoffart, line 9: Either this test does not work, or the Ubuntu 20.04 compiler claim to support c++20  while it doesn't support format

ogoffart:
According to https://en.cppreference.com/w/cpp/compiler_support , it is not supported by GCC yet

tronical:
Right, I was a bit too optimistic about this being supported. I’ll close this again, maybe it’ll become an option in the future (or I port the code away from fmtlib).

PR #816, opened 2022-01-08
Fix warning in C++ bindgen code

hunger:
IntoIter::new is deprecated and the rust compiler asks to use .iter()
instead.

Let's see how CI likes that: Not sure this is already supported on the oldest rust compiler we need to support:-)


ogoffart:
The replacement for IntoIter::new is IntoIterator::into_iter.  (Or just .into_itet() once we move to the 2021 edition)

But here it does not matter since it str& which is copy anyway.

PR #817, opened 2022-01-08
CMake: Require C++ 17

hunger:
Our code uses some C++ 17 extensions, so document that in the
CMakeLists.txt file.


ogoffart:
You know cmake better than I do, so that's probably right. I wonder if that shouldn't go in api/sixtyfps-cpp

hunger:
Is api/sixtyfps-cpp going to be use stand-alone and not via the top-level CMakeLists.txt file?

Then I should definitely add it there:-)

PR #822, opened 2022-01-09
Move from stuctopt to clap 3

hunger:
* Do not explicitly enable color output: That is default anyway
* Consistently use from_os_str() for file pathes
* Fix wrong types used in some places


tools/compiler/main.rs
```diff
@@ -1,46 +1,41 @@
...
 // SPDX-License-Identifier: (GPL-3.0-only OR LicenseRef-SixtyFPS-commercial)
 
+use clap::Parser;
 use sixtyfps_compilerlib::diagnostics::BuildDiagnostics;
 use sixtyfps_compilerlib::*;
 use std::io::Write;
-use structopt::StructOpt;
 
-#[derive(StructOpt)]
+#[derive(Parser)]
```
ogoffart, line 9:
```suggestion
#[derive(clap::Parser)]
```

Because I think the parser term is ambiguous with our own parser

ogoffart (approved):
Looks alright, apart from the fact that the "Parser" drive is a bit an unfortunate generic name choice

tools/compiler/main.rs
hunger: Good point, I'll change that!

PR #825, opened 2022-01-10
PropertyAnimation: Make `loop-count: n` run animations n times

hunger:
We used to run `n+1` times instead, which seems wrong to me. It is also
not how the similar property in CSS is documented to work.

As a side-effect of this change, the animation will no longer play when
overshooting the playback interval.

### DIFF
--- a/docs/langref.md
+++ b/docs/langref.md
@@ -678,7 +678,8 @@ Animation can be configured with the following parameter:
 
 * `delay`: the amount of time to wait before starting the animation
 * `duration`: the amount of time it takes for the animation to complete
-* `loop-count`: FIXME
+* `iteration-count`: The number of times a animation should run. A negative value specifies
+    infinite reruns. Fractual values are possible.
 * `easing`: can be `linear`, `ease`, `ease-in`, `ease-out`, `ease-in-out`, `cubic-bezier(a, b, c, d)` as in CSS
 
 It is also possible to animate several properties with the same animation:
--- a/tools/syntax_updater/from_0_1_0.rs
+++ b/tools/syntax_updater/from_0_1_0.rs
@@ -1,0 +1,36 @@
+// Copyright © SixtyFPS GmbH <info@sixtyfps.io>
+// SPDX-License-Identifier: (GPL-3.0-only OR LicenseRef-SixtyFPS-commercial)
+
+use std::io::Write;
+
+use sixtyfps_compilerlib::parser::{syntax_nodes, SyntaxNode};
+
+/// Rename `loop-count` to `iteration-count` in `PropertyAnimation`s
+pub(crate) fn fold_node(
+    node: &SyntaxNode,
+    file: &mut impl Write,
+    state: &mut crate::State,
+) -> std::io::Result<bool> {
+    if let Some(binding) = syntax_nodes::Binding::new(node.clone()) {
+        let property_name = state.property_name.as_deref().unwrap_or_default();
+        if (property_name == "loop-count" || property_name == "loop_count")
+            && has_parent_anim(&binding.clone().into())
+        {
+            let text = node.text().to_string();
+            let text = text.replace("loop-count", "iteration-count");
+            let text = text.replace("loop_count", "iteration-count");
+            file.write_all(text.as_bytes())?;
+            return Ok(true);
+        }
+    };
+    Ok(false)
+}
+
+fn has_parent_anim(node: &SyntaxNode) -> bool {
+    let node = node.parent();
+    if let Some(node) = node {
+        syntax_nodes::PropertyAnimation::new(node).is_some()
+    } else {
+        false
+    }
+}
--- a/internal/compiler/builtins.60
+++ b/internal/compiler/builtins.60
@@ -408,7 +408,7 @@ PropertyAnimation := _ {
     property <duration> delay;
     property <duration> duration;
     property <easing> easing;
-    property <int> loop-count;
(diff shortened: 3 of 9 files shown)


ogoffart:
This is a behaviour change but I'm tempted to say it's ok because nobody is probably using the loop count

sixtyfps_runtime/corelib/items.rs
```diff
@@ -991,6 +991,12 @@ pub struct PropertyAnimation {
     pub easing: crate::animations::EasingCurve,
 }
 
+impl Default for PropertyAnimation {
+    fn default() -> Self {
+        Self { duration: 0, loop_count: 1, easing: crate::animations::EasingCurve::default() }
```
tronical, line 996: I think Rust generated code and the interpreter call this default function, but C++ generated code will not. Maybe the default needs to go into builtins.60?

hunger: Is that possible already?

All tests still passed for me, but looks like I somehow missed the C++ tests:-/

tronical: It’s not your fault :). I’m not sure if there are tests for loop-count that use a .60 file.

hunger: Found it. Now some more different tests fail though:-/

sixtyfps_compiler/object_tree.rs
```diff
@@ -1283,12 +1283,26 @@ fn animation_element_from_node(
...
+        if let Type::Builtin(builtin_base) = &anim_element.base_type {
+            for (prop, info) in &builtin_base.properties {
+                if let Some(expr) = &info.default_value {
+                    anim_element
+                        .bindings
+                        .entry(prop.clone())
+                        .or_insert_with(|| RefCell::new(expr.clone().into()));
+                }
+            }
+        }
```
tronical, line 1304: Just a thought: Since Element::from_node has "similar" code, maybe it's worth sharing this in a helper function? Like this:

```rust
fn apply_default_values(bindings: &mut BindingsMap, base_type: &Type) {
    // ...
}
```

and `Element::from_node` could call it with `(&mut anim_element.bindings, &anim_element.base_type)`

hunger: Good point, done.

hunger:
OK, let's keep this for 0.2.

tronical:
> OK, let's keep this for 0.2. It needs renaming of the property to `iteration-count` then!

When renaming it would be good to also cover this in the updater tool.

We also need to make this tool available for C++ users and document it, I think.

docs/langref.md
```diff
@@ -678,7 +678,8 @@ Animation can be configured with the following parameter:
 
 * `delay`: the amount of time to wait before starting the animation
 * `duration`: the amount of time it takes for the animation to complete
-* `loop-count`: FIXME
+* `iteration-count`: The number of times a animation should run. A negative value specifies
+    infinite reruns. Fractual values are possible.
```
tronical, line 682:
```suggestion
    infinite iterations. Fractional values are possible.
```

hunger:
This works-for-me(TM) but did not feel like the proper solution... But I was not sure how to extract and replace the identifier properly:-/

ogoffart:
To be perfect, one should check the value and increment it. But I guess it's not so important

tools/syntax_updater/from_0_1_0.rs
```diff
@@ -0,0 +1,34 @@
...
+
+/// Rename `loop-count` to `iteration-count` in `PropertyAnimation`s
+pub(crate) fn fold_node(
+    node: &SyntaxNode,
+    file: &mut impl Write,
+    state: &mut crate::State,
+) -> std::io::Result<bool> {
+    if let Some(binding) = syntax_nodes::Binding::new(node.clone()) {
+        if state.property_name.as_deref().unwrap_or_default() == "loop-count"
+            && has_parent_anim(&binding.clone().into())
```
ogoffart, line 16: I think you also need to consider loop_count with underscore

hunger: I expected earlier steps to have normalized to `loop-count` already. Is that not the case?

hunger:
@simon: The file I used to test with locally failed to convert in the CI. Is that due to the CI taking some input from a different branch again?

It also has an extra fail on `Clip`. How should I transform that?

tronical:
> @simon: The file I used to test with locally failed to convert in the CI. Is that due to the CI taking some input from a different branch again?

I don't believe that's the case. The updater job is spawned directly by `ci.yaml`, there's no separate workflow with `@master` involved AFAICS (for this test).

tools/syntax_updater/from_0_1_0.rs
ogoffart: nope, rowan is not really an AST, more like a concrete syntax tree. It just has the raw data.

ogoffart:
>  Is that due to the CI taking some input from a different branch again?

Yes, that's the case, and on purpose!
It loads the .60 files from previous versions with a checkout of an older tag

tools/syntax_updater/from_0_1_0.rs
hunger: You are right: I somehow missed the code in the `fold_token` function in `main.rs` is just an example:-)

hunger:
Oh, cool, this time it just worked.

ogoffart (approved):

PR #827, opened 2022-01-11
Link to the sixtyfps interpreter crate from the Rust API crate

tronical:
Our web site and READMEs direct to our Rust API crate, which makes sense as it provides
fundamental traits and types. However as pointed out in
#811 we should advertise the existence of the interpreter API.

Ideally we'd link to the crate documentation that

    (1) perfect matches the version number
    (2) uses sixtyfps.io for our built docs and docs.rs for the published crates

I could not find a way to do this (not without making the crate a dependency and adding features), so
this patch is a compromise that I've seen used in other
crates.

### DIFF
--- a/api/sixtyfps-rs/lib.rs
+++ b/api/sixtyfps-rs/lib.rs
@@ -20,16 +20,17 @@ If you are already familiar with SixtyFPS, the following topics provide some
 
 ## How to use this crate:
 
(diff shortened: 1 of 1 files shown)


ogoffart (approved):

hunger (approved):
Not perfect, but better than what we had before.

PR #829, opened 2022-01-11
Add `delay` property to `PropertyAnimation`

hunger:
This is a very non-intrusive implementation of a `delay` property on `PropertyAnimation`.

This version will only add a `delay` before the first animation. In the presence of `loop-count`, these other animations will not be delayed.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -23,6 +23,8 @@ All notable changes to this project will be documented in this file.
    special keys are handled
  - `start()`, `stop()`, `running()` and a default constructor for C++ `sixtyfps::Timer`
  - Math functions `log`, and `pow`
+ - Property animations now have a `delay` property, which will delay the start
+   of the animation.
 
 ### Fixed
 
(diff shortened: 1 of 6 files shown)


ogoffart:
Can you add a test in `transitions.60` ?  (Or in a new file based on it, if you prefer)

ogoffart:
Thanks you that's great already.  :+1: 

The test could cover more case like if there are two properties in a transitions with different delay, and make sure to check the value of the property right before and after the animation is supposed to start.
(Note that in the test, there is no real time, so something like `sixtyfps::testing::mock_elapsed_time(50);` will ensure that exactly 50 fake ms is being "spent". the tests should not be flacky as they don't really spend time, and it will not make the test slow.)

hunger:
I saw that:-) These tests are very "fuzzy" with the values they compare though. Is there a reason for that?

ogoffart:
> These tests are very "fuzzy" with the values they compare though. Is there a reason for that?

Habits :-) 
Some tests are more fuzzy than other.
It is hard to be exact because of float interpolation. Also when using an easing curve (which is not the case here admittedly).
But it should work without being too fuzzy

hunger:
More testing in the latest version:-)

PR #830, opened 2022-01-12
Verify the existence of imported custom fonts

tronical:
... and produce diagnostics.

That way in later phases we don't need to check again.

### DIFF
--- a/docs/langref.md
+++ b/docs/langref.md
@@ -1016,7 +1016,7 @@ instructions the SixtyFPS compiler to include the font and makes the font famili
 
 For example:
 
-```60
+```60,notest
 import "./NotoSans-Regular.ttf";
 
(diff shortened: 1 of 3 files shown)


tronical:
Cherry-picked from the glyph pre-rendering branch. Since we had the discussion about import diagnostics in mattermost (regarding images) it reminded me that there's no reason to wait with this patch :)

ogoffart (approved):
I just wonder if we should have it a warning in the minor release.

tronical:
> I just wonder if we should have it a warning in the minor release.

Fair question. On the other hand I think it will be an error later at compile time, so the sooner the error is reported the better IMO.

tronical:
But I'm open to making this a warning, if you prefer.

PR #831, opened 2022-01-12
LLR (rust part)

ogoffart:
Create a new intermediate representation (Low Level Representation)  to  reduce duplication in the generator code


sixtyfps_compiler/builtins.60
```diff
@@ -187,6 +187,7 @@ export BoxShadow := _ {
 
 export struct Point := {
     //-name:sixtyfps::private_api::Point
+    //-rust_type_constructor:sixtyfps::re_exports::Point::new({})
```
ogoffart, line 190: Do we actually still need that line since we handle this case in rust.rs ?

sixtyfps_compiler/passes.rs
```diff
@@ -96,7 +98,9 @@ pub async fn run_passes(
         .chain(std::iter::once(root_component))
     {
         focus_item::resolve_element_reference_in_set_focus_calls(component, diag);
-        focus_item::determine_initial_focus_item(component, diag);
+        if Rc::ptr_eq(component, root_component) {
+            focus_item::determine_initial_focus_item(component, diag);
+        }
```
ogoffart, line 103: This is done because the determine_initial_focus_item sets the setup_code, and that was before only generated into code for the root component  but now it is done for each sub component.
We could make it such that we still only process the setup_code for the root_component,  but i think it makes more sense to generate it for each component.
Now the question is if the setup_code is a good idea to contain the code that focus the first item, especially since it won't work if there are repeater involved.  But we can think about focus handling later.

sixtyfps_compiler/builtins.60
tronical: I think you're right, we may not need this anymore.

sixtyfps_compiler/passes.rs
tronical: I agree, this patch makes sense.

Somehow we need to convey to the run-time library which item should receive the focus when the `Window` receives the focus the very very first time. Instead of "pushing" that, we could also make an accessor in the component vtable, but I think that's more overhead for something that's needed only "once".

sixtyfps_runtime/corelib/item_tree.rs
```diff
@@ -93,6 +93,23 @@ pub enum ItemTreeNode<T> {
...
+        }
+    }
+    pub fn children_index(&self) -> Option<usize> {
+        match self {
+            ItemTreeNode::Item { item: _, children_count: _, children_index, parent_index: _ } => {
+                Some(*children_index as usize)
+            }
+            ItemTreeNode::DynamicTree { .. } => None,
+        }
+    }
```
tronical, line 110: I don't think we need this one anymore. I'm in favour of removing it. What do you think?

sixtyfps_compiler/generator/rust.rs
```diff
@@ -85,21 +85,10 @@ fn rust_type(ty: &Type) -> Option<proc_macro2::TokenStream> {
...
-        diag.push_error(format!("Cannot map property type {} to Rust", ty), type_node);
-        quote!(_)
-    })
-}
-
 /// Generate the rust code for the given component.
 ///
 /// Fill the diagnostic in case of error.
-pub fn generate(doc: &Document, diag: &mut BuildDiagnostics) -> Option<TokenStream> {
+pub fn generate(doc: &Document, _diag: &mut BuildDiagnostics) -> Option<TokenStream> {
```
tronical: I guess we can remove _diag - a good sign that this can't produce diagnostics.

sixtyfps_compiler/llr/expression.rs
```diff
@@ -0,0 +1,355 @@
...
+#[derive(Debug, Clone)]
+pub enum Expression {
+    /// A string literal. The .0 is the content of the string, without the quotes
+    StringLiteral(String),
+    /// Number
+    NumberLiteral(f64),
+    /// Bool
+    BoolLiteral(bool),
+
+    /// Reference to the callback <name> in the <element>
```
tronical, line 20:
```suggestion
    /// Reference to a property (which can also be a callback) or an element (property name is empty then).
```

sixtyfps_compiler/llr/expression.rs
```diff
@@ -0,0 +1,355 @@
...
+    PropertyReference(PropertyReference),
+
+    // TODO
+    // CallbackReference(P::PropertyReference),
+    /// Reference to a function built into the run-time, implemented natively
+    //BuiltinFunctionReference(BuiltinFunction, Option<SourceLocation>),
+
+    /// A reference to a specific element. This isn't possible to create in .60 syntax itself, but intermediate passes may generate this
+    /// type of expression.
+    //ElementReference(P::ItemReference),
```
tronical, line 30:
```suggestion
```

sixtyfps_compiler/llr/item_tree.rs
```diff
@@ -0,0 +1,245 @@
...
+    InParent { level: NonZeroUsize, parent_reference: Box<PropertyReference> },
+    /// The property within a GlobalComponent
+    Global { global_index: usize, property_index: usize },
+}
+
+#[derive(Debug)]
+pub struct Property {
+    pub name: String,
+    pub ty: Type,
+    //pub binding: Option<BindingExpression>,
```
tronical, line 62:
```suggestion
```

sixtyfps_compiler/llr/lower_expression.rs
```diff
@@ -0,0 +1,977 @@
...
+        }
+        self.mapping.map_property_reference(from, self.state)
+    }
+}
+
+impl super::EvaluationContext for ExpressionContext<'_> {
+    fn property_ty(&self, _: &PropertyReference) -> &Type {
+        todo!()
+    }
+}
```
tronical, line 55: Not for this merge request, but later: I think this is unreachable code, and with that the only reason for EvaluationContext being a trait. In the cpp branch maybe we can remove that.

sixtyfps_compiler/llr/lower_expression.rs
```diff
@@ -0,0 +1,977 @@
...
+
+pub fn lower_expression(
+    expression: &tree_Expression,
+    ctx: &ExpressionContext<'_>,
+) -> Option<llr_Expression> {
```
tronical, line 60: I wonder what it means if we failed to lower an expression.

sixtyfps_compiler/llr/lower_expression.rs
```diff
@@ -0,0 +1,977 @@
...
+        tree_Expression::SolveLayout(l, o) => solve_layout(l, *o, ctx),
+    }
+}
+
+fn lower_assignment(
+    lhs: &tree_Expression,
+    rhs: &tree_Expression,
+    op: char,
+    ctx: &ExpressionContext,
+) -> Option<llr_Expression> {
```
tronical, line 202: AFAICS all branches return `Some()`, so maybe we don't need the `Option`?

tronical (approved):

sixtyfps_compiler/generator/rust.rs
```diff
@@ -596,7 +596,7 @@ fn generate_sub_component(
         init.push(quote!(#sub_component_id::init(
             VRcMapped::map(self_rc.clone(), |x| Self::FIELD_OFFSETS.#field_name.apply_pin(x)),
             &#root_ref_tokens,
-            #global_index, #global_children
+            #global_index, dbg!({ stringify!(#inner_component_id | #sub_component_id); #global_children})
```
tronical, line 599: Looks like a stray intentional change? :)

sixtyfps_compiler/llr/lower_expression.rs
ogoffart: Yeah, currently we just unwrap it or worse, ignore it.
I think the only case is if the expression is Invalid. but maybe we should handle this case in the caller

sixtyfps_compiler/generator/rust.rs
ogoffart: oops

sixtyfps_compiler/llr/lower_to_item_tree.rs
```diff
@@ -264,33 +266,33 @@ fn lower_sub_component(
...
-            let animation = binding
-                .animation
-                .as_ref()
-                .filter(|_| !is_constant)
-                .map(|a| super::lower_expression::lower_animation(a, &ctx).unwrap());
-            sub_component
-                .property_init
-                .push((prop.clone(), BindingExpression { expression, animation, is_constant }))
+        if matches!(binding.expression, tree_Expression::Invalid) {
+            return;
```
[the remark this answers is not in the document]
tronical: Indeed! Well spotted!

tronical (approved):

PR #833, opened 2022-01-13
Parser: allow `.` to access member after any expression

ogoffart:
Before we would only allow `foo.bar` where `foo` was an identifier.
Now we also allow more complex expression such as `"foo".bar` or `(123 + foo).bar`
(in the parser)
In particular, this will allow to get the member of objects returned by functions
or, in the future, part of arrays

### DIFF
--- a/sixtyfps_compiler/parser/expressions.rs
+++ b/sixtyfps_compiler/parser/expressions.rs
@@ -13,6 +13,7 @@ use super::prelude::*;
 /// 42px
 /// #aabbcc
 /// (something)
+/// (something).something
 /// @image-url("something")
 /// @image_url("something")
 /// some_id.some_property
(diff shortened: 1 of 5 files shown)


sixtyfps_compiler/parser/expressions.rs
```diff
@@ -85,12 +87,27 @@ fn parse_expression_helper(p: &mut impl Parser, precedence: OperatorPrecedence)
         }
     }
 
-    if p.nth(0).kind() == SyntaxKind::LParent {
-        {
-            let _ = p.start_node_at(checkpoint.clone(), SyntaxKind::Expression);
+    loop {
+        match p.nth(0).kind() {
```
tronical, line 91: Is this the match where the `SyntaxKind::LBracket` will go for `IndexExpression` later?

tronical (approved):

sixtyfps_compiler/parser/expressions.rs
ogoffart: yes, exactly.
Yhis will allow the parser to understand things like `foo()[34][22].xx`  for example

PR #605, opened 2021-10-23
Array index

jamesblacklock:
This branch adds support for indexing into arrays/models.
As it stands, out-of-bounds access will panic. Will change the behavior as directed.
As before, I am unfortunately not able to test the cpp implementation.

PR #836, opened 2022-01-14
Add support tracked array data access by index

tronical:
This is based on #605 and adds support for array index expressions for writing as well as change tracking for property bindings.

### DIFF
--- a/api/sixtyfps-cpp/include/sixtyfps.h
+++ b/api/sixtyfps-cpp/include/sixtyfps.h
@@ -487,24 +487,43 @@ public:
     /// \private
     /// Internal function called from within bindings to register with the currently
     /// evaluating dependency and get notified when this model's row count changes.
-    void track_row_count_changes() { model_dirty_property.get(); }
+    void track_row_count_changes() { model_row_count_dirty_property.get(); }
+
+    /// \private
+    /// Internal function called from within bindings to register with the currently
+    /// evaluating dependency and get notified when this model's row data changes.
+    void track_row_data_changes(int row)
+    {
+        auto it = std::lower_bound(tracked_rows.begin(), tracked_rows.end(), row);
+        if (it == tracked_rows.end() || row < *it) {
+            tracked_rows.insert(it, row);
+        }
+        model_row_data_dirty_property.get();
+    }
 
 protected:
     /// Notify the views that a specific row was changed
     void row_changed(int row)
     {
+        if (std::binary_search(tracked_rows.begin(), tracked_rows.end(), row)) {
+            model_row_data_dirty_property.mark_dirty();
+        }
         for_each_peers([=](auto peer) { peer->row_changed(row); });
     }
     /// Notify the views that rows were added
     void row_added(int index, int count)
     {
-        model_dirty_property.mark_dirty();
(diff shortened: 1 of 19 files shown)


api/sixtyfps-cpp/include/sixtyfps.h
```diff
@@ -595,6 +616,13 @@ class VectorModel : public Model<ModelData>
         data.erase(data.begin() + index);
         this->row_removed(index, 1);
     }
+
+    /// Inserts the given value as a new row at the specified index
+    void insert_at(size_t index, const ModelData &value)
```
ogoffart, line 621: Why is it called `insert_at`?  std::vector uses `insert`.
I guess that's because the vector::insert takes an iterator and not an index, but i'd still go with the standard terminology because:
 - we called the remove function `erase` , and not `erase_at`
 - one could argue that the iterator equivalent for the model is an int
 - string::insert is also taking a position.
 - that's what C++ developer are the most likely to look for.

docs/langref.md
```diff
@@ -262,6 +262,7 @@ Example := Window {
 ```
 
 * **`length`**: One can query the length of an array and model using the builtin `.length` property.
+* **`array[index]`**: Individual elements of an array can be retrieved using the `array[index]` syntax.
```
ogoffart, line 265: should we specify what happens when we go out of bounds?  should we have examples in the documentation?

api/sixtyfps-cpp/include/sixtyfps.h
tronical: Okay. I don't have a strong feeling about it - reasoning as you mentioned also in the commit message. But I can change the commit to call it `insert()`.

sixtyfps_compiler/llr/lower_expression.rs
```diff
@@ -265,6 +269,25 @@ fn lower_assignment(
...
         }
+        tree_Expression::ArrayIndex { array, index } => {
+            let rhs = lower_expression(rhs, ctx);
+            let array = Box::new(lower_expression(array, ctx));
+            let index = Box::new(lower_expression(index, ctx));
+            let value = Box::new(if op == '=' {
+                rhs
+            } else {
+                // FIXME: this will compute the index and the array twice:
+                // Ideally we should store the index and the array in local variable
```
tronical, line 280: Sort of related to this comment: I wonder if we should do the calls to track_row_data_changes() (as well as track_row_count_changes()) in the llr?

AFAICS it would require us to have scoped local variables.

ogoffart: I think this is not so frequent that we do `foo[xxx] += something;`
Maybe it is more usual to do `foo[xxx].something += something;` which is a different code path.
We can optimize that later if we need to.

tronical: True, optimisation wise this may not be worth it. I was thinking more in terms of complexity in the code generators (all three need to know about when exactly to call the two tracking methods). Anyway, can be done later :-)

api/sixtyfps-cpp/include/sixtyfps.h
tronical: Done :)

Issue #772, opened 2021-12-25
Panic with `Recursion Detected` when using a VerticalLayout inside of a Scrollview


jrmoulton:
I'm trying to use a VerticalLayout inside of a scroll view but the code panics. 

The problematic code is
```
SideBar := Rectangle {
	width: 185px;

	ScrollView {
		VerticalLayout {
			SearchBox {
				width: parent.width - 20px;
			}
			ListenSection {  }
		}
	}
}
```
Is there a better way to do this?

ogoffart:
This is caused by a binding loop, which is not detected by the compiler for some reason (bug in the compiler probably). I would need to have a reproducable example to debug it.

jrmoulton:
```
SideBar := Rectangle {
    width: 185px;

    ScrollView {
        VerticalLayout {
            Rectangle {
                width: parent.width - 20px;
            }
        }
    }
}
```
Ahh ok the problem is coming from the `width: parent.width` - 20px; line. If that line isn't there everything works fine but if it is there then it crashes with a panic

jrmoulton:
And with a little more testing it only happens when the `VerticalLayout` is inside the `ScrollView`. If the `ScrollView` is inside the `VerticalLayout` it compiles. And if only one of the two is present that will also compile.

ogoffart:
I have debugged the reason why we don't detect a loop.
It turns out this can be reproduced in this testcase:

```
Alias := Rectangle {
    property <int> viewport_width ;
    property <int> ps_width <=> viewport_width;
}

Foo := Rectangle {
    property <int> layout_info: alias.viewport_width;

    alias := Alias {   ps_width: layout_info;   }

    Text {
        text: layout_info;
    }
}
```

ogoffart:
Now this is properly detected as a compile time error (binding loop)

PR #840, opened 2022-01-17
Fixes loop detection involving sub components and two ways binding

ogoffart:
The original binding analysis code was considering that all the sub components
were inlined. But when this is not the case, we need to re-analyze each
sub component within their parent.

We also need to take care of reverse aliases

Will cause a .60 compile time binding loop error instead of a
runtime recursion panic in code like issue #772

### DIFF
--- a/sixtyfps_compiler/passes/binding_analysis.rs
+++ b/sixtyfps_compiler/passes/binding_analysis.rs
@@ -3,7 +3,11 @@
 
 //! Compute binding analysis and attempt to find binding loops
 
+use std::collections::HashMap;
+use std::collections::HashSet;
 use std::rc::Rc;
+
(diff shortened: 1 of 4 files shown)


sixtyfps_compiler/passes/binding_analysis.rs
```diff
@@ -80,86 +166,124 @@ fn analyse_binding(
...
+    assert_eq!(&o.unwrap(), current);
+
+    return depends_on_external;
+}
+
+/// Process the property `prop`
+///
+/// This will visit all the bindings from that property
+///
+/// `canonical` is the property wh
```
tronical, line 240: What's the "property wh"? :)

tronical (approved):

PR #842, opened 2022-01-18
LLR C++ generator

tronical:
This is the port of the C++ code generator to use the LLR representation of the compiler.


hunger (approved):

sixtyfps_compiler/llr/item_tree.rs
```diff
@@ -167,6 +167,44 @@ impl TreeNode {
...
+    pub fn item<'a>(&self, root_sub_component: &'a SubComponent) -> Option<&'a Item> {
+        if self.repeated {
+            None
+        } else {
+            let sub_component = self.sub_component(root_sub_component);
+            Some(sub_component.items.get(self.item_index).expect("Illegal item index in TreeNode"))
+        }
+    }
+
+    pub fn repeated_element<'a>(
```
tronical, line 192: AFAICS these three functions are not called - I think that we can remove them again.

sixtyfps_compiler/llr/lower_to_item_tree.rs
```diff
@@ -36,7 +36,15 @@ pub fn lower_to_item_tree(component: &Rc<Component>) -> PublicComponent {
...
-        sub_components: state.sub_components.into_values().map(|sc| sc.sub_component).collect(),
+        sub_components: component
+            .used_types
+            .borrow()
+            .sub_components
+            .iter()
+            .map(|tree_sub_compo| {
+                state.sub_components[&ByAddress(tree_sub_compo.clone())].sub_component.clone()
+            })
+            .collect(),
```
tronical, line 47: I wonder if there's a better way of doing this, so happy to change this if you have any ideas :)

sixtyfps_compiler/generator/cpp.rs
```diff
@@ -500,97 +469,8 @@ fn handle_property_binding(
...
-            ),
-            name: repeater_id,
-            ..Default::default()
-        }),
-    ));
-}
-
 /// Returns the text of the C++ code produced by the given root component
-pub fn generate(doc: &Document, diag: &mut BuildDiagnostics) -> Option<impl std::fmt::Display> {
+pub fn generate(doc: &Document) -> Option<impl std::fmt::Display> {
```
tronical: Ah yes, here, too, we can remove the `Option`. I'll do that.

tests/cases/expr/abs.60
```diff
@@ -0,0 +1,35 @@
+// Copyright © SixtyFPS GmbH <info@sixtyfps.io>
+// SPDX-License-Identifier: (GPL-3.0-only OR LicenseRef-SixtyFPS-commercial)
+
+ TestCase := Rectangle {
+    property<float> t1: abs(42.3);
+    property<float> t2: abs(-42.3);
+    property<int> t3: abs(42.3);
+    property<int> t4: abs(-42.3);
+}
```
ogoffart, line 9: I'm sure you saw that there was already tests for all these functions in `math.60` and `trigo.60` and the like.
I'm not against having more tests, quite the contrary. But each tests has additional overhead in C++ to rerun the compiler which is slow. Same for node.  So maybe grouping similar tests in the same file can speedup the test executions.

Also, for the interpreter test, it is good to have a `property <bool> test: ....` that evaluates to true. (the interpreter is also tested by the nodejs anyway, but the interpreter tests being much faster, it's good to catch problem in the interpreter test itself)

ogoffart (approved):

PR #845, opened 2022-01-19
cpp generator: reduce surface of public C++ API

tronical:
After generating the root component's item tree, with all public fields,
hide all variables and functions, with three exceptions.

All meanwhile added declarations for sub-components (such as popups, child repeaters)
still get access through friendship.

### DIFF
--- a/sixtyfps_compiler/generator/cpp.rs
+++ b/sixtyfps_compiler/generator/cpp.rs
@@ -529,7 +529,14 @@ pub fn generate(doc: &Document) -> impl std::fmt::Display {
     for sub_compo in &llr.sub_components {
         let sub_compo_id = ident(&sub_compo.name);
         let mut sub_compo_struct = Struct { name: sub_compo_id.clone(), ..Default::default() };
-        generate_sub_component(&mut sub_compo_struct, sub_compo, &llr, None, &mut file);
+        generate_sub_component(
+            &mut sub_compo_struct,
+            sub_compo,
+            &llr,
+            None,
+            Access::Public,
+            &mut file,
+        );
         file.definitions.extend(sub_compo_struct.extract_definitions().collect::<Vec<_>>());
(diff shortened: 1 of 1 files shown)


sixtyfps_compiler/generator/cpp.rs
```diff
@@ -621,10 +621,51 @@ fn generate_public_component(file: &mut File, component: &llr::PublicComponent)
...
+                if func.name != "create" && func.name != destructor_name =>
+            {
+                true
+            }
+            _ => false,
+        };
+        if hide {
+            *access = Access::Private;
+        }
     });
```
hunger, line 657: I am sure I will end up adding something with public access later and then wonder why that is generated as private!

Can't we do that when we define those instead of changing the Access behind the back of the programmer?

That should also work if you keep the code that adds the friends below I think.

tronical: I hope I understood your suggestion correct. I've pushed a new version of the patch.

sixtyfps_compiler/generator/cpp.rs
```diff
@@ -621,9 +628,28 @@ fn generate_public_component(file: &mut File, component: &llr::PublicComponent)
...
-    });
+    let old_declarations = file.declarations.len();
+
+    generate_item_tree(
+        &mut component_struct,
+        &component.item_tree,
+        &component,
+        None,
+        component_id.clone(),
+        Access::Private, // Hid eproperties and other fields from the C++ API
```
ogoffart, line 639:
```suggestion
        Access::Private, // Hide properties and other fields from the C++ API
```

ogoffart (approved):

sixtyfps_compiler/generator/cpp.rs
tronical: Thanks :)

PR #846, opened 2022-01-19
Add files to make individual crates reuse compliant

hunger:
This is _almost_ complete: There are still some icons and one image uncovered as of now.

### DIFF
--- a/.reuse/dep5
+++ b/.reuse/dep5
@@ -1,0 +1,56 @@
+Format: https://www.debian.org/doc/packaging-manuals/copyright-format/1.0/
+Upstream-Name: sixtyfps
+Upstream-Contact: SixtyFPS GmbH <info@sixtyfps.io>
+Source: https/sixtyfps.io/
+
+Files: .clang-format .gitattributes .gitignore */.gitignore .pre-commit-config.yaml .vscode/* cspell.json rustfmt.toml
+Copyright: Copyright © SixtyFPS GmbH <info@sixtyfps.io>
+License: (GPL-3.0-only OR LicenseRef-SixtyFPS-commercial)
+
+Files: CHANGELOG.md CONTRIBUTING.md FAQ.md README.md */CHANGELOG.md */README.md
+Copyright: Copyright © SixtyFPS GmbH <info@sixtyfps.io>
+License: (GPL-3.0-only OR LicenseRef-SixtyFPS-commercial)
+
+Files: helper_crates/const_field_offset/*
+Copyright: Copyright © SixtyFPS GmbH <info@sixtyfps.io>
+License: MIT or APACHE-2.0
+
+Files:  api/sixtyfps-cpp/docs/* docs/* api/*/*.json
+Copyright: Copyright © SixtyFPS GmbH <info@sixtyfps.io>
+License: (GPL-3.0-only OR LicenseRef-SixtyFPS-commercial)
+
+Files: examples/README.md examples/*/README.md examples/*.json examples/*/README examples/*/README.txt examples/qt_viewer/interface.ui
+Copyright: Copyright © SixtyFPS GmbH <info@sixtyfps.io>
+License: (GPL-3.0-only OR LicenseRef-SixtyFPS-commercial)
+
+Files: examples/memory/icons/*.svg examples/memory/icons/*.png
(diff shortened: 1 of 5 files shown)


api/sixtyfps-cpp/.reuse/dep5
```diff
@@ -0,0 +1,12 @@
+Format: https://www.debian.org/doc/packaging-manuals/copyright-format/1.0/
```
tronical, line 1: Hm, do we need this for creates that we don't publish?

hunger: I did just set this up for any folder containing a Cargo.toml file: Simplest possible approach:-)

Since we need to keep the copyright straight for all our code, that seems like a sensible thing to me.

ogoffart:
Many crate are not uploaded on crates.io  (the ones with publish = false, basically)  and don't need these symlinks. in particular not the example.
And the less symlinks the easier to maintain

.reuse/dep5
```diff
@@ -0,0 +1,4 @@
+Format: https://www.debian.org/doc/packaging-manuals/copyright-format/1.0/
+Upstream-Name: sixtyfps
+Upstream-Contact: SixtyFPS GmbH <info@sixtyfps.io>
+Source: https://sixtyfps.io/
```
ogoffart, line 4: What's that Source field?

ogoffart:
Do the dep5 files also need a license ?  :-}

ogoffart:
I don't think there is any use of putting the .reuse in the crates themselves. Is there?
None of our dependencies have that anyway so any rust users cannot use that.

api/sixtyfps-cpp/.reuse/dep5
tronical: Maybe I'm missing something in my understanding of who uses this. How are these files parsed by reuse and what is reuse's awareness of crates versus just directories?

ogoffart (changes requested):
I really think we should not add extra files for the examples and such.


Another thing: the filename of the cat is `cat_preview_round.png` but the image is not rounded.
IMHO it should eather be renamed to `cat.jpeg`  (png is really a bad format for pictures)  and we can use it within a `Rectangle{ clip: true; border-radius: 3px; }` which we can do now.  Or the image should be rounded.

.reuse/dep5
hunger: URL to some website where this thing comes from.

api/sixtyfps-cpp/.reuse/dep5
hunger: Reuse is blissfully unaware of crates:-) But we want the sources that are uplaoded to crates.io to be scan-able to get the inventory of licenses used. So those need to be self-contained in the tarball that gets sent to crates.io.

Strictly speaking we only need to have the crates we upload separated out. Maybe I should do that... it gets rid of the LICENSES and the .reuse folder, but comes at the small price of needing to create and populate those when we make decide to publish a formerly private crate.

hunger:
OK, now with less `.reuse` and `LICENSES` directories: Only for published crates and those in `tools` (as I expectthe unpublished ones to mature:-).

.reuse/dep5
hunger: `Source` is:

> Formatted text, no synopsis: an explanation from where the upstream source came from.

I could put in github if you think that is a better fit.

api/sixtyfps-cpp/.reuse/dep5
hunger: I removed this for the non-published crates.

ogoffart:
>  The .reuse/dep5 needs to be at the top level of the directory you want to run reuse from

But why would you want to use reuse from the hidden folder where cargo download our crates when used as dependencies?

hunger:
The patch shrank a lot now that so much of it has been merged:-)

The latest iteration limits itself to one top level reuse project and leaves the LICENSES folders for crates alone for now.

Next step is to have an xtask that runs reuse, generates a list of licenses applicable to all files and then actually populates the LICENSES folders with links to the top level based on what it finds there.

ogoffart (approved):

PR #847, opened 2022-01-19
Bump dependencies of the GL backend

tronical:


tronical:
This updates several dependencies, at the expensive of having two ttf-parser crates - until new upstream releases of rustybuzz and resvg use the latest parser.

ogoffart (approved):

PR #850, opened 2022-01-20
Update license information

hunger:
Use the reuse tool to get a better grip on licenses used in sixtyfps.
Fix a couple of licenses along the way.

* Uses creative commons for our own logo (commercial use!)
* Fixes some license information found in README files and documents
  them with proper .license files.
* Document Apache/MIT for helper_crates/const-field-offset which matches
  what its documentaion site says it uses.
* Add a list of licenses that apply to crates we publish and the tooling
  we have.

This patch only adds static information and does not contain any tooling
support.


ogoffart:
The cat picture needs to be rounded (that's in the name) or renamed.
I'd favor renaming it to just `cat.jpeg` 
(And use a Rectangle with clip and border radius if we need rounded corner)

hunger:
I do not want this patch to touch code, so I try to stay as close to what we had as possible.

My image preview of the original cat image was so small that I did not notice the rounded corners, sorry.

hunger:
fmt, figma_import and syntax_updater no longer come with LICENSES folder.

tronical (approved):

ogoffart (approved):

PR #855, opened 2022-01-20
Only consider that the height of a Text depends of its width of `wrap` is set.

ogoffart:
This should fix the failure in chiptrack

### DIFF
--- a/sixtyfps_compiler/tests/syntax/analysis/binding_loop_layout2.60
+++ b/sixtyfps_compiler/tests/syntax/analysis/binding_loop_layout2.60
@@ -1,6 +1,30 @@
 // Copyright © SixtyFPS GmbH <info@sixtyfps.io>
 // SPDX-License-Identifier: (GPL-3.0-only OR LicenseRef-SixtyFPS-commercial)
 
+Wrap := Rectangle {
+    property woo <=> text.wrap;
+
+    VerticalLayout {
+//  ^error{The binding for the property 'layout-cache' is part of a binding loop}
+//  ^^error{The binding for the property 'height' is part of a binding loop}
+        HorizontalLayout {
+//      ^error{The binding for the property 'layout-cache' is part of a binding loop}
+//      ^^error{The binding for the property 'width' is part of a binding loop}
+//      ^^^error{The binding for the property 'layoutinfo-v' is part of a binding loop}
+            text := Text {
+                text: "Hello World";
+            }
+            square := Rectangle {
+//                   ^error{The binding for the property 'height' is part of a binding loop}
+                width: height;
+//                    ^error{The binding for the property 'width' is part of a binding loop}
+                background: violet;
+            }
+        }
+        Rectangle {}
+    }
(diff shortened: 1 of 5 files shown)


tests/cases/layout/text_no_wrap.60
```diff
@@ -0,0 +1,24 @@
+// Copyright © SixtyFPS GmbH <info@sixtyfps.io>
+// SPDX-License-Identifier: (GPL-3.0-only OR LicenseRef-SixtyFPS-commercial)
+
+
+// A Text that does not use word wrap's height should not depends on the with
```
tronical, line 5:
```suggestion
// A Text that does not use word wrap's height should not depends on the width
```

sixtyfps_compiler/tests/syntax/analysis/binding_loop_layout2.60
```diff
@@ -1,6 +1,30 @@
...
+            square := Rectangle {
+//                   ^error{The binding for the property 'height' is part of a binding loop}
+                width: height;
+//                    ^error{The binding for the property 'width' is part of a binding loop}
+                background: violet;
+            }
+        }
+        Rectangle {}
+    }
+    property <bool> test: square.width == square.height;
```
tronical, line 26: Is this unused?

sixtyfps_compiler/tests/syntax/analysis/binding_loop_layout2.60
```diff
@@ -17,4 +41,9 @@ Test := Rectangle {
 //                ^error{The binding for the property 'width' is part of a binding loop}
         }
     }
+
+    w := Wrap { }
+    callback set_wrap();
+    set_wrap => { w.woo = TextWrap.word-wrap; }
```
tronical, line 47: The `Wrap` component here is produce errors because of the `width: height;` in the `Rectangle`, right?

Do I understand correctly that if the callback were invoked, there'd be a binding loop at run-time?

tronical (approved):

sixtyfps_compiler/tests/syntax/analysis/binding_loop_layout2.60
ogoffart: `test`  is read by the interpreter test, and it would panic if there was a real loop

sixtyfps_compiler/tests/syntax/analysis/binding_loop_layout2.60
ogoffart: correct.  if the text start word-wrapping, the height of the text depends on the width, and so there will be a loop. so that's why we must diagnose the loop at compile time

sixtyfps_compiler/tests/syntax/analysis/binding_loop_layout2.60
tronical: True, but I think the interpreter instantiates the `Test` below, not `Wrap`. Maybe that needs a "forwarding"?

PR #867, opened 2022-01-24
Update to Rust 1.56,  C++20,  and use std::span in C++

ogoffart:


api/sixtyfps-cpp/README.md
```diff
@@ -26,7 +26,7 @@ First you need to install the prerequisites:
 * Install Rust by following the [Rust Getting Started Guide](https://www.rust-lang.org/learn/get-started). Once this is done,
   you should have the ```rustc``` compiler and the ```cargo``` build system installed in your path.
 * **[cmake](https://cmake.org/download/)** (3.19 or newer)
-* A C++ compiler that supports C++17 (e.g., **MSVC 2019** on Windows)
+* A C++ compiler that supports C++20 (e.g., **MSVC 2019 16.6** on Windows)
```
ogoffart, line 29: This one seem to have good support for C++20 already.
I'm wondering if we should simplify this and say `MSVC 2020`  (released last november)

tronical: What is `MSVC 2020`? I only see MSVC 2019 (version 16) and MSVC 2022 (version 17) on the MS website. Maybe I overlooked something?

ogoffart: Sorry, i meant 2022

tronical: Ahh, yes, 2022 - that would be ideal. One thing I noticed with regards to MSVC is that the binaries Qt provides are still on 2019. I hope those are safe to mix with 2022.

CHANGELOG.md
```diff
@@ -3,6 +3,16 @@ All notable changes to this project will be documented in this file.
 
 ## Unreleased
 
+### Changed
+
+ - Minimum rust version is now 1.56
+ - C++ compiler requires C++20
+ - In the C++ interpreter API `std::span` is used for callbacks arguments, instead of `sixtyfps::Slice`
```
tronical, line 10: This is a good example of something where we should start documenting it. $customer is using `invoke_callback` with parameters, for example.

We can do that separately from this merge request though.

PR #869, opened 2022-01-24
Start a migration guide for C++ API for 0.2.0

tronical:
Once we have a Rust incompatible change we can change the sentencen in the ChangeLog to link to the Rust guide, too.

### DIFF
--- a/api/sixtyfps-cpp/docs/cpp_migration.md
+++ b/api/sixtyfps-cpp/docs/cpp_migration.md
@@ -6,7 +6,9 @@ This guide lists all API incompatible changes between major versions and describ
 
 ## Migrating from Version 0.1.x to 0.2.0
 
-In the 0.2.x series we have increased the minimum version of C++ and Rust that we require. You need to have Rust >= 1.56 installed and a C++ compiler that supports C++ 20 or newer. If you have installed Rust using `rustup`, then you can upgrade to the latest Version of Rust by running `rustup update`.
+In the 0.2.x series we have increased the minimum version of C++. You need to have a C++ compiler installed that supports C++ 20 or newer.
+
+If you are building SixtyFPS from source, you need to make sure that your Rust installation is up-to-date. If you have installed Rust using `rustup`, then you can upgrade to the latest Version of Rust by running `rustup update`.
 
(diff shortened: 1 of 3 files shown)


api/sixtyfps-cpp/docs/index.rst
```diff
@@ -39,6 +39,8 @@ Welcome to SixtyFPS C++'s documentation!
 
    markdown/debugging_techniques.md
 
+   cpp_migration.md
```
ogoffart, line 42: not sure if we need to call that cpp_migration.md or just migration.md
We are anyway in the sixtyfps-cpp folder after all.
(on the other hand, we would end up with several migration.md in the project. Which i don't know if it is a problem (we have many lib.rs files already))

tronical: That was exactly what I was thinking as well and motivating me to this prefix :)

api/sixtyfps-cpp/docs/cpp_migration.md
```diff
@@ -0,0 +1,29 @@
+# Migrating from Older Versions
+
+The C++ library is versioned according to the principles of [Semantic Versioning](https://semver.org). We define that the left-most non-zero component of the version is the major version, followed by the minor and optionally patch version. That means releases in the "0.y.z" series treat changes in "y" as a major release, which can contain incompatible API changes, while changes in just "z" are minor. For example the release 0.1.6 is fully backwards compatible to 0.1.5, but it contains new functionality. The release 0.2.0 however is a new major version compared to 0.1.x and may contain API incompatible changes.
+
+This guide lists all API incompatible changes between major versions and describes how you can migrate your application's source code.
+
+## Migrating from Version 0.1.x to 0.2.0
+
+In the 0.2.x series we have increased the minimum version of C++ and Rust that we require. You need to have Rust >= 1.56 installed and a C++ compiler that supports C++ 20 or newer. If you have installed Rust using `rustup`, then you can upgrade to the latest Version of Rust by running `rustup update`.
```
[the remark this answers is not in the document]
tronical: You're right, I'll change the Rust bit to encourage users to just install the latest :)

PR #871, opened 2022-01-24
Update to edition 2021

ogoffart:


sixtyfps_runtime/rendering_backends/qt/qt_window.rs
```diff
@@ -497,7 +497,7 @@ impl ItemRenderer for QtItemRenderer<'_> {
                 }
                 if (last_line_begin < string.size()) {
                     elided = string.left(last_line_begin);
-                    QString to_elide = QStringView(string).mid(last_line_begin, last_line_size).trimmed() % QStringView(u"…");
+                    QString to_elide = QStringView(string).mid(last_line_begin, last_line_size).trimmed() % QStringView(QT_UNICODE_LITERAL("…"));
```
tronical, line 500: Neat workaround :)

tronical:
I think our use of `resolver = "2"` in the top-level `Cargo.toml` can also be removed with this?

tronical (approved):

ogoffart:
Valgrind find the problem in the C++ tests: there are closure that were only using part of the object (in this case the reference to the closure, but not the destructor) so the c++ closure ended up destroyed before it was called. Cargo fix did not see that because the ffi feature was not enabled

sixtyfps_runtime/corelib/properties.rs
```diff
@@ -2387,10 +2387,16 @@ pub(crate) mod ffi {
...
+            fn call(&self) -> i32 {
+                (self.binding)(self.user_data)
+            }
+        }
+
         let c_state_binding = CStateBinding { binding, user_data, drop_user_data };
         let bind_callable = StateInfoBinding {
             dirty_time: Cell::new(None),
-            binding: move || (c_state_binding.binding)(c_state_binding.user_data),
+            binding: move || c_state_binding.call(),
```
tronical, line 2399: Ohh, so here it "moved" just the two fields (`binding` and `user_data`) and then called `drop` on the `c_state`, which called the `x` "destructor"?

tronical (approved):

sixtyfps_runtime/corelib/properties.rs
ogoffart: yes.

PR #873, opened 2022-01-25
Change Model::row_data to return an Option<T>

hunger:
Change API of Model to make row_data return an Option<T>, not a plain T.

This is more in line with how rust does things.


sixtyfps_runtime/corelib/model.rs
```diff
@@ -285,13 +285,9 @@ impl<'a, T> Iterator for ModelIterator<'a, T> {
...
     fn next(&mut self) -> Option<Self::Item> {
-        if self.row < self.model.row_count() {
-            let row = self.row;
-            self.row += 1;
-            Some(self.model.row_data(row))
-        } else {
-            None
-        }
+        let row = self.row;
+        self.row += 1;
```
tronical, line 289: From an overflow perspective it seems wrong to increase `self.row` - also given different behavior in debug vs. release. Maybe an overflow is not a realistic scenario?

sixtyfps_runtime/corelib/model.rs
```diff
@@ -459,12 +469,18 @@ impl<T> Model for ModelHandle<T> {
...
     }
 
-    fn row_data(&self, row: usize) -> Self::Data {
-        self.0.as_ref().unwrap().row_data(row)
+    fn row_data(&self, row: usize) -> Option<Self::Data> {
+        if let Some(model) = self.0.as_ref() {
+            model.row_data(row)
+        } else {
+            None
+        }
```
tronical, line 477: Slightly shorter variant perhaps :-)

```suggestion
        self.0.as_ref().and_then(|model| model.row_data(row))
```

sixtyfps_runtime/interpreter/eval.rs
```diff
@@ -180,7 +180,7 @@ pub fn eval_expression(expression: &Expression, local_context: &mut EvalLocalCon
                 (Value::Model(model), Value::Number(index)) => {
                     if (index as usize) < model.row_count() {
                         model.model_tracker().track_row_data_changes(index as usize);
-                        model.row_data(index as usize)
+                        model.row_data(index as usize).unwrap_or(Value::Void)
```
tronical, line 183: Not 100% sure this suggestion compiles, but basically I think we shouldn't return `Void` here.

```suggestion
                        model.row_data(index as usize).unwrap_or(default_value_for_type(expression.ty()))
```

sixtyfps_runtime/interpreter/eval.rs
```diff
@@ -747,7 +747,10 @@ fn eval_assignment(lhs: &Expression, op: char, rhs: Value, local_context: &mut E
                         if op == '=' {
                             model.set_row_data(index, rhs);
                         } else {
-                            model.set_row_data(index, eval(model.row_data(index)));
+                            model.set_row_data(
+                                index,
+                                eval(model.row_data(index).unwrap_or_default()),
```
tronical, line 752: I think this is similar to above, the fallback should be a non-void type:

```suggestion
                                eval(model.row_data(index).unwrap_or_else(|| default_value_for_type(lhs.type())),
```

sixtyfps_runtime/interpreter/eval.rs
hunger: That makes only sense if I also replace the Value::Void two lines down. This unwrap should never trigger anyway as we have a valid row_count here.

tronical: Indeed, so `unwrap()` is probably the right choice then. If the model decides to return `None` despite an in-range row index, then we want the panic to happen right here and not later when we encounter `Value::Void` but expected something else.

sixtyfps_runtime/interpreter/eval.rs
tronical: (Ignore this, in the light of the above discussion :-)

sixtyfps_runtime/corelib/model.rs
```diff
@@ -351,8 +353,12 @@ impl<T: Clone + 'static> Model for VecModel<T> {
...
     }
 
-    fn row_data(&self, row: usize) -> Self::Data {
-        self.array.borrow()[row].clone()
+    fn row_data(&self, row: usize) -> Option<Self::Data> {
+        if row >= self.row_count() {
+            None
+        } else {
+            Some(self.array.borrow()[row].clone())
+        }
```
ogoffart, line 361:
```suggestion
        self.array.borrow().get(row).cloned()
```

sixtyfps_runtime/corelib/model.rs
```diff
@@ -376,8 +382,12 @@ impl Model for usize {
...
     }
 
-    fn row_data(&self, row: usize) -> Self::Data {
-        row as i32
+    fn row_data(&self, row: usize) -> Option<Self::Data> {
+        if row >= self.row_count() {
+            None
+        } else {
+            Some(row as i32)
+        }
```
ogoffart, line 390: Not sure if we really need to return None in that case.
But if yes, may I suggest this shorter form. (can also be used for the one on bool)

```suggestion
        (row >= self.row_count()).then(|| row)
```

sixtyfps_runtime/interpreter/ffi.rs
```diff
@@ -582,11 +582,15 @@ impl Model for ModelAdaptorWrapper {
...
-            v.assume_init()
+    fn row_data(&self, row: usize) -> Option<Value> {
+        if row >= self.row_count() {
+            None
+        } else {
+            unsafe {
+                let mut v = std::mem::MaybeUninit::<Value>::uninit();
+                self.0.row_data(row, v.as_mut_ptr() as *mut ValueOpaque);
+                Some(v.assume_init())
+            }
```
tronical, line 593: This highlights that we probably should do the same change in the C++ Model side, and then the row_data function pointer would indicate to the caller if it returned something or not. So this would probably become the suggestion below, provided that `row_data` returns a `bool` and the C++ Model returns `std::optional`. That could be done in a separate change though?

```suggestion
    fn row_data(&self, row: usize) -> Option<Value> {
        unsafe {
            let mut v = std::mem::MaybeUninit::<Value>::uninit();
            if self.0.row_data(row, v.as_mut_ptr() as *mut ValueOpaque) {
                Some(v.assume_init())
            } else {
                None
            }
        }
```

ogoffart (approved):
Nice :+1:

ogoffart:
Still todo:
 - Document the change in the changelog and in the documentaton
 - Same thing in C++  (or do we not care for C++ because C++ dev are used to check their indices?)
 - Tests

ogoffart:
Regarding the test, i have this test stashed which i did not commit because it did not pass at the time:

`````diff
diff --git a/tests/cases/models/write_to_model.60 b/tests/cases/models/write_to_model.60
index 9b009eb9..a3631564 100644
--- a/tests/cases/models/write_to_model.60
+++ b/tests/cases/models/write_to_model.60
@@ -68,6 +68,10 @@ sixtyfps::testing::send_mouse_click(&instance, 25., 5.);
 assert_eq!(instance.get_clicked_score(), 439);
 assert_eq!(another_model.row_data(2).2, 439.);
 assert_eq!(another_model.row_data(2).1, sixtyfps::SharedString::from("world"));
+
+// FIXME: this panics
+// instance.invoke_manual_score_update(200, 100);
+assert_eq!(another_model.row_count(), 3);
 ```
 
 ```cpp
@@ -93,6 +97,11 @@ sixtyfps::testing::send_mouse_click(&instance, 25., 5.);
 assert_eq(instance.get_clicked_score(), 439);
 assert_eq(std::get<2>(another_model->row_data(2)), 439.);
 assert_eq(std::get<1>(another_model->row_data(2)), "world");
+
+// FIXME: this segfault
+// instance.invoke_manual_score_update(200, 100);
+assert_eq(another_model->row_count(), 3);
+
 ```
 
 
@@ -117,6 +126,9 @@ instance.send_mouse_click(25., 5.);
 assert.equal(instance.clicked_score, 439);
 assert.equal(another_model.rowData(2).score, 439.);
 assert.equal(another_model.rowData(2).name, "world");
+
+instance.manual_score_update(200, 100); // should do nothing
+assert.equal(another_model.length, 3);
 ```
 
 

`````

File: api/sixtyfps-node/native/js_model.rs [hunk 8]
```diff
@@ -64,22 +64,27 @@ impl Model for JsModel {
...
+                    .ok()
+                    .and_then(|func| func.downcast::<JsFunction>().ok())
+                    .and_then(|func| func.call(cx, obj, std::iter::once(row)).ok())
+                    .and_then(|res| {
+                        crate::to_eval_value(res, self.data_type.clone(), cx, persistent_context)
+                            .ok()
+                    })
+                    .map(|res| r.set(res));
+            });
+            Some(r.into_inner())
```
tronical, line 86: Similar to the C++ suggestion below, for JavaScript we should probably change the `Model` interface in `index.ts` to use `rowData(row: number): T | undefined;` and allow the JS model to do the check.

sixtyfps_runtime/corelib/model.rs
hunger: I really like this one!

CHANGELOG.md
```diff
@@ -10,6 +10,7 @@ This version changes some APIs in incompatible ways. For details how to migrate
  - Minimum rust version is now 1.56
  - C++ compiler requires C++20
  - In the C++ interpreter API `std::span` is used for callbacks arguments, instead of `sixtyfps::Slice`
+ - In rust `Model::row_data(...)` will now return a `Option<T>` instaed of a plain `T`
```
tronical, line 13:
```suggestion
 - In rust `Model::row_data(...)` will now return a `Option<T>` instead of a plain `T`
```

docs/rust_migration.md
```diff
@@ -0,0 +1,27 @@
+# Migrating from Older Versions
+
+The rust library is versioned according to the principles of [Semantic Versioning](https://semver.org). We define that the left-most non-zero component of the version is the major version, followed by the minor and optionally patch version. That means releases in the "0.y.z" series treat changes in "y" as a major release, which can contain incompatible API changes, while changes in just "z" are minor. For example the release 0.1.6 is fully backwards compatible to 0.1.5, but it contains new functionality. The release 0.2.0 however is a new major version compared to 0.1.x and may contain API incompatible changes.
+
+This guide lists all API incompatible changes between major versions and describes how you can migrate your application's source code.
+
+## Migrating from Version 0.1.x to 0.2.0
+
+In the 0.2.x series we have increased the minimum version of rust. You need to have rust compiler version >= 1.56 installed.
```
ogoffart, line 9: this seems to assume that we will not raise it again in the 0.2.x series.
I'd leave that out, or use similar phrasing than in C++

docs/rust_migration.md
```diff
@@ -0,0 +1,27 @@
...
+
+### Rust API
+
+#### Models
+
+`Model::row_data` now returns an `Option<T>` instead of a simple `T`.
+
+Old code:
+
+```rust
```
ogoffart, line 19: this won't compile

````suggestion
```rust,ignore
````

File: CHANGELOG.md [hunk 12]
```diff
@@ -10,6 +10,7 @@ This version changes some APIs in incompatible ways. For details how to migrate
  - Minimum rust version is now 1.56
  - C++ compiler requires C++20
  - In the C++ interpreter API `std::span` is used for callbacks arguments, instead of `sixtyfps::Slice`
+ - In rust `Model::row_data(...)` will now return a `Option<T>` instead of a plain `T`
```
ogoffart, line 13:
```suggestion
 - `Model::row_data` will now return a `Option<T>` / `std::optional<T>` instead of a plain `T`. 
```

docs/rust_migration.md
hunger: The C++ guide says this:

> In the 0.2.x series we have increased the minimum version of C++. You need to have a C++ compiler installed that supports C++ 20 or newer.

Does that need to get updated as well?

I do get your point though and will change this to read 0.2.0.

hunger: I'll adapt it as I add the API change for C++.

docs/rust_migration.md
```diff
@@ -0,0 +1,27 @@
...
+
+## Migrating from Version 0.1.x to 0.2.0
+
+In the 0.2.x series we have increased the minimum version of rust. You need to have rust compiler version >= 1.56 installed.
+
+### Rust API
+
+#### Models
+
+`Model::row_data` now returns an `Option<T>` instead of a simple `T`.
```
ogoffart, line 15: It is also importent to doucment that set_row_data and row_data must no longer panic, for those who implemented Model.

docs/rust_migration.md
```diff
@@ -0,0 +1,27 @@
...
+
+Old code:
+
+```rust,ignore
+let row_five = model.row_data(5);
+```
+
+New code:
+
+```rust
```
ogoffart, line 25:
````suggestion
```rust,ignore
````

File: examples/imagefilter/main.rs [hunk 15]
```diff
@@ -64,8 +64,8 @@ impl sixtyfps::Model for Filters {
         self.0.len()
     }
 
-    fn row_data(&self, row: usize) -> Self::Data {
-        self.0[row].name.clone()
+    fn row_data(&self, row: usize) -> Option<Self::Data> {
+        (row < self.row_count()).then(|| Some(self.0[row].name.clone()))
```
ogoffart, line 68: Does that compile?

Anyway, you can use .get
```suggestion
        self.0.get().map(|x| x.name.clone())
```

sixtyfps_runtime/corelib/model.rs
```diff
@@ -396,7 +400,13 @@ impl Model for bool {
         }
     }
 
-    fn row_data(&self, _row: usize) -> Self::Data {}
+    fn row_data(&self, row: usize) -> Option<Self::Data> {
+        if row >= self.row_count() {
+            None
+        } else {
+            Some(())
+        }
```
ogoffart, line 408: (row < self.row_count()).then(|| ())

sixtyfps_runtime/interpreter/ffi.rs
```diff
@@ -582,11 +583,15 @@ impl Model for ModelAdaptorWrapper {
         self.0.row_count()
     }
 
-    fn row_data(&self, row: usize) -> Value {
+    fn row_data(&self, row: usize) -> Option<Value> {
+        println!("Row data({})", row);
```
ogoffart, line 587: leftover
```suggestion
```

sixtyfps_runtime/interpreter/ffi.rs
```diff
@@ -595,8 +600,11 @@ impl Model for ModelAdaptorWrapper {
     }
 
     fn set_row_data(&self, row: usize, data: Value) {
-        let val: &ValueOpaque = unsafe { std::mem::transmute::<&Value, &ValueOpaque>(&data) };
-        self.0.set_row_data(row, val);
+        println!("Set row data({})", row);
```
ogoffart, line 603:
```suggestion
```

ogoffart (commented):
In order to ease the review, can i ask you not to rebase but add the commit to the review?
So then we can see the new commits.
One can rebase the commit later.

api/sixtyfps-node/native/js_model.rs
hunger: I can look into this, but the tests @ogoffart suggested work without this, so I'll keep that for an extra patch. This one is too big already.

ogoffart: my tests are written in Javascript, not TypeScript

Yeah, we don't have typescript tests :-s

hunger:
@ogoffart: Oh, cool, I totally missed those tests so far! I will need to look into those.

ogoffart:
And speaking of tests, we also should test that in .60, a [] to an invalid expression does not panic.
maybe a test like this:

```
property <[{a: string}]> foo;
property<bool> test:  foo[456456].a == "";
```

api/sixtyfps-node/native/js_model.rs
tronical: > I can look into this, but the tests @ogoffart suggested work without this, so I'll keep that for an extra patch. This one is too big already.

Sounds good.

The `write_to_model.60` test in the `js` section calls for example `assert.equal(another_model.rowData(2).score, 439.);`.

If we changed `rowData` also in JS to return `undefined` for an out-of-bounds row, then this could be extended here.

api/sixtyfps-cpp/docs/cpp_migration.md
```diff
@@ -6,7 +6,7 @@ This guide lists all API incompatible changes between major versions and describ
 
 ## Migrating from Version 0.1.x to 0.2.0
 
-In the 0.2.x series we have increased the minimum version of C++. You need to have a C++ compiler installed that supports C++ 20 or newer.
+In 0.2.0 we have increased the minimum version of C++. You need to have a C++ compiler installed that supports C++ 20 or newer.
```
tronical, line 9:
```suggestion
In version 0.2.0 we have increased the minimum version of C++. You need to have a C++ compiler installed that supports C++ 20 or newer.
```

api/sixtyfps-cpp/docs/cpp_migration.md
```diff
@@ -29,3 +29,19 @@ New code:
 sixtyfps::Value args[] = { SharedString("Hello"), 42. };
 instance->invoke_callback("foo", args);
 ```
+
+#### Models
+
+`Model::row_data` will now return a `std::optional<ModelData>` and can thus be used with indices that are out of bounds.
```
tronical, line 35:
```suggestion
`Model::row_data` returns now a `std::optional<ModelData>` and can thus be used with indices that are out of bounds.
```

api/sixtyfps-cpp/docs/cpp_migration.md
```diff
@@ -29,3 +29,19 @@ New code:
...
+Old code:
+
+```cpp
+std::get<2>(another_model->row_data(2)), 336.);
+```
+
+New code:
+
+```cpp
+std::get<2>(*another_model->row_data(2)), 336.); // Note the extra *. You should check whether the returned value `has_value()` before accessing it
```
tronical, line 46:
```suggestion
```cpp
// `another_model` is a model that contains floats.
std::optional<float> entry = another_model->row_data(2);
if (entry.has_value()) {
    do_something(*entry);
} else {
    // row index 2 is out of bounds
}
```

docs/rust_migration.md
hunger: I updated that as well as the C++ docs which I copied the text from.

I needed to touch the C++ migration guide anyway to add the C++ side of things there.

CHANGELOG.md
```diff
@@ -10,6 +10,8 @@ This version changes some APIs in incompatible ways. For details how to migrate
  - Minimum rust version is now 1.56
  - C++ compiler requires C++20
  - In the C++ interpreter API `std::span` is used for callbacks arguments, instead of `sixtyfps::Slice`
+ - `Model::row_data` will now return a `Option<T>` / `std::optional<T>` instead of a plain `T`. 
+ - In C++ `Model::row_data(...)` will now return a `std::optional<ModelData>` instead of a plain `ModelData`
```
ogoffart, line 14:
```suggestion
 - `Model::row_data` will now return a `Option<T>` / `std::optional<T>` instead of a plain `T`. 
```

This was now dupliacated :-)

docs/rust_migration.md
```diff
@@ -0,0 +1,29 @@
+# Migrating from Older Versions
```
ogoffart, line 1: I would rather put this file in api/sixtyfps-rs/micration.md

sixtyfps_runtime/corelib/model.rs
```diff
@@ -170,8 +170,12 @@ impl ModelTracker for ModelNotify {
...
 ///     }
 ///
-///     fn row_data(&self, row: usize) -> Self::Data {
-///         self.array.borrow()[row].clone()
+///     fn row_data(&self, row: usize) -> Option<Self::Data> {
+///         if row >= self.row_count() {
+///             None
+///         } else {
+///             Some(self.array.borrow()[row].clone())
+///         }
```
ogoffart, line 178:
```suggestion
///         self.array.borrow().get(row).cloned()
```

sixtyfps_runtime/interpreter/value_model.rs
```diff
@@ -62,13 +62,17 @@ impl Model for ValueModel {
...
-            x => panic!("Invalid model {:?}", x),
+    fn row_data(&self, row: usize) -> Option<Self::Data> {
+        if row >= self.row_count() {
+            None
+        } else {
+            Some(match &*self.value.borrow() {
+                Value::Bool(_) => Value::Void,
+                Value::Number(_) => Value::Number(row as _),
+                Value::Array(a) => a[row].clone(),
+                Value::Model(model_ptr) => model_ptr.row_data(row).unwrap_or(Value::Void),
```
ogoffart, line 73:
```suggestion
                Value::Model(model_ptr) => model_ptr.row_data(row)?,
```
If the inner moder return None for some reason, we must should probably propagate that.

api/sixtyfps-cpp/docs/cpp_migration.md
```diff
@@ -29,3 +29,27 @@ New code:
...
+#### Models
+
+`Model::row_data` returns now a `std::optional<ModelData>` and can thus be used with indices that are out of bounds.
+
+This also means that `Model`s must handle invalid indices and may not crash when a invalid index is passed in.
+
+Old code:
+
+```cpp
+another_model->row_data(2);
```
tronical, line 42: I think it would be a good idea to use the return value in a way that is similar to the "new" code.

```suggestion
float value = another_model->row_data(2);
do_something(value);
```

tronical:
Nice! This is shaping up to be a very good API improvement IMO.

examples/imagefilter/main.rs
```diff
@@ -64,8 +64,8 @@ impl sixtyfps::Model for Filters {
         self.0.len()
     }
 
-    fn row_data(&self, row: usize) -> Self::Data {
-        self.0[row].name.clone()
+    fn row_data(&self, row: usize) -> Option<Self::Data> {
+        self.0.get().map(|x| x.name.clone())
```
tronical, line 68:
```suggestion
        self.0.get(row).map(|x| x.name.clone())
```

sixtyfps_compiler/generator/cpp.rs
```diff
@@ -1768,7 +1768,7 @@ fn compile_expression(expr: &llr::Expression, ctx: &EvaluationContext) -> String
             _ => panic!("Expression::ObjectAccess's base expression is not an Object type"),
         },
         Expression::ArrayIndex { array, index } => {
-            format!("[&](const auto &model, const auto &index){{ model->track_row_data_changes(index); return model->row_data(index); }}({}, {})", compile_expression(array, ctx), compile_expression(index, ctx))
+            format!("[&](const auto &model, const auto &index){{ model->track_row_data_changes(index); const auto v = model->row_data(index); if (v.has_value()) return *v; else return {{}}; }}({}, {})", compile_expression(array, ctx), compile_expression(index, ctx))
```
tronical, line 1771: It's fascinating that this works nowadays in C++, without the need to specify the return type - especially the `{}` part. Very cool :)

hunger: Yes, I was also rather surprised by that.

But let's see: Maybe the CI will uncover that I did something wrong there:-)

tronical (approved):

sixtyfps_compiler/generator/cpp.rs
hunger: See? Ci says it does not work:-/ I wonder why it did for me... maybe I used a different compiler?

tronical: Maybe this could be done by "re-exporting" the type parameter for `Model` as `template <typename ModelData> class Model { public: typedef ModelData ValueType; ... }` and then use `decltype(model)::Valuetype` ?

Or perhaps this could be a lambda with explicit template parameters? `[&]<ModelData>(const sixtyfps::Model<ModelData> &model, const auto &index) -> ModelData { ... }` ?

ogoffart, line 1771:
```suggestion
            format!("[&](const auto &model, const auto &index){{ model->track_row_data_changes(index); if (const auto v = model->row_data(index)) return *v; return {{}}; }}({}, {})", compile_expression(array, ctx), compile_expression(index, ctx))
```

ogoffart: actually:

```suggestion
            format!("[&](const auto &model, const auto &index){{ model->track_row_data_changes(index); return model->row_data(index).value_or({{}}); }}({}, {})", compile_expression(array, ctx), compile_expression(index, ctx))
```

sixtyfps_runtime/interpreter/ffi.rs
```diff
@@ -595,8 +599,10 @@ impl Model for ModelAdaptorWrapper {
     }
 
     fn set_row_data(&self, row: usize, data: Value) {
-        let val: &ValueOpaque = unsafe { std::mem::transmute::<&Value, &ValueOpaque>(&data) };
-        self.0.set_row_data(row, val);
+        if row < self.row_count() {
```
ogoffart, line 602: Why do we need this check?

hunger: You are right, that is probably not needed.

Issue #431, opened 2021-08-24
Breaking API change to do in next version


tronical:
- [x] Change `sixtyfps::Weak<T>::upgrade_in_event_loop` to take `&self` instead of `self`:
       This makes the API more intuitive to use by not requiring the caller to call `clone()` each time.
- [ ] In sixtyfps_interpreter: add `#[non_exhaustive]` to all the error enum.
       Important so we can extend the `CallCallbackError` with error if the arguments of the function are of the wrong type
- [x] Also rename `sixtyfps_interpreter::CallCallbackError` to `InvokeCallbackError`
- [x] `Image::size`  should return an integer size
- [x] Consider removing `Value::Array` (interpreter is not creating it anymore)

PR #878, opened 2022-01-27
Make `sixtyfps::Weak<T>::upgrade_in_event_loop` easier to use

tronical:
... by taking &self instead of self

cc #431

### DIFF
--- a/api/sixtyfps-rs/lib.rs
+++ b/api/sixtyfps-rs/lib.rs
@@ -653,12 +653,13 @@ mod weak_handle {
         /// handle.run();
         /// ```
         #[cfg(feature = "std")]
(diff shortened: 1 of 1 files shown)


ogoffart (approved):
Would be nice to have at least one test that uses that fact.

tronical:
I'll land this but see about keeping the test in mind for #881, although that might prove to the difficult if the `Weak` code ends up in corelib, where in tests we can't run the event loop :-/

ogoffart:
> corelib, where in tests we can't run the event loop

I wonder if we can depend on the testing backend as a dev-dependency
And there we should have a mock eventloop.

But the "test" can also be a documentation example in the sixtyfps-rs crate.

PR #879, opened 2022-01-27
Remove deprecated syntax from tests

hunger:


hunger:
I think this should not have any influence on the tests themselves.

tronical (approved):

ogoffart (approved):
It's fine to remove it from tests/cases.  But obviously not from the compiler/tests/syntax.
(there they will become actual error that need to be tested later)

PR #880, opened 2022-01-27
Remove Value::Array in the interpreter

ogoffart:


sixtyfps_runtime/corelib/model.rs
```diff
@@ -269,6 +269,12 @@ pub struct ModelIterator<'a, T> {
     row: usize,
 }
 
+impl<'a, T> ModelIterator<'a, T> {
+    pub fn new(model: &'a dyn Model<Data = T>) -> Self {
```
tronical, line 273: This will likely cause a warning/error about missing docs :)

tronical:
Ahh, I think I see now how you're mapping this to C++ - nice :)

I think this needs tweaks to the docs though. From the `Value` docs:

```cpp
    /// Returns a std::optional that contains a vector of values if the type of this Value is
    /// Type::Array, otherwise an empty optional is returned.
    inline std::optional<sixtyfps::SharedVector<Value>> to_array() const;
```

This should perhaps explain that only a non-null variant is returned *if* the model was initialised from an array in `.60`. Maybe

```
/// Returns a std::optional that contains a vector of values if the type of this Value is
/// Type::Model and the Value was initialised from an array in `.60` markup; otherwise
/// an empty optional is returned.
```

It might also be worth referring to this in the `Value` class docs. What do you think?

tronical (approved):

ogoffart:
(cc  #431)

PR #881, opened 2022-01-27
Rust Interpreter API cleanup

tronical:
Migration docs are a known TODO, but I'd like to see first if this is the right direction - especially since it changes the dependencies between crates.


sixtyfps_runtime/corelib/api.rs
```diff
@@ -0,0 +1,272 @@
...
+///     sixtyfps::invoke_from_event_loop(move || handle_copy.unwrap().set_foo(foo));
+/// });
+/// # thread.join().unwrap(); return; // don't run the event loop in examples
+/// handle.run();
+/// ```
+pub fn invoke_from_event_loop(func: impl FnOnce() + Send + 'static) {
+    if let Some(backend) = crate::backend::instance() {
+        backend.post_event(alloc::boxed::Box::new(func))
+    } else {
+        panic!("sixtyfps::invoke_from_event_loop() must be called after starting the event loop.")
```
tronical, line 270: This is new and strictly speaking a behavioural change. When `invoke_from_event_loop` was in the API crate, we could access the default backend, which would be implicitly created. With this function here - and it needs to be in order for `Weak` to call it - we can't do that anymore.

I'm inclined to say that's an acceptable change, but what do you think?

ogoffart:
```suggestion
        panic!("sixtyfps::invoke_from_event_loop() must be called after the SixtyFPS backend is initialized.")
```

This is unfortunate, but i think it's probably fine because creating any component will also initialize the backend. and this function is most usefull when component is already created.

ogoffart (approved):

sixtyfps_runtime/rendering_backends/testing/lib.rs
```diff
@@ -31,7 +31,7 @@ impl sixtyfps_corelib::backend::Backend for TestingBackend {
     }
 
     fn run_event_loop(&'static self, _behavior: sixtyfps_corelib::backend::EventLoopQuitBehavior) {
-        unimplemented!("running an event loop with the testing backend");
+        // Nothing to spin, we return right away. Otherwise tests will hang.
```
ogoffart, line 34: I think it's better to panic because we are not testing things properly.  

For example, we vould have an event loop that just spings and pass mock times for so so many iterations.
And it should be assumed that the test has setup a timer or posted events.

That's why if we panic, we know that this is not yet implemented and not wonder why the eventloop don't run with test and how to fix it.

sixtyfps_runtime/corelib/Cargo.toml
```diff
@@ -72,3 +72,5 @@ wasm-bindgen = { version = "0.2" }
 pin-weak = "1"
 image = { version = "0.23.14", default-features = false, features = [ "png" ] }
 tiny-skia = "0.6.1"
+sixtyfps = { version = "=0.2.0", path = "../../api/sixtyfps-rs", default-features = false, features = ["std"] }
+sixtyfps-rendering-backend-testing = { version = "=0.2.0", path="../rendering_backends/testing" }
```
ogoffart, line 76:
```suggestion
sixtyfps-rendering-backend-testing = { path="../rendering_backends/testing" }
```

Otherwise we can't push to crates.io  :-)

sixtyfps_runtime/corelib/api.rs
```diff
@@ -0,0 +1,276 @@
...
+    }
+}
+
+impl crate::window::WindowHandleAccess for Window {
+    fn window_handle(&self) -> &Rc<crate::window::Window> {
+        &self.0
+    }
+}
+/// This trait is used to obtain references to global singletons exported in `.60`
+/// markup. Alternatively, you can use [`ComponentHandle::global`] to obtain access.
```
ogoffart, line 44: I hope this will still work properly.

sixtyfps_runtime/corelib/api.rs
```diff
@@ -0,0 +1,276 @@
...
+    fn clone_strong(&self) -> Self;
+
+    /// Internal function used when upgrading a weak reference to a strong one.
+    #[doc(hidden)]
+    fn from_inner(_: vtable::VRc<ComponentVTable, Self::Inner>) -> Self;
+
+    /// Marks the window of this component to be shown on the screen. This registers
+    /// the window with the windowing system. In order to react to events from the windowing system,
+    /// such as draw requests or mouse/touch input, it is still necessary to spin the event loop,
+    /// using [`crate::run_event_loop`].
```
ogoffart, line 106: I wonder if this will still work properly.
Will it work if you do `[sixtyfps::run_event_loop`].  I don't think so.

ogoffart (approved):
It helps to use git locally with the color for moved lines...

sixtyfps_runtime/corelib/api.rs
tronical: The `global_accessor_api.60` tests use this in the Rust section, so I think we have test coverage:

```rust
instance.global::<PublicGlobal>().set_hello(100);
```

sixtyfps_runtime/corelib/api.rs
tronical: Indeed, this link and the others are broken. /me digs

tronical: I have not been able to solve this without either

   (1) Having the real implementation of `run_event_loop` in corelib, but then it would `panic!` if no backend was initialised yet.

    (2) using a HTML link (`fn.run_event_loop.html`)

Hacks using `#[cfg(doc)]` with conditional exporting in `corelib` didn't work because they were not in place anymore when generating docs for the API crate.

So I've opted for (2) for now.

sixtyfps_runtime/rendering_backends/testing/lib.rs
tronical: Ok, I've reverted that and in exchange added a `no_run` to one of the Timer doc tests that ended up calling it.

ogoffart (approved):

PR #884, opened 2022-01-27
Add compatibility feature to our public crates

ogoffart:
We want to be able to put existing functionality behind a feature flag while keeping
the semver compatibility.
This is only possible if that new feature flag is enabled by default, but this is not
working if the users have done `default-features = false` in their Cargo.toml.
So we add new `compat-x-y-z` feature that is mandatory to have and which is
enforced with a `compile_error!`

Now, users that whishes to not have the default features must enable it explicitly.
Say we want only x11 but not qt and wayland, the user will do
```toml
sixtyfps = { version = "0.2", default-features = false, features = ["x11", "compat-0-2-0"] }
```

Now, imagine that in the version 0.2.3, we put the SVG support behind a feature flag.
we will do this in out Cargo.toml:

```toml
[features]
default = ["compat-0-2-0", "x11", "wayland"]
compat-0-2-0 = ["compat-0-2-3", "svg"]
compat-0-2-3 = []

svg = [...]
...
```

That way, the svg feature will be enabled by default for all the users who used previous version
of SixtyFPS, and people that want to disable "svg" can just change from compat-0-2-0 to
compat-0-2-3 in their Cargo.toml

### DIFF
--- a/api/sixtyfps-cpp/Cargo.toml
+++ b/api/sixtyfps-cpp/Cargo.toml
@@ -33,7 +33,7 @@ default = ["backend-gl", "x11", "backend-qt"]
 
 [dependencies]
 sixtyfps-corelib = { version = "=0.2.0", path="../../internal/core", features = ["ffi"] }
-sixtyfps-interpreter = { version = "=0.2.0", path="../../internal/interpreter", default-features = false, features = ["ffi"], optional = true }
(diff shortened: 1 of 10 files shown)


tronical (approved):
Very nice idea!

I think a typical way people would run into this is by checking what the features are (something we could also document, but let's assume they look it up in the sources). So then they would write `default-features = false, features = ["backend-qt"]` and then they'd get a compile error.

Now if then that compile error would refer to the docs that explain the entire concept, I think it would be easier to understand for users why they need to add `compat-0-2-0` to the list of features.

PR #887, opened 2022-01-28
Transform some warnings into errors

ogoffart:


ogoffart:
Notably, I kept `Rectangle::color` and `minimum-*` `maximum-*`.  I think it's good to keep as deprecated in case user reach for it, then they'll get a message telling them to rename to the right one.
What do you think?

ogoffart:
Ah, i pushed in the wrong branch by mistake again...  Sorry.
I think i should change my workflow and not commit on master localy.

tronical (commented):
post-merge 👍  :)

PR #889, opened 2022-01-28
Mark some enums that end up in the public Rust API as non_exhaustive

tronical:
This allows adding new variants in minor releases without breaking
source compatibility.

cc #431

### DIFF
--- a/sixtyfps_runtime/interpreter/api.rs
+++ b/sixtyfps_runtime/interpreter/api.rs
@@ -21,6 +21,7 @@ pub use sixtyfps_corelib::window::api::Window;
 /// the contained values.
 #[derive(Debug, Copy, Clone, PartialEq)]
 #[repr(i8)]
+#[non_exhaustive]
 pub enum ValueType {
     /// The variant that expresses the non-type. This is the default.
     Void,
@@ -1032,6 +1033,7 @@ impl WeakComponentInstance {
 
 /// Error returned by [`ComponentInstance::get_property`]
 #[derive(Debug, Clone, Copy, PartialEq, Eq, thiserror::Error)]
+#[non_exhaustive]
 pub enum GetPropertyError {
     /// There is no property with the given name
     #[error("no such property")]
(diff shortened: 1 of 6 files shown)


tronical:
[`DiagnosticsLevel`](https://sixtyfps.io/snapshots/master/docs/rust/sixtyfps_interpreter/enum.diagnosticlevel) is not marked as non_exhaustive. We match this ourselves in the lsp, for example - making it non_exhaustive would mean we don't get an error ourselves anymore in our code base.

Three options:

    (1) Accept that (and use that warning once it stabilises)
    (2) Separate public and private API (a bit of work, but ok)
    (3) Keep it non_exhaustive


I'm leaning towards (1). What do you think?

sixtyfps_runtime/interpreter/api.rs
```diff
@@ -21,6 +21,7 @@ pub use sixtyfps_corelib::window::api::Window;
 /// the contained values.
 #[derive(Debug, Copy, Clone, PartialEq)]
 #[repr(i8)]
+#[non_exhaustive]
```
ogoffart, line 24: maybe the `Other` should be `#[doc(hidden)]` and handle with the `_` pattern anyway?

sixtyfps_runtime/interpreter/api.rs
```diff
@@ -1059,6 +1063,7 @@ pub enum SetCallbackError {
 
 /// Error returned by [`ComponentInstance::invoke_callback`]
 #[derive(Debug, Clone, Copy, PartialEq, Eq, thiserror::Error)]
+#[non_exhaustive]
```
ogoffart, line 1066: In a folloup patch, we can add the error if arguments type are incorrect

ogoffart (approved):

ogoffart:
> DiagnosticsLevel is not marked as non_exhaustive.

Option 1.  accept that.

sixtyfps_runtime/interpreter/api.rs
tronical: 👍

sixtyfps_runtime/interpreter/api.rs
tronical: 👍 Please do

PR #891, opened 2022-01-28
janitor: Rename some enum values

hunger:


hunger:
@tronical How attached are you to these names? Clippy does not like them repeating a part that is already in the enum name, which I think makes sense.

This is not C++ where enum values were (used to be) defined in the scope above the enum itself.

tronical (approved):
That’s fine with me :)

PR #892, opened 2022-01-28
Semi-automatic clippy fixes

hunger:
Nothing interesting, I mostly want to let CI run over it.

Afterwards we have some more warnings about functions with too many arguments, the cpp transmute thing that I should fix upstream, some complex types. But apart from those we are almost clippy clean after this set of patches and #794.

### DIFF
--- a/internal/backends/testing/lib.rs
+++ b/internal/backends/testing/lib.rs
@@ -77,7 +77,7 @@ impl sixtyfps_corelib::backend::Backend for TestingBackend {
             )
             .map(|img| img.dimensions().into())
             .unwrap_or_default(),
-            ImageInner::StaticTextures { size, .. } => size.clone(),
+            ImageInner::StaticTextures { size, .. } => *size,
         }
     }
 }
--- a/internal/compiler/generator.rs
+++ b/internal/compiler/generator.rs
@@ -188,7 +188,7 @@ pub fn build_item_tree<T: ItemTreeBuilder>(
 
     fn visit_children<T: ItemTreeBuilder>(
         state: &T::SubComponentState,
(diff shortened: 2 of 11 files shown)


ogoffart:
the second commit (more clippy stuff) looks alright.  
I'm a bit torn about the stray @.  I think it's usefull to see that this is indeed a binding and be explicit.
consider:
```rust
   x => { ... }
   x @ _ => { ... }
```
In the first case, x could also be a constant.
I think having the @ make it clear that this is a binding.  
What do you think?

sixtyfps_compiler/generator/cpp.rs
```diff
@@ -634,21 +634,18 @@ fn generate_public_component(file: &mut File, component: &llr::PublicComponent)
...
 
     // Give generated sub-components, etc. access to our fields
 
     for new_decl in file.declarations.iter().skip(old_declarations) {
-        match new_decl {
-            Declaration::Struct(struc @ Struct { .. }) => {
-                component_struct.friends.push(struc.name.clone());
-            }
-            _ => {}
+        if let Declaration::Struct(struc @ Struct { .. }) = new_decl {
```
ogoffart, line 647:
```suggestion
        if let Declaration::Struct(struc) = new_decl {
```

Paradoxically, i think this stray @ has little value... I guess i'm not consistent

hunger:
@ogoffart consistency is overrated anyway:-)

For me "clippy compliance" is a value in itself, simply because I consider crates with hundreds of warnings not very trustworthy and try to avoid those.

On the other hand, I tend to get suspicious on crates without warnings: Did they just do stupid work-arounds to make clippy happy or is the code just good?

PR #901, opened 2022-01-30
ModelRc: Remove some more unnecessary derives

hunger:
The contents is not longer accessible, so there is no need for deref and
deref_mut.

@ogoffart pointed that out in the commit that introduced the ModelRc.


ogoffart (approved):
👍 
> The contents is not longer accessible, so there is no need for deref and deref_mut.

This is not entirely accurate. Deref *makes* the contents accessible in the public API and we don't want that (for example, it caused problem because .iter() of the Option was taken instead of the one of Model if one does not `use Model`)

PR #903, opened 2022-01-31
Pr/code move

hunger:
Move code into the `internal` folder.


ogoffart (approved):
:+1:

(It helps to review that from the command line interface.  GitHub sucks at displaying diff where lots of lines or files moved)

ogoffart:
I'm still unsure about the "-selector" backend, but we don't have better name, i guess.

PR #908, opened 2022-02-01
Custom rendering API

tronical:
This is the initial merge request for supporting underlay and overlay rendering in Rust and C++.

There are a few caveats:

* Explicit backend and native style selection needs to be implemented for Rust and C++.
  * For C++ I think the style could simply be a string target property and we call back to the "default" (which is not configurable right now, but that's a separate issue).
  * For Rust and C++ we need the ability to select a backend. This could be a "one-time" function `mod sixtyfps { pub fn set_backend(name: &'str) -> std::result::Result<(), String>; `
* There's a hack in the `request_redraw()` API :-(
* The first change in the series adds support for alpha in the Window surface, so in the most extreme case you can write `background: transparent;` on a `Window` and it'll make the background shine through. This works on macOS, Wayland and Windows with the GL backend for me, but somehow not with X11. Also it is not implemented with the Qt backend. This also means that the C++ overlay alpha mask example doesn't work on X11 (not on Windows anyway because of GLES2 headers). I'm not sure how far to go here. Should I try to implement it for Qt? We know that the use-case is limited right now.

The graphics exposure API I tried to keep (a) free of public dependencies (b) flexible for future expansion away from OpenGL and (b) with a life-time attached so that we *can* pass temporaries from the backend to the user-code.

Edit: TODO:

 - [x] Provide a return value for `set_rendering_notifier` that allows error detection. This also captures the backend selection part then.
 - [x] Make the allocation of the ARGB surface an opt-in. This does come at the expense of memory and strictly speaking also GPU performance, so we should make it opt-in for the rare case that it's needed instead of always paying the price for it.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -29,6 +29,7 @@ as well as the [Rust migration guide for the `sixtyfps` crate](api/sixtyfps-rs/m
 
  - `TextEdit::font-size` and `LineEdit::font-size` have been added to control the size of these widgets.
  - Added `sixtyfps::Window::set_rendering_notifier` to get a callback before and after a new frame is being rendered.
+ - Added `sixtyfps::Window::request_redraw()` to schedule redrawing of the window contents.
 
 ### Fixed
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -28,6 +28,7 @@ as well as the [Rust migration guide for the `sixtyfps` crate](api/sixtyfps-rs/m
 ### Added
 
  - `TextEdit::font-size` and `LineEdit::font-size` have been added to control the size of these widgets.
+ - Added `sixtyfps::Window::set_rendering_notifier` to get a callback before and after a new frame is being rendered.
 
 ### Fixed
 
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -14,6 +14,7 @@ members = [
     'examples/gallery',
     'examples/imagefilter',
     'examples/memory',
+    'examples/opengl_underlay',
     'examples/plotter',
     'examples/printerdemo_old/rust',
     'examples/printerdemo/rust',
--- a/api/cpp/cbindgen.rs
+++ b/api/cpp/cbindgen.rs
@@ -268,6 +268,7 @@ fn gen_corelib(
             "sixtyfps_windowrc_set_component",
             "sixtyfps_windowrc_show_popup",
             "sixtyfps_windowrc_set_rendering_notifier",
+            "sixtyfps_windowrc_request_redraw",
             "sixtyfps_new_path_elements",
             "sixtyfps_new_path_events",
             "sixtyfps_color_brighter",
--- a/api/cpp/cbindgen.rs
+++ b/api/cpp/cbindgen.rs
@@ -128,6 +128,9 @@ fn gen_corelib(
     .map(|x| x.to_string())
     .collect();
 
+    let mut private_exported_types: std::collections::HashSet<String> =
+        config.export.include.iter().cloned().collect();
+
     config.export.exclude = [
         "SharedString",
         "SharedVector",
@@ -157,8 +160,10 @@ fn gen_corelib(
         "sixtyfps_color_darker",
         "sixtyfps_image_size",
         "sixtyfps_image_path",
-        "TimerMode", // included in generated_public.h
-        "IntSize",   // included in generated_public.h
+        "TimerMode",                 // included in generated_public.h
+        "IntSize",                   // included in generated_public.h
+        "RenderingState",            // included in generated_public.h
+        "SetRenderingNotifierError", // included in generated_public.h
     ]
     .iter()
     .map(|x| x.to_string())
@@ -201,6 +206,7 @@ fn gen_corelib(
         .insert("StateInfo".to_owned(), "    using Instant = uint64_t;".into());
     properties_config.structure.derive_eq = true;
     properties_config.structure.derive_neq = true;
+    private_exported_types.extend(properties_config.export.include.iter().cloned());
     cbindgen::Builder::new()
         .with_config(properties_config)
         .with_src(crate_dir.join("properties.rs"))
@@ -261,6 +267,7 @@ fn gen_corelib(
             "sixtyfps_windowrc_set_focus_item",
             "sixtyfps_windowrc_set_component",
             "sixtyfps_windowrc_show_popup",
+            "sixtyfps_windowrc_set_rendering_notifier",
             "sixtyfps_new_path_elements",
             "sixtyfps_new_path_events",
             "sixtyfps_color_brighter",
@@ -289,6 +296,9 @@ fn gen_corelib(
         // Property<> fields uses the public `sixtyfps::Blah` type
         special_config.namespaces =
             Some(vec!["sixtyfps".into(), "cbindgen_private".into(), "types".into()]);
+
+        private_exported_types.extend(special_config.export.include.iter().cloned());
+
         cbindgen::Builder::new()
             .with_config(special_config)
             .with_src(crate_dir.join("graphics.rs"))
@@ -309,8 +319,15 @@ fn gen_corelib(
     let mut public_config = config.clone();
     public_config.namespaces = Some(vec!["sixtyfps".into()]);
     public_config.export.item_types = vec![cbindgen::ItemType::Enums, cbindgen::ItemType::Structs];
-    public_config.export.include = vec!["TimerMode".into(), "IntSize".into()];
-    public_config.export.exclude.clear();
+    // Previously included types are now excluded (to avoid duplicates)
+    public_config.export.exclude = private_exported_types.into_iter().collect();
+    public_config.export.exclude.push("Point".into());
+    public_config.export.include = vec![
+        "TimerMode".into(),
+        "IntSize".into(),
+        "RenderingState".into(),
+        "SetRenderingNotifierError".into(),
+    ];
 
     public_config.export.body.insert(
         "IntSize".to_owned(),
@@ -324,6 +341,8 @@ fn gen_corelib(
         .with_config(public_config)
         .with_src(crate_dir.join("timers.rs"))
         .with_src(crate_dir.join("graphics.rs"))
+        .with_src(crate_dir.join("window.rs"))
+        .with_src(crate_dir.join("api.rs"))
         .with_after_include(format!(
             r"
 /// This macro expands to the to the numeric value of the major version of SixtyFPS you're
--- a/api/cpp/include/sixtyfps.h
+++ b/api/cpp/include/sixtyfps.h
@@ -158,6 +158,8 @@ public:
         }
     }
 
+    void request_redraw() const { cbindgen_private::sixtyfps_windowrc_request_redraw(&inner); }
+
 private:
     cbindgen_private::WindowRcOpaque inner;
 };
@@ -177,7 +179,8 @@ constexpr inline ItemTreeNode make_dyn_node(std::uintptr_t offset, std::uint32_t
                                                            parent_index } };
 }
 
-inline ItemRef get_item_ref(ComponentRef component, cbindgen_private::Slice<ItemTreeNode> item_tree, int index)
+inline ItemRef get_item_ref(ComponentRef component, cbindgen_private::Slice<ItemTreeNode> item_tree,
+                            int index)
 {
     const auto &item = item_tree.ptr[index].item.item;
     return ItemRef { item.vtable, reinterpret_cast<char *>(component.instance) + item.offset };
@@ -341,6 +344,9 @@ public:
         return inner.set_rendering_notifier(std::forward<F>(callback));
     }
 
+    /// This function issues a request to the windowing system to redraw the contents of the window.
+    void request_redraw() const { inner.request_redraw(); }
+
     /// \private
     private_api::WindowRc &window_handle() { return inner; }
     /// \private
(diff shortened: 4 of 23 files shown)


CHANGELOG.md
```diff
@@ -28,6 +28,8 @@ as well as the [Rust migration guide for the `sixtyfps` crate](api/sixtyfps-rs/m
 ### Added
 
  - `TextEdit::font-size` and `LineEdit::font-size` have been added to control the size of these widgets.
+ - Use `sixtyfps::Window::set_rendering_notifier` to get a callback before and after a new frame is being rendered.
+ - Use `sixtyfps::Window::request_redraw()` to schedule redrawing of the window contents.
```
ogoffart, line 32:
```suggestion
 - Added `sixtyfps::Window::set_rendering_notifier` to get a callback before and after a new frame is being rendered.
 - Added `sixtyfps::Window::request_redraw()` to schedule redrawing of the window contents.
```

Cargo.toml
```diff
@@ -14,6 +14,7 @@ members = [
     'examples/gallery',
     'examples/imagefilter',
     'examples/memory',
+    'examples/opengl_underlay',
```
ogoffart, line 17: what do you feel about dashes?

api/cpp/cbindgen.rs
```diff
@@ -157,8 +160,9 @@ fn gen_corelib(
         "sixtyfps_color_darker",
         "sixtyfps_image_size",
         "sixtyfps_image_path",
-        "TimerMode", // included in generated_public.h
-        "IntSize",   // included in generated_public.h
+        "TimerMode",      // included in generated_public.h
+        "IntSize",        // included in generated_public.h
```
ogoffart, line 164: Unrelated... but do we really have IntSize in our public api?  It should probably be `Size`  or maybe `Size<int>`

api/cpp/include/sixtyfps.h
```diff
@@ -314,6 +328,17 @@ class Window
     /// De-registers the window from the windowing system, therefore hiding it.
     void hide() { inner.hide(); }
 
+    /// This function allows registering a callback that's invoked during the different phases of
+    /// rendering. This allows custom rendering on top or below of the scene.
+    template<typename F>
+    void set_rendering_notifier(F callback) const
+    {
+        inner.set_rendering_notifier(std::move(callback));
```
ogoffart, line 336:
```suggestion
    void set_rendering_notifier(F &&callback) const
    {
        inner.set_rendering_notifier(std::forward<F>(callback));
```

examples/opengl_underlay/build.rs
```diff
@@ -0,0 +1,7 @@
+// Copyright © SixtyFPS GmbH <info@sixtyfps.io>
+// SPDX-License-Identifier: (GPL-3.0-only OR LicenseRef-SixtyFPS-commercial)
+
+fn main() {
+    let config = sixtyfps_build::CompilerConfiguration::default().with_style("fluent".into());
```
ogoffart, line 5: why do we need to set the style?

examples/opengl_underlay/scene.60
```diff
@@ -0,0 +1,34 @@
...
+// SPDX-License-Identifier: (GPL-3.0-only OR LicenseRef-SixtyFPS-commercial)
+
+import { ScrollView, Button, CheckBox, SpinBox, Slider, GroupBox, LineEdit, StandardListView,
+    ComboBox, HorizontalBox, VerticalBox, GridBox, TabWidget, TextEdit, AboutSixtyFPS } from "sixtyfps_widgets.60";
+
+App := Window {
+    preferred-width: 500px;
+    preferred-height: 600px;
+    title: "SixtyFPS Gallery";
+    icon: @image-url("../../vscode_extension/extension-logo.png");
```
ogoffart, line 11: we should have a better location for the image

examples/opengl_underlay/scene.60
```diff
@@ -0,0 +1,34 @@
+// Copyright © SixtyFPS GmbH <info@sixtyfps.io>
+// SPDX-License-Identifier: (GPL-3.0-only OR LicenseRef-SixtyFPS-commercial)
+
+import { ScrollView, Button, CheckBox, SpinBox, Slider, GroupBox, LineEdit, StandardListView,
+    ComboBox, HorizontalBox, VerticalBox, GridBox, TabWidget, TextEdit, AboutSixtyFPS } from "sixtyfps_widgets.60";
+
+App := Window {
+    preferred-width: 500px;
+    preferred-height: 600px;
+    title: "SixtyFPS Gallery";
```
ogoffart, line 10:
```suggestion
```

Not knowing what title to use, i'd leave it the default (so it will be "SixtyFPS Window").
But we could also give a better title

examples/opengl_underlay/scene.60
```diff
@@ -0,0 +1,34 @@
...
+            background: #ffffff92;
+            HorizontalBox {
+                Text {
+                    text: "This text is rendered using SixtyFPS. The animation below is rendered using OpenGL code.";
+                    wrap: word-wrap;
+                }
+
+                VerticalLayout {
+                    alignment: start;
+                    disable-toggle := CheckBox {
```
ogoffart, line 24:
```suggestion
                    CheckBox {
```

I was wondering what the checkbox was setting.  Maybe you could make it so that it change something in the rendering

internal/backends/gl/glcontext.rs
```diff
@@ -41,7 +48,10 @@ impl OpenGLContext {
                 let current_ctx = unsafe { not_current_ctx.make_current().unwrap() };
                 OpenGLContextState::Current(current_ctx)
             }
+            #[cfg(not(target_arch = "wasm32"))]
             state @ OpenGLContextState::Current(_) => state,
+            #[cfg(target_arch = "wasm32")]
+            state @ OpenGLContextState::Current { .. } => state,
```
ogoffart, line 54:
```suggestion
            state @ OpenGLContextState::Current { .. } => state,
```

you can use the `{ .. }` syntax to match a tuple, iirc.

otherwise, just  `state => state`

internal/backends/gl/glcontext.rs
```diff
@@ -60,8 +70,15 @@ impl OpenGLContext {
         }
     }
 
+    pub fn is_current(&self) -> bool {
+        #[cfg(not(target_arch = "wasm32"))]
+        return matches!(self.0.borrow().as_ref().unwrap(), OpenGLContextState::Current(_));
+        #[cfg(target_arch = "wasm32")]
+        return matches!(self.0.borrow().as_ref().unwrap(), OpenGLContextState::Current { .. });
```
ogoffart, line 77:
```suggestion
        return matches!(self.0.borrow().as_ref().unwrap(), OpenGLContextState::Current { .. });
```

internal/backends/qt/qt_window.rs
```diff
@@ -1349,6 +1349,13 @@ impl PlatformWindow for QtWindow {
         }
     }
 
+    fn set_rendering_notifier(
+        &self,
+        _callback: Box<dyn sixtyfps_corelib::window::RenderingNotifier>,
+    ) {
+        panic!("set_rendering_notifier() is not supported in the Qt backend.")
```
ogoffart, line 1356: Is it better to panic or to ignore? or maybe print a warning?
Is there a way to detect it from the application whether or not it's going to be supported?
Perhaps one should return an Error

internal/core/window.rs
```diff
@@ -678,4 +769,47 @@ pub mod ffi {
...
+        impl Drop for CNotifier {
+            fn drop(&mut self) {
+                (self.drop_user_data)(self.user_data)
+            }
+        }
+
+        impl RenderingNotifier for CNotifier {
+            fn notify(&mut self, rust_state: super::RenderingState) {
+                let cpp_state = match rust_state {
+                    super::RenderingState::RenderingSetup { .. } => RenderingState::RenderingSetup,
```
ogoffart, line 796: why is it ok to ignore the contents?

internal/core/window.rs
```diff
@@ -564,6 +568,74 @@ pub trait WindowHandleAccess {
...
+    /// The rendering is done on a HTML Canvas element using WebGL.
+    WebGL {
+        /// The DOM element id of the HTML Canvas element used for rendering.
+        canvas_element_id: &'a str,
+        /// The drawing context type used on the HTML Canvas element for rendering. This is the argument to the
+        /// `getContext` function on the HTML Canvas element.
+        context_type: &'a str,
+    },
+    #[doc(hidden)]
+    Other,
```
ogoffart, line 590:
```suggestion
```

I don't think it is used, so we can as well leave it out, right?

tronical: That's true, the change that adds the WebGL variant can remove this.

internal/core/window.rs
tronical: Exposing the Rust API to C++ doesn't make sense here. Neither the WebGL bits nor the non-ffi-safe get-proc-address function. The former is never used in C++, the latter isn't typically used: It's common to link against the GL libraries directly, instead of the Rust approach of dynamic linkage for the bindings.

If/when we switch to WGPU, then we could perhaps expose the C API: https://github.com/gfx-rs/wgpu-native, if that's somehow possible.

internal/backends/qt/qt_window.rs
tronical: That's a very good question. The the very least the function should indeed return an error if it fails, i.e. if the backend doesn't support it - or if it is called twice then. What would be the right type? `std::result::Result<(), Box<dyn std::error::Error>`?

internal/backends/gl/glcontext.rs
tronical: Thanks, that's indeed much better.

examples/opengl_underlay/scene.60
tronical: You're right, that should be removed.

examples/opengl_underlay/scene.60
tronical: Yes, I was thinking as well - we could add functionality. How about toggling whether to rotate the entire "thing" or not?

api/cpp/cbindgen.rs
tronical: ACK, filed #909

Cargo.toml
tronical: Oh I broke convention - indeed. I'll fix that :)

examples/opengl_underlay/build.rs
ogoffart: Normally it should default to native only if the Qt backend is in the dependency tree. And it is ok to use the native style with the GL backend at runtime.

If you had an error maybe the style detection didn't work. There is a bug that you need to recompile twice after changing the feature set because the build script is not re-run

examples/opengl_overlay_cpp/CMakeLists.txt
```diff
@@ -0,0 +1,13 @@
+# Copyright © SixtyFPS GmbH <info@sixtyfps.io>
+# SPDX-License-Identifier: (GPL-3.0-only OR LicenseRef-SixtyFPS-commercial)
+
+cmake_minimum_required(VERSION 3.14)
+project(opengl_overlay_cpp LANGUAGES CXX)
```
ogoffart, line 5: Can we remove the _cpp from the example name.  (Also in the directory name)

Cargo.toml
tronical: On second thought, I think we have a lot more `_` than `-` (see under `tools` and `examples`, like `fancy_demo`, `slide_puzzle`, etc. - `iot-dashboard` is the odd one out).

ogoffart: ... `printerdemo_old`

but I think that the underscores are legacy...

Anyway, up to you.

examples/opengl_overlay_cpp/CMakeLists.txt
ogoffart: The idea is that we could have both rust and cpp code for the same demo  (like we have for other demos)

internal/backends/qt/qt_window.rs
ogoffart: Maybe `std::result::Result<(), SetRenderingNotifierError>`

```rust 
enum SetRenderingNotifierError { Unsupported, AlreadySet }
```

Something like that.

internal/backends/gl/glcontext.rs
tronical: Awesome, I didn't know that `{ .. }` also matches a tuple. Then I can get rid of the function altogether :)

examples/opengl_overlay_cpp/CMakeLists.txt
tronical: This is done now.

examples/opengl_underlay/scene.60
tronical: This is implemented now.

examples/opengl_underlay/build.rs
tronical: I had some local issues but it works indeed, so this is now removed :)

ogoffart:
For the example, there could be a link from examples/README.md
and maybe some explanation of what the example does. (either in examples/README.md or in the readme or doc of that example)

internal/core/window.rs
```diff
@@ -32,7 +32,10 @@ pub trait PlatformWindow {
 
     /// This function is called through the public API to register a callback that the backend needs to invoke during
     /// different phases of rendering.
-    fn set_rendering_notifier(&self, callback: Box<dyn crate::api::RenderingNotifier>);
+    fn set_rendering_notifier(
+        &self,
+        callback: Box<dyn crate::api::RenderingNotifier>,
+    ) -> std::result::Result<(), crate::api::SetRenderingNotifierError>;
```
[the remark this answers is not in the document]
tronical: Oh yes, that's an excellent idea - will do. Thanks!

tronical:
> For the example, there could be a link from examples/README.md
> and maybe some explanation of what the example does. (either in examples/README.md or in the readme or doc of that example)

 - [ ] Good point, I'll add a README in the directory and link to it from `examples/README.md`

ogoffart (approved):
:+1:

tronical:
Rebasing now, then squashing.

PR #913, opened 2022-02-05
Rust generator: Fix compilation returning from void callbacks

ogoffart:
If a callback returning voids ends in a statement that returns something,
we would have an error in the rust generated code as we end our function
with an expression that is not `()`


hunger (approved):
This looks sensible to me.

PR #916, opened 2022-02-06
Printer Demo: ability to change the image in the preview

ogoffart:


ogoffart:
Preview: https://sixtyfps.io/snapshots/master/editor/preview.html?load_url=https://raw.githubusercontent.com/sixtyfpsui/sixtyfps/olivier/demos/examples/printerdemo/ui/printerdemo.60
(in the USB page)

BTW: the cat picture does not work in the online editor/ preview, because it is a symbolic link. This is not a regression from this patch but was already broken in the 0.1.6 release since we made it a symlink.
Probably we want to keep it as a copy or as a reference to the file directly, but we should avoid symlinks in our demos if we want them to be loaded through githubusercontent

examples/printerdemo/ui/usb_page.60
```diff
@@ -5,21 +5,37 @@ import { DemoPalette, Page, SpinBox, Label, ComboBox, PushButton, CheckBox } fro
...
+struct File := { name: string, preview: image }
 
 export UsbPage := Page {
     has-back-button: true;
     header: "USB";
 
+    property <[File]> files: [
+        { name: ".." },
+        { name: "cat.png", preview: @image-url("images/cat_preview_round.png") },
+        { name: "dog.png", preview: @image-url("images/dog.jpg") },
```
tronical, line 17: The name says dog.png but the file is jpeg. Is that intentional?

examples/printerdemo/ui/usb_page.60
```diff
@@ -31,11 +47,13 @@ export UsbPage := Page {
         }
 
         list-view := StandardListView {
+            current-item: 1;
             col: 1;
             row: 1;
             colspan: 2;
             horizontal-stretch: 1;
             vertical-stretch: 1;
+            // TODO: maybe we want something like `model: files.map(item => { text: item.preview })`
```
tronical, line 56: That would be neat :)

tronical (approved):
Looks good to me, just a jpeg vs png mismatch. Is jpeg preferred over png?

PR #921, opened 2022-02-07
Fix sizing of image with opacity in for with layout

### LINKED ISSUE #915: opacity within a `for` messes up the size
Starting with sixtyfps v0.1.6, images in a certain part of my GUI are no longer resized but instead always shown with their full size. This is the affected .60 snippet:

```
ScrollView {
padding: 0px;
viewport-height: 100px;
height: similar-images-model.length > 1 ? 125px : 0px;
visible: similar-images-model.length > 1;

HorizontalBox { 
    alignment: start;
for item[i] in similar-images-model:
    Image {
        source: item.image;
        opacity: item.take-over ? 1.0 : 0.2;
        image-fit: contain;
        width: 180px;
        height: 100px;
        TouchArea { 
            clicked => {
                current-image.image = item.image;
                current-image.take-over = item.take-over;
                current-image.text = item.text;
                current-image.local-index = item.local-index;
            }
        }
        if (item.local-index == current-image.local-index) : Rectangle { opacity: 0.4; background: blue;}
    }
}
```

That's how it looks with v0.1.5:
Screenshot from v0.1.5

And here with v0.1.6
Screenshot from v0.1.6

I was not yet able to create a minimum example to narrow it down.

ogoffart:
Fixes #915

See commit messages for details

### DIFF
--- a/tests/cases/layout/opacity_in_layout.60
+++ b/tests/cases/layout/opacity_in_layout.60
@@ -14,7 +14,16 @@ export TestCase := Window {
             opacity: 1;
         }
     }
+
+    // issue #915
+    layout3 := HorizontalLayout {
+        for _ in 3 : Rectangle {
+            width: 10px;
+            opacity: 0.5;
+        }
+    }
     property <length> layout1_height : layout1.min-height;
     property <length> layout2_width : layout2.max-width;
-    property<bool> test: layout1_height == 55px && layout2_width == 88px;
+    property <length> layout3_width : layout3.preferred-width;
+    property<bool> test: layout1_height == 55px && layout2_width == 88px && layout3_width == 30px;
 }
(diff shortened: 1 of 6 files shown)


tests/cases/layout/opacity_in_layout.60
```diff
@@ -14,7 +14,16 @@ export TestCase := Window {
...
         }
     }
+
+    // issue #915
+    layout3 := HorizontalLayout {
+        for _ in 3 : Rectangle {
+            width: 10px;
+            opacity: 0.5;
+        }
+    }
```
tronical, line 23: Do I understand correctly that  here the Rectangle would become an "optimised" element because only the `width` is set, the opacity is moved into the hidden `Opacity` element?

tronical (approved):

